PR #31, opened 2012-01-05
Add show and list "time" templates

EmmanuelVella:
Related to PR #475


webdevilopers:
BTW: Is there an render setting / field option to influence the format e.g. `H:i` without overriding the template?
```php
        $listMapper
            ->add('workBegin', 'time', array('format' => 'H:i'))
```

PR #156, opened 2012-11-10
Use field groups for inline editing of embedded admins

ethanresnick:
This diff is a bit hard to follow because it has the new features together with spacing and other tweaks. Apologies. 

But the basic idea is this: for the inline table editing mode, each field is currently given it's own cell in the table. For associated admins with more than a few fields, this makes the each `<tr>` very wide/awkward and, more importantly, it ignores the groups set up by the user in the associated admin class. This PR implements a more-usable approach, in which each cell/column holds a group rather than a single field, that way there's a clear hierarchy of *group heading* -> *group fields* as you go across each row.

### DIFF
--- a/Resources/views/CRUD/edit_orm_one_to_many.html.twig
+++ b/Resources/views/CRUD/edit_orm_one_to_many.html.twig
@@ -13,48 +13,56 @@ file that was distributed with this source code.
         {{ element|render_relation_element(sonata_admin.field_description) }}
     {% endfor %}
 {% else %}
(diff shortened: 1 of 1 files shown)


greg0ire:
> This diff is a bit hard to follow because it has the new features together with spacing and other tweaks. Apologies. 

Maybe you could make several commits? `git add --patch` is your friend.

ethanresnick:
@greg0ire I haven't used Sonata in years, so I wouldn't even know where to start with this PR at this point. But I'll leave it open in case anyone else wants to run with it. Or, you're free to close it!

PR #163, opened 2012-12-10
[Fix] One-to-many inline edit view issue when used with INDEX BY 

kagux:
If we inline one-to-many edit using table, then it takes headers for this table from form's child with key ```0```. However, if we use ```INDEX BY``` in our query, then children might not have key ```0``` and we end up with error ```Key "0" for array with keys "en" does not exist in SonataDoctrineORMAdminBundle:CRUD:edit_orm_one_to_many.html.twig at line 25```
To fix this instead of  ```0``` key we have to take first key from children keys and use it.

### DIFF
--- a/Resources/views/CRUD/edit_orm_one_to_many.html.twig
+++ b/Resources/views/CRUD/edit_orm_one_to_many.html.twig
@@ -22,7 +22,7 @@ file that was distributed with this source code.
                         <table class="table table-bordered">
                             <thead>
                                 <tr>
-                                    {% for field_name, nested_field in form.children|first.children %}
+                                    {% for field_name, nested_field in (form.children|first).children %}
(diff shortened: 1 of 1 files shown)


core23:
Are you still working on this? Please squash and rebase or close this PR @kagux

kagux:
@core23 I obviously moved on long time ago, my PR was 3 1/2 years ago :) 
I'll try to rebase over weekend and see what comes out.

OskarStark:
👍  @kagux

kagux:
Done.
It actually appears as if changes similar to this PR were already introduced at some point. 
Updated PR only adds brackets :)

core23:
IMHO we don't need the brackets for this

Soullivaneuh:

Regards.

PR #239, opened 2013-07-14
Fix an issue where rendering of the n-1 crud element would sometimes fail

koyaan:
This fixes a bug i stumbled upon in the sonata sandbox. If you create a new gallery block in the container container block and try to edit it, a twig exception gets thrown, i traced it down to the rendering of the relationship element which fails for gallery blocks which do not have at least GalleryID=false in their settings.


core23:
Are you still working on this? Please squash and rebase or close this PR @koyaan

koyaan:
Can´t believe this is still relevant!

PR #309, opened 2013-12-11
Update hidden input by adding newly added option to selectbox of orm-man...

katstra:
When adding a new media item inline while editing a block, the select box is not updated after adding the new media. Current behaviour, the very first item gets selected. Commit adds the new id to the list.


core23:
Are you still working on this? Please squash and rebase or close this PR @katstra

PR #345, opened 2014-04-16
doctrine_orm_date type with = operator

cassianotartari:
I don't know if I'm doing something wrong but when I use doctrine_orm_date in filter to a datetime field with the default operator ('=') it don't work because the sql tries to find some record with the exactly same datetime, ie, yyyy-mm-dd 00:00:00 so it will just return if some record has this time.  What did is compare the range of one day, starting at the giving date to next day:

In the file: Sonata\DoctrineORMAdminBundle\Filter\AbstractDateFilter in the "else" of "if ($this->range) {" is now like this:

if (!$data['value']) {
                return;
            }

            // default type for simple filter
            $data['type'] = !isset($data['type']) || !is_numeric($data['type']) ? DateType::TYPE_EQUAL : $data['type'];

            // just find an operator and apply query
            $operator = $this->getOperator($data['type']);

			//var_dump($operator); // "="

            // transform types
            if ($this->getOption('input_type') == 'timestamp') {
                $data['value'] = $data['value'] instanceof \DateTime ? $data['value']->getTimestamp() : 0;
            }
			

            // null / not null only check for col
            if (in_array($operator, array('NULL', 'NOT NULL'))) {
                $this->applyWhere($queryBuilder, sprintf('%s.%s IS %s ', $alias, $field, $operator));
            } else {
				
				if($operator === '=') {
					
					$this->applyWhere($queryBuilder, sprintf('%s.%s %s :%s', $alias, $field, '>=', $parameterName));

					$end = clone $data['value'];
					$endDateParameterName = $this->getNewParameterName($queryBuilder);
					$this->applyWhere($queryBuilder, sprintf('%s.%s %s :%s', $alias, $field, '<', $endDateParameterName));
					if ($this->getOption('input_type') == 'timestamp') {
						$end = strtotime('+1 day', $end);
					} else {
						$end->add(new \DateInterval('P1D'));
					}

					$queryBuilder->setParameter($endDateParameterName, $end);
				} else {
					$this->applyWhere($queryBuilder, sprintf('%s.%s %s :%s', $alias, $field, $operator, $parameterName));
				}
            }


rande:
fixed by https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/477

PR #346, opened 2014-04-17
Update AbstractDateFilter.php

cassianotartari:
I was testing date filter with "=" operator in a datetime field and looking to the SQL generated I've observed that it compares the field with a date like yyyy-mm-dd 00:00:00 so never matches any record because it is almost impossible to have some record with 00:00:00 time. Reading a SQL solution for this the best and easy solution was to compare the 1 day range starting in 0 time to next day.


rande:
fixed by https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/477

PR #362, opened 2014-07-03
ModelManager::find allow none string id

boekkooi:
Updated ModelManager::find to only explode the id argument if it is a string. This is according to the [Sonata\AdminBundle\Model\ModelManagerInterface::find](https://github.com/sonata-project/SonataAdminBundle/blob/master/Model/ModelManagerInterface.php#L74) doc block (id = mixed).

Possible BC break if someone relies on string casting.


boekkooi:

PR #406, opened 2014-11-27
Removed javascript code blocking select button

### LINKED ISSUE #404: sonata_type_model_list select not working
The model pop up "Select" button to select existing entries does not do anything. 

Commenting the following lines in Resources/views/CRUD/edit_orm_many_association_script.html.twig fixes the issue
``` javascript
if (this.nodeName == 'A' && (target.attr('href').length == 0 || target.attr('href')[0] == '#')) {
    Admin.log('[{{ id }}|field_dialog_form_list_link] element is an anchor, skipping action', this);
    return;
}
```

tseho:
The commit ce5bd40 broke the select button when using sonata_type_model_list.
I removed this condition since it's blocking the default select action and I can't see how it's usefull.

fixes #404

### DIFF
--- a/Resources/views/CRUD/edit_orm_many_association_script.html.twig
+++ b/Resources/views/CRUD/edit_orm_many_association_script.html.twig
@@ -35,12 +35,6 @@ This code manage the many-to-[one|many] association field popup
 
         var target = jQuery(this);
 
-        // return if the link is an anchor inside the same page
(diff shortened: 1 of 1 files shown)


pulzarraider:
The code that breaks sonata_admin_model_list adds features for SonataAdminBundle `master` version, but breaks 2.3. I have created same PR as this one to fix the 2.3 branch. In master is working correctly.

@tseho thank you for this PR.

PR #410, opened 2014-12-10
Allow value object identifiers

boekkooi:
After updating to 2.3 I got some strange error messages about entities not being found etc.

I seem that some where along the way the support of custom doctrine types as identifiers got lost this patch fixes this.

### DIFF
--- a/Model/ModelManager.php
+++ b/Model/ModelManager.php
@@ -47,7 +47,7 @@ class ModelManager implements ModelManagerInterface
     }
 
     /**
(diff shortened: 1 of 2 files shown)


boekkooi:

rande:
@boekkooi why closing issues ? You might not use SonataAdminBundle, however the different issues you closed are valid and need some reviews before merging. Maintainers are also busy with their own internals projects so things go no as fast as we all want.

PR #285, opened 2013-10-10
Fixed DateRange for datetime fields

sirian:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

This patch fix query generation when your entity has datetime field and you use doctrine_orm_date_range in filters

PR #412, opened 2014-12-17
AbstractDateFilter implementation overhaul

boekkooi:
So as all know who have tried to use the date filter for doctrine is that it's a bit strange.

This PR tries to make the AbstractDateFilter simpler and easier to extend.  
It also tries to deprecate #346, #345 and #285

### DIFF
--- a/Filter/DateFilter.php
+++ b/Filter/DateFilter.php
@@ -14,16 +14,45 @@ namespace Sonata\DoctrineORMAdminBundle\Filter;
 class DateFilter extends AbstractDateFilter
 {
     /**
-     * This filter has no range
-     *
-     * @var boolean
+     * {@inheritdoc}
      */
-    protected $range = false;
+    public function getFieldType()
+    {
+        return $this->getOption('field_type', 'date');
+    }
 
     /**
-     * This filter does not allow filtering by time
-     *
-     * @var boolean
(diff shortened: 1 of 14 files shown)


Filter/TimeFilter.php
```diff
@@ -14,16 +14,18 @@
...
     /**
-     * This filter has no range
-     *
-     * @var boolean
+     * {@inheritdoc}
      */
-    protected $range = false;
+    protected function getRenderName()
+    {
+        return 'sonata_type_filter_time';
```
pulzarraider: There is no filter type called `sonata_type_filter_time`.

Filter/DateFilter.php
```diff
@@ -14,16 +14,45 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function normalizeValue($value, $ranged, $filterType)
+    {
+        if ($ranged && !is_array($value)) {
+            /** @var \DateTime $start */
+            $start = $value->setTime(0, 0, 0);
```
pulzarraider: Error: Call to a member function setTime() when using simple filter:
```
$datagridMapper->add('dateChange', 'doctrine_orm_date');
```

boekkooi:

PR #405, opened 2014-11-22
Don't overwrite list templates defined in config

Soullivaneuh:
Same method as #235 but for list templates.

Also remove trailing whitespaces on `showBuilder` class.

Could you push a new 2.3 stable version after this merge?

Thanks.

PR #413, opened 2014-12-22
Make relation-templates overwriteable

Gladhon:
Currently, templates of relations are hardcoded and can't be overwritten via configuration.
I have changed this, first checking for config and if no one found load the hardcoded one


pulzarraider:
Closing in favor of #405.

PR #434, opened 2015-03-14
LabelFix

mushkin-v:
The problem that the media-type's label isn't passed to the template which results in an error when invoking the translator/calling the trans method in the template.

### DIFF
--- a/Resources/views/CRUD/edit_orm_one_to_many.html.twig
+++ b/Resources/views/CRUD/edit_orm_one_to_many.html.twig
@@ -27,7 +27,9 @@ file that was distributed with this source code.
                                             <th>{{ 'action_delete'|trans({}, 'SonataAdminBundle') }}</th>
                                         {% else %}
                                             <th {{ nested_field.vars['required']  ? 'class="required"' : '' }}{% if (nested_field.vars['attr']['hidden'] is defined) and (nested_field.vars['attr']['hidden']) %} style="display:none;"{% endif %}>
-                                                {{ nested_field.vars['sonata_admin'].admin.trans(nested_field.vars.label) }}
+                                                {% if nested_field.vars.label %}
+                                                    {{ nested_field.vars['sonata_admin'].admin.trans(nested_field.vars.label) }}
+                                                {% endif %}
                                             </th>
                                         {% endif %}
                                     {% endfor %}


rande:
Can you provide more information ? It can be a bug in the MediaBundle ?

mushkin-v:
The problem is when nested_field.vars.label does not exist, the method TRANS in {{ nested_field.vars['sonata_admin'].admin.trans(nested_field.vars.label) }}  - fails... Error message is 'Item "trans" for "" does not exist in SonataDoctrineORMAdminBundle:CRUD:edit_orm_one_to_many.html.twig at line 29'...

mushkin-v:
Here is another example of this error: 

https://groups.google.com/forum/#!topic/sonata-users/cq9M9dIXgeg 

Thank you.

mushkin-v:
Hello today once again meet this problem... The problem is the same but with edit_orm_many_to_many.html.twig... Here what we have:
Two entities with fields wich have connection many-to-many, here this fields
    /**
     * @ORM\ManyToMany(targetEntity="AppBundle\Entity\Tag", inversedBy="posts", cascade={"persist"})
     * @ORM\JoinTable(name="post_tag")
     */
    private $tags;

and second field in another entity:
    /**
     * @ORM\ManyToMany(targetEntity="AppBundle\Entity\Post", mappedBy="tags", cascade={"persist"})
     *
     */
    private $posts;

So how can you see it is simple connection between posts and tags. Lets go next.
We have Admin classes for both. So when I write in my TagAdmin class this:
        ->add('posts', 'sonata_type_collection',
                [
                    'by_reference' => false,
                ], [
                    'edit' => 'inline',
                    'inline' => 'table',
                    'sortable'  => 'position',
                ]
            )
        ;

I HAVE SUCH ERROR:
Impossible to invoke a method ("trans") on a NULL variable ("") in SonataDoctrineORMAdminBundle:CRUD:edit_orm_many_to_many.html.twig at line 25

So my PR tries to fix the same problem for CRUD/edit_orm_one_to_many.html.twig. If you like and merge my PR, I will do the same for CRUD:edit_orm_many_to_many.html.twig.  Thank you.

mushkin-v:
And there is no use of MediaBundle, so it is not a bug in MediaBundle. Thank you again))

OskarStark:
@rande any news?

Soullivaneuh:

Regards.

PR #440, opened 2015-03-23
support order asc nulls first for postgresql database

josephzhao:
Hello

SonataDoctrineORMAdminBundle

 support order asc nulls first for postgresql database

in Sonata\ClassificationBundle\Entity\CategoryManager

protected function loadCategories(ContextInterface $context)
    {
        if (array_key_exists($context->getId(), $this->categories)) {
            return;
        }
        $class = $this->getClass();
        $categories = $this->getObjectManager()->createQuery(sprintf('SELECT c FROM %s c WHERE c.context = :context ORDER BY c.parent ASC NULLS FIRST', $class))
            ->setParameter('context', $context->getId())
            ->execute();

Thanks

Joseph

### DIFF
--- a/Admin/FieldDescription.php
+++ b/Admin/FieldDescription.php
@@ -3,7 +3,7 @@
 /*
  * This file is part of the Sonata package.
  *
- * (c) 2010-2011 Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
  *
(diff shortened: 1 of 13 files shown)


pulzarraider:
@josephzhao `NULLS FIRST` is not supported in Mysql so we cannot use it in universal code. If you need it like this, you have to overload the code.

Anyway, it seems related to the [SonataClassificationBundle](https://github.com/sonata-project/SonataClassificationBundle) so this issue belongs there,

PR #447, opened 2015-04-19
Do not add left join to already joined entities with custom query

pulzarraider:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | sonata-project/SonataAdminBundle#2884
| License       | MIT

If developer overload `createQuery` method in admin class, we should not `leftJoin` entities that are already joined. Too many same `leftJoin`s breaks the query performance and may cause unexpected results.

### DIFF
--- a/Datagrid/ProxyQuery.php
+++ b/Datagrid/ProxyQuery.php
@@ -51,11 +51,11 @@ class ProxyQuery implements ProxyQueryInterface
         if ($this->getSortBy()) {
             $sortBy = $this->getSortBy();
             if (strpos($sortBy, '.') === false) { // add the current alias
-                $sortBy = $queryBuilder->getRootAlias() . '.' . $sortBy;
+                $sortBy = $queryBuilder->getRootAlias().'.'.$sortBy;
             }
             $queryBuilder->addOrderBy($sortBy, $this->getSortOrder());
         } else {
-            $queryBuilder->resetDQLPart( 'orderBy' );
+            $queryBuilder->resetDQLPart('orderBy');
         }
 
         return $this->getFixedQueryBuilder($queryBuilder)->getQuery()->execute($params, $hydrationMode);
@@ -83,7 +83,7 @@ class ProxyQuery implements ProxyQueryInterface
         // step 3 : retrieve the different subjects id
         $select = sprintf('%s.%s', $queryBuilderId->getRootAlias(), $idName);
         $queryBuilderId->resetDQLPart('select');
-        $queryBuilderId->add('select', 'DISTINCT ' . $select);
+        $queryBuilderId->add('select', 'DISTINCT '.$select);
 
         // for SELECT DISTINCT, ORDER BY expressions must appear in select list
         /* Consider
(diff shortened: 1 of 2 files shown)


webdevilopers:
Has anything changed in the last days with this behaviour @pulzarraider ?

Suddenly I have a problem similar to the original one again.

This time I simply add - and only add(!) - a filter:
```php
        $filterMapper
            ->add('supplierContacts.contract.supplierBranch', 'doctrine_orm_model_autocomplete', array(
                    'operator_type' => 'hidden',
                ), null, array(
                    'property' => 'name'
                ))
        ;
```

My original query method:
```php
    {

        $query->addSelect(
                'PARTIAL ' . $query->getRootAlias() . '.{id, firstName, lastName, email, phone, fax, postcode, street, city, country}',
                'PARTIAL u.{id, username}'
            )
            ->leftJoin($query->getRootAlias() . '.user', 'u')
        ;

        return $query;
    }
```

Without even using the filter at all it starts adding LEFT JOINs to my query which results in:
```sql
SELECT 
  DISTINCT p0_.id AS id_0, 
  p0_.id AS id_1 
FROM 
  persons p0_ 
  LEFT JOIN users u1_ ON p0_.user_id = u1_.id 
  LEFT JOIN contacts c2_ ON p0_.id = c2_.contact_id 
  AND c2_.contacttype IN ('supplier') 
  LEFT JOIN contracts c3_ ON c2_.contract_id = c3_.id 
  LEFT JOIN branches b4_ ON c3_.supplier_branch_id = b4_.id 
  LEFT JOIN compartments c5_ ON p0_.compartment_id = c5_.id 
ORDER BY 
  p0_.id ASC 
LIMIT 
  10 OFFSET 0
```

I guess this behaviour was introduced to select an autocomplete label if a filter was applied but in my case the filter was only added to the `$filterMapper` - not used at all.

webdevilopers:
I can confirm that the nested sub.properties cause the issue. I tried a different one:
```php
        $filterMapper
//            ->add('supplierContacts.contract.supplierBranch', 'doctrine_orm_model_autocomplete', array(
            ->add('supplierContacts.contract', 'doctrine_orm_model_autocomplete', array(
                    'operator_type' => 'hidden',
                ), null, array(
                    'property' => 'number'
                ))
        ;
```

The add these LEFT JOINs:
```sql


SELECT 
  DISTINCT p0_.id AS id_0, 
  p0_.id AS id_1 
FROM 
  persons p0_ 
  LEFT JOIN users u1_ ON p0_.user_id = u1_.id 
  LEFT JOIN contacts c2_ ON p0_.id = c2_.contact_id 
  AND c2_.contacttype IN ('supplier') 
  LEFT JOIN contracts c3_ ON c2_.contract_id = c3_.id 
  LEFT JOIN compartments c4_ ON p0_.compartment_id = c4_.id 
ORDER BY 
  p0_.id ASC 
LIMIT 
  10 OFFSET 0
```

I think the auto-fire still is the wrong behaviour since no value has been selected by the filter.

Is there a way to prevent it?

pulzarraider:
Thank you for reporting the issue @webdevilopers. I will look at it this week.

PR #407, opened 2014-11-28
Fix ProxyQuery error on relational identifiers

Soullivaneuh:
Actually, the ProxyQuery can't handle identifier with relational link properly. This PR patch this.

It's a corrected version of #226 and concern issus #225, https://github.com/sonata-project/SonataAdminBundle/issues/1392 and https://github.com/sonata-project/SonataAdminBundle/issues/1000 as far I know.

I think it's quite critical, please have a look! :+1: 

Thanks.

I give you some steps to reproduce the error with the current master version.

If we have an entity with relational identifiers like this:

```php
/**
 * UserBrowser
 *
 * @ORM\Table(name="user_browser")
 * @ORM\Entity
 */
class UserBrowser
{
    /**
     * @var User
     *
     * @ORM\Id
     * @ORM\ManyToOne(targetEntity="Nexy\UserBundle\Entity\User", inversedBy="browsers")
     * @ORM\JoinColumn(name="user_id", referencedColumnName="id")
     */
    private $user;

    /**
     * @var User
     *
     * @ORM\Id
     * @ORM\ManyToOne(targetEntity="AppBundle\Entity\Browser")
     * @ORM\JoinColumn(name="browser_id", referencedColumnName="id")
     */
    private $browser;

    // [...]
```

And a configured admin like this:

```php
/**
 * Class UserBrowserAdmin
 */
class UserBrowserAdmin extends Admin
{
    protected $baseRouteName = 'admin_user_browser';
    protected $baseRoutePattern = 'user/browser';
    protected $datagridValues = array(
        '_sort_by'      => 'lastUsedAt',
        '_sort_order'   => 'DESC',
    );

    protected function configureListFields(ListMapper $list)
    {
        $list
            ->addIdentifier('user')
            ->addIdentifier('browser')
            ->add('lastUsedAt')
            ->add('_action', 'actions', array(
                'actions'   => array(
                    'show'      => array(),
                )
            ))
        ;
    }

    protected function configureShowFields(ShowMapper $filter)
    {
        $filter
            ->add('user')
            ->add('browser.id')
            ->add('lastUsedAt')
            ->add('browser.details')
        ;
    }
}
```

We got this error:

```
An exception has been thrown during the rendering of a template ("[Semantical Error] line 0, col 18 near 'user FROM AppBundle\Entity\UserBrowser': Error: Invalid PathExpression. Must be a StateFieldPathExpression.") in SonataAdminBundle:CRUD:base_list.html.twig at line 33.
```

PR #476, opened 2015-07-21
Build DISTINCT query only if needed

### LINKED ISSUE #473: List view for big tables very slow due to DISTINCT query
There was an issue where this was mentioned but at the end not adressed: #297
We're using the SimplePager which indeed speeds things up as it does not do a count(...) query.

To visualize the difference, first one with DISTINCT, second one without:

```
drivenow_production=# explain analyze SELECT 
  DISTINCT e0_.id AS id_0, 
  e0_.timestamp AS timestamp_1 
FROM 
  EmittedEventRecord e0_ 
ORDER BY 
  e0_.timestamp DESC 
LIMIT 
  33 OFFSET 0;
                                                                       QUERY PLAN                                                                       
--------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=795326.66..795326.91 rows=33 width=45) (actual time=64059.389..64059.467 rows=33 loops=1)
   ->  Unique  (cost=795326.66..814667.44 rows=2578770 width=45) (actual time=64059.386..64059.432 rows=33 loops=1)
         ->  Sort  (cost=795326.66..801773.59 rows=2578770 width=45) (actual time=64059.384..64059.401 rows=33 loops=1)
               Sort Key: "timestamp", id
               Sort Method: external merge  Disk: 147920kB
               ->  Seq Scan on emittedeventrecord e0_  (cost=0.00..282727.70 rows=2578770 width=45) (actual time=0.040..48492.275 rows=2608727 loops=1)
 Total runtime: 64102.981 ms
```

```
drivenow_production=# explain analyze SELECT 
  e0_.id AS id_0, 
  e0_.timestamp AS timestamp_1 
FROM 
  EmittedEventRecord e0_ 
ORDER BY 
  e0_.timestamp DESC 
LIMIT 
  33 OFFSET 0;
                                                                                   QUERY PLAN                                                                                   
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.43..4.77 rows=33 width=45) (actual time=0.037..0.174 rows=33 loops=1)
   ->  Index Scan Backward using emittedeventrecordtimestamp on emittedeventrecord e0_  (cost=0.43..339144.15 rows=2578780 width=45) (actual time=0.034..0.115 rows=33 loops=1)
 Total runtime: 0.232 ms
```

Probably we can improve a bit by tuning Postgres, but generally I totally don't understand why doing DISTINCT on the primary key(s) field(s) is required?

It seems there is no sustainable way to change this behaviour as we'd have to overwrite a couple of services, so we'd have to change if the bundles are updated.

I guess a config option to disable the DISTINCT behaviour would do the job. This was suggested in the closed issue reference above but never done.

Best regards, Anton

AntonStoeckl:
This adds DISTINCT to the query to get IDs only if there are multiple select parts.
Fixes #473

### DIFF
--- a/Datagrid/ProxyQuery.php
+++ b/Datagrid/ProxyQuery.php
@@ -97,8 +97,12 @@ class ProxyQuery implements ProxyQueryInterface
             }
             $idxSelect .= ($idxSelect !== '' ? ', ' : '').$idSelect;
         }
+
+        $needsDistinct = (count($selects) > 1) ? true : false;
+        $distinct = ($needsDistinct === true) ? 'DISTINCT ' : '';
+
         $queryBuilderId->resetDQLPart('select');
-        $queryBuilderId->add('select', 'DISTINCT '.$idxSelect);
+        $queryBuilderId->add('select', $distinct.$idxSelect);
 
         // for SELECT DISTINCT, ORDER BY expressions must appear in idxSelect list
         /* Consider
@@ -107,7 +111,7 @@ class ProxyQuery implements ProxyQueryInterface
         values.  Which one will you use to sort that x-value in the output?
         */
         // todo : check how doctrine behave, potential SQL injection here ...
-        if ($this->getSortBy()) {
+        if ($needsDistinct === true && $this->getSortBy()) {
             $sortBy = $this->getSortBy();
             if (strpos($sortBy, '.') === false) { // add the current alias
                 $sortBy = $queryBuilderId->getRootAlias().'.'.$sortBy;


Soullivaneuh:
Well, this is not a configuration.

Why remove distinct if multiple select? Can you explain a little bit more?

Please note that distinct is used for getting IDs for last entities selection.

ping @rande

AntonStoeckl:
Well, it removes distinct if there is only ONE identifier field, with multiple it keeps distinct.
Identifiers must always be unique, no need for distinct, no need to select the sort field(s).
Result is a query that performs well without a compound index over id_field + sort_field.

AntonStoeckl:
I still think even for a compond identifier / primary key the distinct is not necessary because they must also be unique, but maybe there is something I'm missing / some case I don't see.

Soullivaneuh:
Please consider https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/407. Is this case will still be valid?

AntonStoeckl:
I see no specific problem with #407 

The question is, if a table is created without a primary key, what will $metadata->getIdentifierFieldNames() return?
If it returns nothing, there is a different problem, resulting in a broken query.

AntonStoeckl:
Seems to have to many unknowns.
I'm trying a different approach, disabling distinct in the query via passing something in $datagridValues

PR #481, opened 2015-08-20
[ProxyQuery] Convert identifiers values to database value

ju1ius:
This allows using custom DBAL types as identifiers.

For example when using the [UuidBinaryType](https://github.com/ramsey/uuid-doctrine/blob/master/src/UuidBinaryType.php) from the `ramsey/uuid-doctrine` package:
```php
use Ramsey\Uuid\Uuid;

class Example
{
    /**
     * @var \Ramsey\Uuid\Uuid
     *
     * @ORM\Column(type="uuid_binary")
     * @ORM\Id
     */
    private $id;

    public function __construct()
    {
        $this->id = Uuid::uuid4();
    }
}
```

the identifier cannot simply be cast to string, since the database requires it's binary representation.
So we must make sure we convert the identifiers to their database value.

### DIFF
--- a/Datagrid/ProxyQuery.php
+++ b/Datagrid/ProxyQuery.php
@@ -11,6 +11,7 @@
 
 namespace Sonata\DoctrineORMAdminBundle\Datagrid;
 
+use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
@@ -50,7 +51,7 @@ class ProxyQuery implements ProxyQueryInterface
      */
     public function __construct($queryBuilder)
     {
-        $this->queryBuilder      = $queryBuilder;
+        $this->queryBuilder = $queryBuilder;
         $this->uniqueParameterId = 0;
         $this->entityJoinAliases = array();
     }
@@ -90,7 +91,7 @@ class ProxyQuery implements ProxyQueryInterface
         $queryBuilderId = clone $queryBuilder;
 
         // step 1 : retrieve the targeted class
-        $from  = $queryBuilderId->getDQLPart('from');
+        $from = $queryBuilderId->getDQLPart('from');
         $class = $from[0]->getFrom();
         $metadata = $queryBuilderId->getEntityManager()->getMetadataFactory()->getMetadataFor($class);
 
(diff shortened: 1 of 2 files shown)


Datagrid/ProxyQuery.php
```diff
@@ -117,10 +117,17 @@ private function getFixedQueryBuilder(QueryBuilder $queryBuilder)
...
         $results    = $queryBuilderId->getQuery()->execute(array(), Query::HYDRATE_ARRAY);
+        $platform   = $queryBuilderId->getEntityManager()->getConnection()->getDatabasePlatform();
         $idxMatrix  = array();
         foreach ($results as $id) {
             foreach ($idNames as $idName) {
-                $idxMatrix[$idName][] = $id[$idName];
+                $phpValue = $id[$idName];
+                // Convert ids to database value in case of custom type
+                $fieldType = $metadata->getTypeOfField($idName);
+                $type = \Doctrine\DBAL\Types\Type::getType($fieldType);
```
Soullivaneuh: Namespaces should be imported.

ju1ius: :ok_hand:

Model/ModelManager.php
```diff
@@ -322,20 +323,27 @@ public function getIdentifierValues($entity)
...
-        foreach ($class->getIdentifierValues($entity) as $value) {
+        foreach ($metadata->getIdentifierValues($entity) as $name => $value) {
             if (!is_object($value)) {
                 $identifiers[] = $value;
                 continue;
             }
+            $fieldType = $metadata->getTypeOfField($name);
+            $type = Type::getType($fieldType);
+            if ($type) {
+                $identifiers[] = $type->convertToDatabaseValue($value, $platform);
```
ju1ius: @Soullivaneuh: This is needed to avoid:
>  RuntimeException: No entity manager defined for class Ramsey\Uuid\Uuid

when using child admin list of an admin whose entity has a custom identifier type.

adrienbrault:
This seems to fix my issue (I am also using the `ramsey/uuid-doctrine` package).

@ju1ius Can you squash your commits ?

ju1ius:
@adrienbrault @Soullivaneuh 
Squashed & rebased onto latest master.

OskarStark:
Can you please add some tests @ju1ius ?

:+1:

Soullivaneuh:

Regards.

Issue #472, opened 2015-06-24
$field in doctrine_orm_callback is empty


webdevilopers:
```php
use Sonata\CoreBundle\Form\Type\BooleanType;

class PersonAdmin extends Admin
{
    protected function configureDatagridFilters(DatagridMapper $filterMapper)
    {
        $filterMapper
            ->add('employmentContracts', 'doctrine_orm_callback', array(
                    'callback' => function($queryBuilder, $alias, $field, $value) {
                            if (!isset($value['value'])) {
                                return;
                            }

                            if ($value['value'] == BooleanType::TYPE_YES) {
//                                $queryBuilder->andWhere(sprintf('SIZE(%s.%s) > 0', $alias, $field));
                                $queryBuilder->andWhere(sprintf('SIZE(%s.%s) > 0', $alias, 'employmentContracts'));
                            }

                            if ($value['value'] == BooleanType::TYPE_NO) {
                                $queryBuilder->andWhere(sprintf('SIZE(%s.%s) = 0', $alias, 'employmentContracts'));
                            }

                            return true;
                        }
                    ), 'choice', array(
                        'translation_domain' => 'SonataCoreBundle',
                        'choices' => array(
                            BooleanType::TYPE_YES => 'label_type_yes',
                            BooleanType::TYPE_NO  => 'label_type_no'
                        )
                    )
                )
```

When the filter is applied `$field` will be empty. It has to be set manually.

Source:
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/master/Filter/CallbackFilter.php#L29-36

But when I comment out the `callback` method the error message correctly displays the content:
*Please provide a valid callback option "filter" for field "employmentContracts"*
Since it uses `$this->getName()`.

I guess `$this->getName()` should be passed to the `call_user_func` too, right?

core23:
This issue should be fixed in the latest master version. Please have a look at the docs.

If this is still present, feel free to reopen this issue

PR #482, opened 2015-08-20
Fix CallbackFilter alias and field variables

vchabot:
Hello,

I tried to use the `field` and `alias` param from the CallbackFilter callback, but I had some issues.

- The `field` param was always empty.
- The `alias` param was incorrect.

Example for incorrect `alias` param :
```php
{
    $datagridMapper
        ->add('contact.last_name', 'doctrine_orm_callback', array('callback' => array('MyClass', 'myStaticMethod')))
    ;
}
``` 

After some research, I found this issue: #472 (for the `field` param)

And here is the pull request to fix it :)

Thank you for your time,

Vincent

### DIFF
--- a/Filter/CallbackFilter.php
+++ b/Filter/CallbackFilter.php
@@ -20,7 +20,9 @@ class CallbackFilter extends Filter
      */
     protected function association(ProxyQueryInterface $queryBuilder, $data)
     {
-        return array($this->getOption('alias', $queryBuilder->getRootAlias()), false);
+        $alias = $queryBuilder->entityJoin($this->getParentAssociationMappings());
+
+        return array($this->getOption('alias', $alias), $this->getFieldName());
     }
 
     /**
--- a/Tests/Filter/CallbackFilterTest.php
+++ b/Tests/Filter/CallbackFilterTest.php
@@ -73,4 +73,26 @@ class CallbackFilterTest extends \PHPUnit_Framework_TestCase
 
         $filter->filter($builder, 'alias', 'field', 'myValue');
     }
+
+    public function testApplyMethod()
+    {
+        $builder = new ProxyQuery(new QueryBuilder());
+
+        $filter = new CallbackFilter();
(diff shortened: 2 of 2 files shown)


Filter/CallbackFilter.php
```diff
@@ -18,14 +18,6 @@ class CallbackFilter extends Filter
     /**
      * {@inheritdoc}
      */
-    protected function association(ProxyQueryInterface $queryBuilder, $data)
```
vchabot: I removed this method to let the `Filter` class handle this method.

Soullivaneuh: This is BC break. Please deprecated it if this method is not needed anymore.

Filter/CallbackFilter.php
```diff
@@ -20,7 +20,9 @@ class CallbackFilter extends Filter
      */
     protected function association(ProxyQueryInterface $queryBuilder, $data)
     {
-        return array($this->getOption('alias', $queryBuilder->getRootAlias()), false);
+        $alias = $queryBuilder->entityJoin($this->getParentAssociationMappings());
```
vchabot: I think with this change I will not break BC anymore. What do you think @Soullivaneuh ?

Filter/CallbackFilter.php
GromNaN: I don't get where is the BC break. This method is defined in the [`Filter`](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/master/Filter/Filter.php#L37) class. That was only the overloading of this method that was removed.

GromNaN: Got it. The difference is in the `$this->getOption('alias', $alias)`.

GromNaN:
:+1:

GromNaN:
@rande Can you please accept this PR ?

rande:
@vchabot can you squash the PR ? Thanks

GromNaN:
Thanks.

PR #496, opened 2015-10-23
Fixed phpdoc

core23:


Datagrid/Pager.php
```diff
@@ -21,8 +21,6 @@
  */
 class Pager extends BasePager
 {
-    protected $queryBuilder = null;
```
Soullivaneuh: Why removing this?

DependencyInjection/Compiler/AddAuditEntityCompilerPass.php
```diff
@@ -14,10 +14,6 @@
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
-/*
- *
- * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
```
Soullivaneuh: This should be here. Just remove the extra line.

DependencyInjection/Compiler/AddGuesserCompilerPass.php
```diff
@@ -15,10 +15,6 @@
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
 
-/*
- *
- * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
```
Soullivaneuh: This should be here. Just remove the extra line.

DependencyInjection/Compiler/AddTemplatesCompilerPass.php
```diff
@@ -15,10 +15,6 @@
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
 
-/*
- *
- * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
```
Soullivaneuh: This should be here. Just remove the extra line.

Admin/FieldDescription.php
```diff
@@ -16,7 +16,7 @@
 class FieldDescription extends BaseFieldDescription
 {
     /**
-     * {@inheritdoc}
+     * Constructor.
```
jvasseur: I would completely remove the phpdoc here, saying that __construct id the constructor doesn't add anything.

Soullivaneuh: Coding Style require phpdoc on each method AFAIK.

DependencyInjection/Compiler/AddAuditEntityCompilerPass.php
jvasseur: There is a ```*``` missing on the first line of the docblock that should be fixed to.

Soullivaneuh: Good catch @jvasseur!

Admin/FieldDescription.php
jvasseur: Which coding style ?

For symfony they don't add phpdoc where it doesn't had any meaningful information.

DependencyInjection/Compiler/AddAuditEntityCompilerPass.php
core23: Thought this is redundant because of the existing header of every single php file https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/master/DependencyInjection/Compiler/AddAuditEntityCompilerPass.php#L3-L10

Admin/FieldDescription.php
core23: **Add PHPDoc blocks for all classes, methods, and functions;**
http://symfony.com/doc/current/contributing/code/standards.html#documentation

IMO *all* means really all, there is no exception.

DependencyInjection/Compiler/AddAuditEntityCompilerPass.php
jvasseur: The header is juste here for the license and always refer to the same person.

The ```@author``` tag is where we get the actual author of the class.

Datagrid/Pager.php
core23: Cause it is never accessed. Could create a serperate PR if you want

Admin/FieldDescription.php
jvasseur: Sadly this page doesn't really reflect what is done in the symfony repository i've seen some pull request that add phpdoc to undocummented methods getting refused because they don't add any information.

An example is https://github.com/symfony/symfony/pull/16081

Datagrid/Pager.php
jvasseur: It's protected so it could be used by a child class. Removing it is a BC break.

Admin/FieldDescription.php
Soullivaneuh: I would prefer to strictly follow the documentation.

jvasseur: I opened a PR for the doc to match what is done on the repo : https://github.com/symfony/symfony-docs/pull/5838

core23:
@Soullivaneuh Should I leave the constructor comment as it is or is here anything to do?

Soullivaneuh:
Let's merge as is. Comment *may* be removed regarding the evolution of the symfony-docs PR.

PR #499, opened 2015-10-26
Removed unused property

core23:
This property is never used.


jvasseur:
This is a potential BC break since it could be accessed by a child class.

But i don't think anyone would have done that (and according to http://packanalyst.com/class?q=Sonata\DoctrineORMAdminBundle\Datagrid\Pager no one does).

OskarStark:
ping @Soullivaneuh

Soullivaneuh:
Don't know. @rande @pulzarraider, your opinion?

core23:
For BC convenience I added a deprection message.

PR #503, opened 2015-11-19
Embedded identifier support

mihai-stancu:
When declaring a field from an embedded entity as a table identifier
the ProxyQuery will use the fieldName as a parameter placeholder. This
triggers a semantical error in the query builder.

### DIFF
--- a/Datagrid/ProxyQuery.php
+++ b/Datagrid/ProxyQuery.php
@@ -143,6 +143,7 @@ class ProxyQuery implements ProxyQueryInterface
         foreach ($idxMatrix as $idName => $idx) {
             if (count($idx) > 0) {
                 $idxParamName = sprintf('%s_idx', $idName);
+                $idxParamName = preg_replace('/[^\w]+/', '_', $idxParamName);
                 $queryBuilder->andWhere(sprintf('%s IN (:%s)', $selects[$idName], $idxParamName));
                 $queryBuilder->setParameter($idxParamName, $idx);
                 $queryBuilder->setMaxResults(null);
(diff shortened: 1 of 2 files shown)


Datagrid/ProxyQuery.php
```diff
@@ -143,6 +143,7 @@ private function getFixedQueryBuilder(QueryBuilder $queryBuilder)
         foreach ($idxMatrix as $idName => $idx) {
             if (count($idx) > 0) {
                 $idxParamName = sprintf('%s_idx', $idName);
+                $idxParamName = str_replace('.', '_', $idxParamName);
```
mihai-stancu: We could use a preg_replace here to replace `/[^\w]/` with `_`.

OskarStark:
@mihai-stancu could you please add a test case for this bug fix?

Thank you

mihai-stancu:
@OskarStark sorry for the delay, here's the test to cover this bugfix.

OskarStark:
Thank you!

Could you please fix the CS here:
```
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
    //...
   2) Tests/Datagrid/ProxyQueryTest.php (unused_use, extra_empty_lines)
```

Datagrid/ProxyQuery.php
OskarStark: :+1: preg_replace is much better IMO

mihai-stancu:
I assumed running `php-cs-fixer fix` would remove unused `use` statements -- forgot to check `.php_cs` to see if the rule is there.

Datagrid/ProxyQuery.php
mihai-stancu: Done.

OskarStark:
Thank you @mihai-stancu

PR #506, opened 2015-12-09
Bugfix StringFilter

rapemer:
Bugfix: Warning: trim() expects parameter 1 to be string, array given

Sample:

```php
    protected function configureDatagridFilters(DatagridMapper $filter)
    {
        $filter
            ->add('name')
            ->add('region', null, [], 'choice', [
                'label' => 'label.cat.region',
                'choices' => Cat::$regions,
                'expanded' => true,
                'multiple' => true
            ])
        ;
    }
```

### DIFF
--- a/Filter/StringFilter.php
+++ b/Filter/StringFilter.php
@@ -25,6 +25,10 @@ class StringFilter extends Filter
             return;
         }
 
+        if (is_array($data['value'])) {
+            $data['value'] = array_shift($data['value']);
+        }
+
         $data['value'] = trim($data['value']);
 
         if (strlen($data['value']) == 0) {


OskarStark:
ping @rande

greg0ire:
Please rebase on the tip of master so that tests pass.

And please improve your commit message. Half the commits are bugfixes.

rapemer:
@greg0ire sorry, but I don't know how ;-)

greg0ire:
No problem! Assuming you already are on your branch : 

1. Get the latest commits : `git fetch origin` (assuming `origin` refers to the main repo, and not your fork, you can check that with `git remote -vvv`).
2. Rebase on the tip of origin/master : `git rebase origin/master`
3. Force push the result on your fork : `git push --force`

Regarding the commit message, as I said, half the commits are bugfixes, so starting your commit message with "bugfix", is not very helpful. Plus, fixing a bug is a consequence of what you did (hopefully). The first thing you should clarify : what did you do ? After reading your commit, it looks like you made the StringFilter compatible with array data. The second thing you should clarify : why did you do it ? What situation lead you there ?

So the commit message might look like this : 

```
make the StringFilter compatible with array data

When using the StringFilter with multiple choices, there is an error message caused by an error in string filter.
```

Note that the "why" part raises a question : why the hell is `StringFilter` called instead of `ChoiceFilter` ? And that is why you should write good commit messages.

Soullivaneuh:

Regards.

PR #508, opened 2015-12-22
[Translation] Add delete checkbox translation in collection

matthv:
On default behavior (inline, not table), the `action_delete` translation was not included.
Tested on local machine, everything seems working fine.


nicovak:
+1

OskarStark:
Thank you @matthv

PR #511, opened 2015-12-26
Always sort

### LINKED ISSUE #507: Make sure the sort always gives the same outcome
When sorting with a non unique value like a creation date, for instance, two successive requests can give different outcomes (in our example, when two objects have been created at the same seconds), which can lead to some nasty bugs. I think adding a default, extra sort by primary key could be the solution.
I can do a P.R. if you agree.

greg0ire:
If no ORDER BY query is specified, the order of elements cannot be
guaranteed between executions. See
http://stackoverflow.com/questions/1793147/sql-best-practice-to-deal-with-default-sort-order
This change finds out the identifier keys of the main entity at hand,
and adds a sort based on these keys.
Fixes #507 .

### DIFF
--- a/Datagrid/ProxyQuery.php
+++ b/Datagrid/ProxyQuery.php
@@ -74,6 +74,25 @@ class ProxyQuery implements ProxyQueryInterface
             $queryBuilder->resetDQLPart('orderBy');
         }
 
+        /* By default, always add a sort on the identifier fields of the first
+         * used entity in the query, because RDBMS do not guarantee a
+         * particular order when no ORDER BY clause is specified, or when
+         * the field used for sorting is not unique.
+         */
+
+        $identifierFields = $queryBuilder
+            ->getEntityManager()
+            ->getMetadataFactory()
+            ->getMetadataFor(current($queryBuilder->getRootEntities()))
+            ->getIdentifierFieldNames();
+
+        foreach ($identifierFields as $identifierField) {
+            $queryBuilder->addOrderBy(
+                $queryBuilder->getRootAlias().'.'.$identifierField,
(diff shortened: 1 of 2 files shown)


Datagrid/ProxyQuery.php
```diff
@@ -74,6 +74,24 @@ public function execute(array $params = array(), $hydrationMode = null)
             $queryBuilder->resetDQLPart('orderBy');
         }
 
+        /* By default, always add a sort on the identifier fields of the first
+            * used entity in the query, because RDBMS do not guarantee a
```
OskarStark: indention looks weird

greg0ire: Doesn't it? Vim does this when using `gqq` I was wondering if it was a standard or something, but since it looks weird to you too, I guess it probably is not. Anyway, I won't  be able to changr that soon because of holidays, sorry.

OskarStark: thank you @greg0ire , i would use 

```
// foo
// bar
etc...
```

greg0ire: You'd use single line style for a multiline comment?

OskarStark: i think i saw something like this, but i'm not sure :smile: 

ping @rande

greg0ire: I changed the comment to something more reasonable. As per the `//`, is is best suited for when your comment is composed of things that are independent and could thus be moved around, but here, I have a long sentence, so…

greg0ire:
Seems to fail because of newly added tests… I'll have a look at that later.

greg0ire:
Closing until then.

Issue #485, opened 2015-09-13
Symfony 2.8: Use FQCN as field names


WouterJ:
This bundle currently relies on field names to build the form (for instance, `FormContractor#getDefaultOptions`). As of Symfony 2.8, field names are deprecated and the FQCN has to be used instead.

rande:
This has been resolved by adding a compatibility layer with the SonataCoreBundle

PR #513, opened 2015-12-29
Replace 'form' field type with FQCN to silence Symfony >=2.8 deprecation notices

tehplague:


tehplague:
Same situation as in #485

rande:
@tehplague see comment on the issue related issue.

PR #517, opened 2016-01-05
Always sort

greg0ire:


Datagrid/ProxyQuery.php
```diff
@@ -63,17 +63,38 @@ public function execute(array $params = array(), $hydrationMode = null)
...
+         */
+
+        $identifierFields = $queryBuilder
+            ->getEntityManager()
+            ->getMetadataFactory()
+            ->getMetadataFor(current($queryBuilder->getRootEntities()))
+            ->getIdentifierFieldNames();
+
+        foreach ($identifierFields as $identifierField) {
+            $queryBuilder->addOrderBy(
```
Soullivaneuh: What happen if the querybuilder already have an orderBy?

I think this will be appended, not replaced.

Soullivaneuh:

Regards.

Datagrid/ProxyQuery.php
greg0ire: Yes, and it is intended.

greg0ire: For instance, imagine you have a sort on the date, but two people happen to post something at the same exact second. You don't want the request to return things in a different order when you press F5, and a good way to achieve that is to append a sort on a unique key

Soullivaneuh: Yes but what about `$this->getSortOrder()`?

greg0ire: As said in the comment "reusing the sort order is the most natural way to go". If you sort by date ASC, and then by id ASC, and you press the button that inverts the sort order, you expect items that have the same date to appear in reverse order. Reusing `getSortOrder` allows that.

greg0ire:
See #544

PR #519, opened 2016-01-19
change ProxyQuery::getFixedQueryBuilder to protected

marcoslibre:
I needed to extend ProxyQuery in order to add query hints in the execute() method.
Does it make sense to make the getFixedQueryBuilder method protected instead of private ?


OskarStark:
Thank you @marcoslibre

PR #523, opened 2016-01-31
Get database value of identifier to support other identifier type classes

acrobat:
I came across this issue/case when using an entity with uuid identifier and prefilled (uuid generated in the entity) identifier

Example:
```php
use Ramsey\Uuid\Uuid;

class Example
{
    /**
     * @var \Ramsey\Uuid\Uuid
     *
     * @ORM\Column(type="uuid")
     * @ORM\Id
     */
    private $id;

    public function __construct()
    {
        $this->id = Uuid::uuid4();
    }
}
```

Problems:
- Create page could not be opened (as the identifier class could not be resolved)
- prefilled identifier caused the code to think it was editing an existing item

Fixes:
- Use the database value of each type to display the entity
- Check with the unit of work if an entity is managed or not (to check if it is create or edit)
- I had to use the old uuid lib from Rhumsaa instead of Ramsey's to keep the tests compatible with php 5.3

I've also added tests to check this behavior. Based this pr on changes in #481.

### DIFF
--- a/Datagrid/ProxyQuery.php
+++ b/Datagrid/ProxyQuery.php
@@ -11,6 +11,7 @@
 
 namespace Sonata\DoctrineORMAdminBundle\Datagrid;
 
+use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
@@ -132,10 +133,17 @@ class ProxyQuery implements ProxyQueryInterface
         }
 
         $results    = $queryBuilderId->getQuery()->execute(array(), Query::HYDRATE_ARRAY);
+        $platform = $queryBuilderId->getEntityManager()->getConnection()->getDatabasePlatform();
         $idxMatrix  = array();
         foreach ($results as $id) {
             foreach ($idNames as $idName) {
-                $idxMatrix[$idName][] = $id[$idName];
+                $phpValue = $id[$idName];
+                // Convert ids to database value in case of custom type
+                $fieldType = $metadata->getTypeOfField($idName);
+                $type = Type::getType($fieldType);
+                $dbValue = $type->convertToDatabaseValue($phpValue, $platform);
+
+                $idxMatrix[$idName][] = $dbValue;
             }
         }
 
--- a/Tests/Datagrid/ProxyQueryTest.php
+++ b/Tests/Datagrid/ProxyQueryTest.php
@@ -11,15 +11,25 @@
 
 namespace Sonata\DoctrineORMAdminBundle\Tests\Datagrid;
 
+use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Query\Expr\From;
+use Rhumsaa\Uuid\Doctrine\UuidType;
 
 class ProxyQueryTest extends \PHPUnit_Framework_TestCase
 {
+    public static function setUpBeforeClass()
+    {
+        if (!Type::hasType('uuid')) {
+            Type::addType('uuid', 'Rhumsaa\Uuid\Doctrine\UuidType');
+        }
+    }
+
     public function dataGetFixedQueryBuilder()
     {
         return array(
-            array('aaa', 'bbb', 'id', 'id_idx', 33),
-            array('aaa', 'bbb', 'id.value', 'id_value_idx', 33),
+            array('aaa', 'bbb', 'id', 'id_idx', 33, Type::INTEGER),
+            array('aaa', 'bbb', 'id.value', 'id_value_idx', 33, Type::INTEGER),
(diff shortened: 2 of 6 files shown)


acrobat:
I don't think the test failures are related to this change

acrobat:
@rande , @OskarStark, @Soullivaneuh do you have time to review (and merge if ok) the changes? Thanks!

acrobat:
@rande , @OskarStark, @Soullivaneuh do you have time to review the changes? Thanks!

greg0ire:
Please follow these guidelines for your commit message http://rakeroutes.com/blog/deliberate-git/ (in particular, keep your commit message subject under 50 chars).

Tests/Datagrid/ProxyQueryTest.php
```diff
@@ -47,12 +60,27 @@ public function testGetFixedQueryBuilder($class, $alias, $id, $expectedId, $valu
             ->with($this->equalTo($class))
             ->willReturn($meta);
 
+        $platform = $this->getMockBuilder('Doctrine\DBAL\Platforms\PostgreSqlPlatform')
+            ->disableOriginalConstructor()
+            ->getMock();
+        //_getCommonIntegerTypeDeclarationSQL, Doctrine\DBAL\Platforms\AbstractPlatform::initializeDoc/trineTypeMappings
```
greg0ire: No commented code please.

composer.json
```diff
@@ -30,7 +30,8 @@
     },
     "require-dev": {
         "simplethings/entity-audit-bundle": "~0.1",
-        "symfony/phpunit-bridge": "~2.7|~3.0"
+        "symfony/phpunit-bridge": "~2.7|~3.0",
+        "rhumsaa/uuid": "^2.8"
```
greg0ire: Please sort packages alphabetically.

acrobat:
@greg0ire Thanks for the comments! Commit message length is fixed!

Tests/Datagrid/ProxyQueryTest.php
acrobat: Hmm missed it when cleaning up after writing, fixed!

composer.json
acrobat: Fixed!

composer.json
```diff
@@ -29,6 +29,7 @@
         "sonata-project/core-bundle": "~2.3,>=2.3.1"
     },
     "require-dev": {
+        "rhumsaa/uuid": "^2.8",
```
rande: this package is deprecated; also I don't see why we need this for testing

acrobat: I had to use the old uuid lib from Rhumsaa instead of Ramsey's to keep the tests compatible with php 5.3

I use the lib because I use a uuid class to generate uuid (and have it mapped by doctrine) to test the changes.

Datagrid/ProxyQuery.php
```diff
@@ -132,10 +133,17 @@ protected function getFixedQueryBuilder(QueryBuilder $queryBuilder)
...
+        $platform = $queryBuilderId->getEntityManager()->getConnection()->getDatabasePlatform();
         $idxMatrix  = array();
         foreach ($results as $id) {
             foreach ($idNames as $idName) {
-                $idxMatrix[$idName][] = $id[$idName];
+                $phpValue = $id[$idName];
+                // Convert ids to database value in case of custom type
+                $fieldType = $metadata->getTypeOfField($idName);
+                $type = Type::getType($fieldType);
+                $dbValue = $type->convertToDatabaseValue($phpValue, $platform);
```
rande: This is weird, doctrine should perform the work in the query mapper.

composer.json
greg0ire: Maybe you could hardcode what you know to be a valid uuid instead.

acrobat: I could do that, but i'm also checking if the [correct type is returned](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/523/files#diff-304847c39531513d83f2682805eb7b6cR321)

Or should I mock the whole uuid class and doctrine mapping etc?

greg0ire: Writing a dummy `UuidType` should be enough, shouldn't it?

acrobat: The UuidType and Uuid object to generate a uuid and have it set in the entity so doctrine knows what to convert. I will try to implement the extra mock objects and remove the uuid package dependency

Datagrid/ProxyQuery.php
acrobat: Doctrine will return an object for the identifier (in my case an object of type Uuid). But further on in the code it expects a string value to display and use in urls. Therefor getting the db value solves to problem. It's the code from PR #481 but i've added the missing tests and some code cleanup.

OskarStark: ping @rande

acrobat:
Can someone get me on the way on how to mock the ramsey/uuid and ramsey/uuid-doctrine-type class? As I also need to access it in the fixture entity, but don't how I should really mock these classes

Soullivaneuh:

Regards.

PR #527, opened 2016-03-17
Check the entity identifier attribute is named id

jmontoyaa:
In some cases (legacy code) the id is not set, this fixes the error.

"Key "id" for array with keys "iid" does not exist in SonataDoctrineORMAdminBundle:Block:block_audit.html.twig at line 41"

### DIFF
--- a/Resources/views/Block/block_audit.html.twig
+++ b/Resources/views/Block/block_audit.html.twig
@@ -36,7 +36,11 @@ file that was distributed with this source code.
                                     {% for changedEntity in revision.entities %}
                                         <li>
                                             {{ changedEntity.entity }} / {{ changedEntity.revisionType }}
-                                            / {{ changedEntity.className }} - {{ changedEntity.id.id }}
+                                            / {{ changedEntity.className }} 
+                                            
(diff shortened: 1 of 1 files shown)


OskarStark:
LGTM

ping @greg0ire

greg0ire:
Can you change your commit message? The `Add "is defined"` part is too low level. Maybe

> Check the entity identifier attribute is named id

would be better, because it raises some more questions : what if it isn't ?

Also, what is this `iid` thing about? Maybe it is the legacy way of doing things and it contains the identifier, and we should actually display it instead? Is it under `changedEntity.id` or under `changedEntity` directly ?

greg0ire:
> Also, what is this iid thing about?

Looks like [some people](http://classic.chamilo.googlecode.com/hg/main/inc/Entity/EntityCGroupInfo.php) actually use that in their code.

@jmontoyaa : why not be nice and display it instead, then ? And in the process, support composite keys, like this : `{{ changedEntity.id|join(', ')}}`. Can you try this?

Soullivaneuh:

Regards.

PR #530, opened 2016-04-09
Replaced accessing type by string name

toooni:
Accessing type "sonata_type_boolean" by its string name is deprecated since Symfony 2.8 and has been removed in 3.0.

### DIFF
--- a/Guesser/FilterTypeGuesser.php
+++ b/Guesser/FilterTypeGuesser.php
@@ -13,6 +13,7 @@ namespace Sonata\DoctrineORMAdminBundle\Guesser;
 
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Sonata\AdminBundle\Model\ModelManagerInterface;
+use Sonata\CoreBundle\Form\Type\BooleanType;
 use Symfony\Component\Form\Guess\Guess;
 use Symfony\Component\Form\Guess\TypeGuess;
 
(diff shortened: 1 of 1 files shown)


Guesser/FilterTypeGuesser.php
```diff
@@ -65,7 +66,7 @@ public function guessType($class, $property, ModelManagerInterface $modelManager
 
         switch ($metadata->getTypeOfField($propertyName)) {
             case 'boolean':
-                $options['field_type']    = 'sonata_type_boolean';
+                $options['field_type']    = BooleanType::class;
```
core23: This works for symfony 2.3 too? @toooni

toooni: @core23 Sorry I didn't know that the master branch still should preserve BC compatibility with sf 2.3. I don't think it's working on 2.3, but it works with 3.0+. Should I change the target branch for the PR? Or how do you achieve 3.0+ support?

core23:
Until now, we support both versions https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/master/composer.json#L25

toooni:
@core23 then this feature (self guessing boolean type in filter) does not work in sf 3 yet.

Guesser/FilterTypeGuesser.php
greg0ire: We use an if statement that detect the current symfony form version, using the presence of the `getBlockPrefix` method.

Soullivaneuh:

Regards.

PR #535, opened 2016-04-26
do not test dependencies

greg0ire:
A unit test should not test methods defined outside the class under
test, even if it is a parent class, and especially if this class is
defined in another package.


greg0ire:
ping @OskarStark @core23

Issue #205, opened 2013-04-19
sonata_type_collection doesn't pick up "class" attr


tiagojsag:
Hi,

I have several sonata_type_collection fields in a single admin form, and I need different classes in it. I tried using the usual 'attr' => array('class' => 'foo') but it's not being rendered by the corresponding twig from SonataDoctrineORMAdmin

cheers

core23:
This issue is very old. Can you please check this against the latest master version? 

Feel free to reopen if this error still exists.

Mattabile:
The issue already exists :  "sonata-project/admin-bundle": "^3.4",

PR #536, opened 2016-04-28
Fix for #205 sonata_type_collection doesn't pick up class attr

sujayjaju:
Class attribute added to the field for sonata_type_collection

### DIFF
--- a/Resources/views/CRUD/edit_orm_many_to_many.html.twig
+++ b/Resources/views/CRUD/edit_orm_many_to_many.html.twig
@@ -10,7 +10,7 @@ file that was distributed with this source code.
 #}
 {% if sonata_admin.field_description.hasassociationadmin %}
     <div id="field_container_{{ id }}"
-		class="field-container {{ form.vars["attr"]["class"] is defined ? form.vars["attr"]["class"] }}">
+        class="field-container {{ form.vars["attr"]["class"] is defined ? form.vars["attr"]["class"] }}">
         <span id="field_widget_{{ id }}" >
             {% if sonata_admin.edit == 'inline' %}
                 {% if sonata_admin.inline == 'table' %}
--- a/Resources/views/CRUD/edit_orm_many_to_many.html.twig
+++ b/Resources/views/CRUD/edit_orm_many_to_many.html.twig
@@ -9,7 +9,7 @@ file that was distributed with this source code.
 
 #}
 {% if sonata_admin.field_description.hasassociationadmin %}
-    <div id="field_container_{{ id }}" class="field-container">
+    <div id="field_container_{{ id }}" class="field-container {{ form.vars["attr"]["class"] is defined ? form.vars["attr"]["class"] }}">
         <span id="field_widget_{{ id }}" >
             {% if sonata_admin.edit == 'inline' %}
                 {% if sonata_admin.inline == 'table' %}
--- a/Resources/views/CRUD/edit_orm_many_to_many.html.twig
+++ b/Resources/views/CRUD/edit_orm_many_to_many.html.twig
@@ -10,7 +10,7 @@ file that was distributed with this source code.
 #}
 {% if sonata_admin.field_description.hasassociationadmin %}
     <div id="field_container_{{ id }}"
-        class="field-container {{ form.vars["attr"]["class"] is defined ? form.vars["attr"]["class"] }}">
+        class="field-container {{ form.vars['attr']['class'] is defined ? form.vars['attr']['class'] }}">
         <span id="field_widget_{{ id }}" >
             {% if sonata_admin.edit == 'inline' %}
                 {% if sonata_admin.inline == 'table' %}
--- a/Resources/views/CRUD/edit_orm_many_to_many.html.twig
+++ b/Resources/views/CRUD/edit_orm_many_to_many.html.twig
@@ -9,7 +9,8 @@ file that was distributed with this source code.
 
 #}
 {% if sonata_admin.field_description.hasassociationadmin %}
-    <div id="field_container_{{ id }}" class="field-container {{ form.vars["attr"]["class"] is defined ? form.vars["attr"]["class"] }}">
+    <div id="field_container_{{ id }}"
+		class="field-container {{ form.vars["attr"]["class"] is defined ? form.vars["attr"]["class"] }}">
         <span id="field_widget_{{ id }}" >
             {% if sonata_admin.edit == 'inline' %}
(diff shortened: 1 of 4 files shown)


Resources/views/CRUD/edit_orm_many_to_many.html.twig
```diff
@@ -9,7 +9,7 @@ file that was distributed with this source code.
 
 #}
 {% if sonata_admin.field_description.hasassociationadmin %}
-    <div id="field_container_{{ id }}" class="field-container">
+    <div id="field_container_{{ id }}" class="field-container {{ form.vars["attr"]["class"] is defined ? form.vars["attr"]["class"] }}">
```
greg0ire: This makes the lines too long, please linebreak this.

sujayjaju: There are longer lines in the same file - is there a standard to what the length of a line should be in terms of characters?

Will this PR get blocked because the line looks long?

greg0ire: > There are longer lines in the same file

Yes, and that's sad. We're trying to fix that gradually.

For php, there is a standard : 

> There MUST NOT be a hard limit on line length; the soft limit MUST be 120 characters; lines SHOULD be 80 characters or less.

So as you see, it's not a hard rule, just a recommendation, because there might be some cases where this is impossible to fix (I don't know which, but…)

> Will this PR get blocked because the line looks long?

I hope so. We should stop with that crap. We already have to scroll vertically, if we have to also scroll horizontally, what's next? Scrolling in the third dimension ?

greg0ire: ping @sujayjaju

sujayjaju: Updated the formatting

OskarStark:
could you add some tests or extend existing ones? @sujayjaju

Resources/views/CRUD/edit_orm_many_to_many.html.twig
```diff
@@ -9,7 +9,8 @@ file that was distributed with this source code.
 
 #}
 {% if sonata_admin.field_description.hasassociationadmin %}
-    <div id="field_container_{{ id }}" class="field-container">
+    <div id="field_container_{{ id }}"
+		class="field-container {{ form.vars["attr"]["class"] is defined ? form.vars["attr"]["class"] }}">
```
OskarStark: what about `'` instead of `"` ? @greg0ire

greg0ire: Using `'` would be better

Resources/views/CRUD/edit_orm_many_to_many.html.twig
```diff
@@ -9,7 +9,8 @@ file that was distributed with this source code.
 
 #}
 {% if sonata_admin.field_description.hasassociationadmin %}
-    <div id="field_container_{{ id }}" class="field-container {{ form.vars["attr"]["class"] is defined ? form.vars["attr"]["class"] }}">
+    <div id="field_container_{{ id }}"
+		class="field-container {{ form.vars["attr"]["class"] is defined ? form.vars["attr"]["class"] }}">
```
greg0ire: It is way better like this ! Can you align `class` with `id` though ?

sujayjaju: I typically indent attributes on a new line by "1 tab" - aligning it to id would mean that we are aligning it by the length of the container tag. So it would be different for "div" and different for say "textarea" ... I think the issue is that the rest of the file is indented by spaces, while that line is indented by tabs.

Resources/views/CRUD/edit_orm_many_to_many.html.twig
sujayjaju: Enclose all attributes by `'` instead of `"`? Why?

Resources/views/CRUD/edit_orm_many_to_many.html.twig
greg0ire: O_o didn't notice that! Don't mix tabs and spaces, use only tabs please!

sujayjaju: I use tabs, the rest of the file uses spaces. I just added a line

Soullivaneuh: We use spaces under Sonata. Please replace your tabs.

Resources/views/CRUD/edit_orm_many_to_many.html.twig
greg0ire: `""` should be used only when string interpolation is needed (it helps noticing it when all the rest of the code uses single quotes as enclosures)

Resources/views/CRUD/edit_orm_many_to_many.html.twig
greg0ire: There is even a comic for that one :

for the lulz

Resources/views/CRUD/edit_orm_many_to_many.html.twig
sujayjaju: Sorry, I don't quite follow.
Do you need it to be:
` form.vars['attr']['class'] is defined ? form.vars['attr']['class']`
or
`class='field-container {{ form.vars["attr"]["class"] is defined ? form.vars["attr"]["class"] }}'`
?

greg0ire: The first one please.

Resources/views/CRUD/edit_orm_many_to_many.html.twig
```diff
@@ -9,7 +9,8 @@ file that was distributed with this source code.
 
 #}
 {% if sonata_admin.field_description.hasassociationadmin %}
-    <div id="field_container_{{ id }}" class="field-container">
+    <div id="field_container_{{ id }}"
+        class="field-container {{ form.vars['attr']['class'] is defined ? form.vars['attr']['class'] }}">
```
greg0ire: Fair point for the "container width" argument.

greg0ire:
LGTM

Resources/views/CRUD/edit_orm_many_to_many.html.twig
core23: The ``>`` should be on a new line

Soullivaneuh:

Regards.

PR #545, opened 2016-05-10
change StringFilter::getOperator to protected

### LINKED ISSUE #515: StringFilter : getOperator should be private
I'd like to extend StringFilter in order to use the "ILIKE" operator unstead of the "LIKE" operator (I'm using Postgresql and I'd like the string filter to be case insensitive).
The problem is that the getOperator() method is protected.
It would make sense to change it to private so we can override it without having to override the filter() method.

OskarStark:
Fixes #515

### DIFF
--- a/Filter/StringFilter.php
+++ b/Filter/StringFilter.php
@@ -55,7 +55,7 @@ class StringFilter extends Filter
      *
      * @return bool
      */
-    private function getOperator($type)
+    protected function getOperator($type)
     {
         $choices = array(
             ChoiceType::TYPE_CONTAINS => 'LIKE',
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,3 +4,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
+
+### Changed
+- StringFilter::getOperator to protected instead of private


Filter/StringFilter.php
```diff
@@ -55,7 +55,7 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
      *
      * @return bool
      */
-    private function getOperator($type)
+    protected function getOperator($type)
```
greg0ire: :-1: , maybe lines 61 and 62 could be dynamic?

Soullivaneuh: I'm not a big fan of this, because this does not concern only `LIKE`.

Why not adding methods like:

* `getTypeContainsOperator`
* `getTypeNotContainsOperator`
* `getTypeEqualOperator`

And call them from `getOperator`?

For #515, @marcoslibre could override `getTypeContainsOperator` and specify `ILIKE` on it.

greg0ire: Let's state the requirements cleary : the string filter should be case insensitive by default (because who wants case sensitive filtering ?), which means different operators should be injected into it, and implies some sort of guesser does this injection base on the current database platform (given by `Connection::getDatabasePlatform()`). Options to also have case sensitive filters might be added on another PR.

greg0ire: I did something like that on another bundle : https://github.com/lexik/LexikFormFilterBundle/pull/91/files

greg0ire: Also, one could imagine that there are several RDBMS in use in the project, so there should either be several instances of the String filter, or the string filter should be able to behave differently based on the filtered entity (the entity manager should be found, and the connection retrieved from it)

greg0ire: There is an even easier way than injection : `$proxyQuery->getEntityManager()->getConnection()->getDatabasePlatform()`. What do you think @Soullivaneuh @OskarStark ?

Soullivaneuh: @greg0ire Sounds good. Do you want to propose a POC?

greg0ire: I'll work on it :)

greg0ire: Actually what I did in the link above is use strtolower on both operands, which is suboptimal because an index on the field cannot be used, and function based indexes in doctrine support is unsure. Related : https://github.com/doctrine/dbal/issues/980

greg0ire: > Sadly, ILIKE does not seem to work in DQL : line 0, col 226: Error: Expected =, <, <=, <>, >, >=, !=, got 'ILIKE' 

@marcoslibre : did you test editing the file and using `ILIKE`?

Soullivaneuh: @greg0ire Did you test with PostreSQL?

greg0ire: Yes I did, `ILIKE` works in `psql`, but is just not a valid DQL keyword.

greg0ire: IMO it's more of a doctrine bug. Doctrine should use `ILIKE` when I write `LIKE` in DQL.

greg0ire:
Closing since making the method protected won't even allow @marcoslibre to solve his problem (`ILIKE` is not recognized).

PR #547, opened 2016-05-12
Get database value of identifier to support other identifier type classes

acrobat:
_Resubmitted PR #523 to the correct branch after switching release strategy_

Description from #523:

> 
> I came across this issue/case when using an entity with uuid identifier and prefilled (uuid generated in the entity) identifier
> 
> Example:
> ```php
> use Ramsey\Uuid\Uuid;
> 
> class Example
> {
>     /**
>      * @var \Ramsey\Uuid\Uuid
>      *
>      * @ORM\Column(type="uuid")
>      * @ORM\Id
>      */
>     private $id;
> 
>     public function __construct()
>     {
>         $this->id = Uuid::uuid4();
>     }
> }
> ```
> 
> Problems:
> - Create page could not be opened (as the identifier class could not be resolved)
> - prefilled identifier caused the code to think it was editing an existing item
> 
> Fixes:
> - Use the database value of each type to display the entity
> - Check with the unit of work if an entity is managed or not (to check if it is create or edit)
> - I had to use the old uuid lib from Rhumsaa instead of Ramsey's to keep the tests compatible with php 5.3
> 
> I've also added tests to check this behavior. Based this pr on changes in #481.

Done:
- [x] Imported fixes from #481

Todo:
- [x] Rework tests so they don't rely on rhumsaa/uuid
- [x] Add changelog

Can someone help me a bit to get started on the mock of the uuid class and uuid doctirne type for the unittests?

### DIFF
--- a/Model/ModelManager.php
+++ b/Model/ModelManager.php
@@ -14,6 +14,7 @@ namespace Sonata\DoctrineORMAdminBundle\Model;
 use Doctrine\Common\Util\ClassUtils;
 use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\LockMode;
+use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\OptimisticLockException;
@@ -338,19 +339,28 @@ class ModelManager implements ModelManagerInterface, LockInterface
         //    throw new \RuntimeException('Entities passed to the choice field must be managed');
         //}
 
-        $class = $this->getMetadata(ClassUtils::getClass($entity));
+        $class = ClassUtils::getClass($entity);
+        $metadata = $this->getMetadata($class);
+        $platform = $this->getEntityManager($class)->getConnection()->getDatabasePlatform();
 
         $identifiers = array();
 
-        foreach ($class->getIdentifierValues($entity) as $value) {
+        foreach ($metadata->getIdentifierValues($entity) as $name => $value) {
             if (!is_object($value)) {
                 $identifiers[] = $value;
                 continue;
             }
 
-            $class = $this->getMetadata(ClassUtils::getClass($value));
+            $fieldType = $metadata->getTypeOfField($name);
+            $type = Type::getType($fieldType);
+            if ($type) {
+                $identifiers[] = $type->convertToDatabaseValue($value, $platform);
+                continue;
+            }
 
-            foreach ($class->getIdentifierValues($value) as $value) {
+            $metadata = $this->getMetadata(ClassUtils::getClass($value));
+
+            foreach ($metadata->getIdentifierValues($value) as $value) {
                 $identifiers[] = $value;
             }
         }
@@ -376,7 +386,7 @@ class ModelManager implements ModelManagerInterface, LockInterface
         }
 
         // the entities is not managed
-        if (!$entity /*|| !$this->getEntityManager($entity)->getUnitOfWork()->isInIdentityMap($entity) // commented for perfomance concern */) {
+        if (!$entity || !$this->getEntityManager($entity)->getUnitOfWork()->isInIdentityMap($entity)) {
             return;
         }
 
--- a/Tests/Model/ModelManagerTest.php
+++ b/Tests/Model/ModelManagerTest.php
@@ -11,18 +11,29 @@
 
 namespace Sonata\DoctrineORMAdminBundle\Tests\Model;
 
+use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\OptimisticLockException;
 use Doctrine\ORM\Version;
 use Sonata\DoctrineORMAdminBundle\Admin\FieldDescription;
 use Sonata\DoctrineORMAdminBundle\Model\ModelManager;
+use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\DoctrineType\UuidType;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\AssociatedEntity;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\ContainerEntity;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\Embeddable\EmbeddedEntity;
+use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\UuidEntity;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\VersionedEntity;
+use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Util\NonIntegerIdentifierTestClass;
 
 class ModelManagerTest extends \PHPUnit_Framework_TestCase
 {
+    public static function setUpBeforeClass()
+    {
+        if (!Type::hasType('uuid')) {
+            Type::addType('uuid', 'Sonata\DoctrineORMAdminBundle\Tests\Fixtures\DoctrineType\UuidType');
+        }
+    }
+
     public function testSortParameters()
     {
         $registry = $this->getMock('Symfony\Bridge\Doctrine\RegistryInterface');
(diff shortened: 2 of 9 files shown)


OskarStark:
@acrobat could you please add a changelog note? Thank yoU!

acrobat:
I've add the changelog entry and I think I found a way to get the tests done!

acrobat:
PR should be done

- Rebased agianst the latest master
- Fixed some extra style issues after rebase
- Reworked tests and removed the `rhumsaa/uuid` dependency in the tests
- Added the changelog entry

Tests/Model/ModelManagerTest.php
```diff
@@ -296,3 +362,35 @@ private function getMetadata($class, $isVersioned)
         return $metadata;
     }
 }
+
+class Uuid
```
Soullivaneuh: This class should be on a separated file under `Fixtures` folder.

And if possible, use a better name that show it's a class only for test.

Soullivaneuh:
@OskarStark As a bug fix, the label should be `patch`.

Soullivaneuh:
@acrobat Could you please add class comment with `@author` tag on new class introduced on this PR?

Soullivaneuh:
Documentation is also needed for this case.

acrobat:
@Soullivaneuh I've addressed the comments you had. What kind of documention do you expect? Little example of an entity that uses an uuid or just a small comment that we also support those classes?

Soullivaneuh:
Just some line specifying uuid is supported with a tiny example like you put on your PR would be great.

Maybe @sonata-project/contributors have some other suggestions?

acrobat:
@Soullivaneuh I've added a small note including a sample entity with uuid identifier in `list_field_definition.rst`. Don't know if that is the correct place to show the note. Maybe you have a better place to show note?

acrobat:
@OskarStark @Soullivaneuh  I've rebased the branch so it's updated again with master, anything else needed here?

Model/ModelManager.php
```diff
@@ -376,7 +386,7 @@ public function getNormalizedIdentifier($entity)
         }
 
         // the entities is not managed
-        if (!$entity /*|| !$this->getEntityManager($entity)->getUnitOfWork()->isInIdentityMap($entity) // commented for perfomance concern */) {
+        if (!$entity || !$this->getEntityManager($entity)->getUnitOfWork()->isInIdentityMap($entity)) {
```
Soullivaneuh: This line was "commented for perfomance concern".

It would be great to know who did that and why before merging.

Soullivaneuh: BTW, why did you need that?

Soullivaneuh:
Except my comment, all is good for me.

Model/ModelManager.php
acrobat: This check is needed because I set the uuid in php code (in the constructor) but that way sonata thinks it's an existing entity and goes to the edit screen, but it should check doctrine if the entity exists and decised that way if it is a create or an update.

I will check when and why this was added and try to get a benchmark!

acrobat: The check was commented in e52468c but the commit message isn't clear about why exactly this is commented

acrobat: @Soullivaneuh I've created a blackfire benchmark but I can't a real change in perfomance so I don't know why it was commented (by a bug report or ..). 

https://blackfire.io/profiles/compare/ceb1e796-d4bf-4c78-a9ab-0208df1f5fd8/graph

acrobat:
Rebased the branch against current master, pr is back up to date!

greg0ire:
Ready to merge.

OskarStark:
Thank you @acrobat 👍

Soullivaneuh:
Thank you @acrobat: You can see you work on the new release: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/releases/tag/3.0.1

Model/ModelManager.php
caponica: Putting this line back into the code seems to cause problems for me. I get the error 

    Parameter "id" for route "admin_caponica_foo_edit" must match "[^/]++" ("" given) 
    to generate a corresponding URL. 

After some digging around, I found that it was failling at this line of the code. The object already exists (it is being edited) and a call to `$values = $this->getIdentifierValues($entity);` works fine. 

However, for some possibly-doctrine-related reason that I don't fully understand, this condition (checking if the entity is in the unit of work) is returning false ... which leads to sonata trying to build an edit URL with a null id. Is it possible that it's getting a different EntityManager instance when using `$this->getEntityManager($entity)`? Or is there another reason it might not be seeing the (valid, existing) entity in the UOW?

If we look back at your case, @acrobat, is there any other way that you can test if your object is new instead of using this UOW check?

caponica:
I have left a comment above - the UOW check in this PR has caused problems in my experience.

Model/ModelManager.php
greg0ire: @caponica [related ? ](https://github.com/sonata-project/SonataAdminBundle/issues/4027#issuecomment-236136644)

caponica: @greg0ire - possibly, but in my case the entity should be managed... we're using the standard Sonata\AdminBundle\Controller\CRUDController::editAction(). Let me dig a bit more - there must be a reason why the Entity Manager doesn't see the entity.

greg0ire: Maybe you have several entity managers? I think there is indeed some more debugging to do. If you think it is the same entity manager, try to add breakpoints to methods that can make the entity become unmanaged (`clean` or `detach`, for instance). Also, maybe try the `$this->getEntityManager($entity)->getUnitOfWork()->isInIdentityMap($entity)` earlier in the request to see if it works at one point, then suddenly doesn't.

caponica: There's just one EntityManager (as in one database connection). We have a custom getObject() method in FooAdmin, maybe that's the cause? If not them I'm not sure what could be happening, since we're using standard Sonata methods for most of the rest of the call.

    public function getObject($id)
    {
        /** @var FooRepository $fooRepo */
        $fooRepo = $this->em->getRepository(FooRepository::CLASS_NAME);

        /** @var Foo $object */
        $object = $fooRepo->doComplicatedRetrieval($some, $options);

        $object->doSomething();
        return $object;
    }

And the entity manager in the Admin class is set via admin.yml:

    sonata.admin.foo:
        # ...
        calls:
            - [ setEntityManager, [ '@doctrine' ]]

greg0ire: What does `$this->em->getUnitOfWork()->isInIdentityMap($object)` return inside your custom method? Can you try disabling the custom method temporarily?

greg0ire: Uuuuh `@doctrine` is supposed to be the doctrine registry, not the entity manager, right?

caponica: Yes, it is. That's why the method called contains:

    $this->em = $doctrine->getManager();

If that was the problem, I think we would have known about it over the past two years this codebase has been running! :-)

greg0ire: Sure, I'm not questioning that, just raising an eyebrow at the naming : your method should be called `setDoctrineRegistry` instead of `setEntityManager`, shouldn't it?

caponica: (@doctrine is a parameter in the setEntityManager method. We don't store the doctrine service in the admin, just the EM. Maybe it could be called `setEntityManagerBasedOnDoctrineRegistry()`, but it's not something high on the 'to do' list of changes! Allowing users to edit objects without getting an error message or being bounced to edit// is the matter in hand :-) )

greg0ire: Oh I get it, kinda strange, but it makes some sense. What about my on topic questions?

caponica: // persist if the form was valid and if in preview mode the preview was approved
    if ($isFormValid && (!$this->isInPreviewMode() || $this->isPreviewApproved())) {
        try {
            // still in UOW here
            $object = $this->admin->update($object);
            // not in UOW here

greg0ire: Not sure what the difference is between being managed by the em, and being in the Iow. In the former, you are still here after a call to flush(). Does update call flush()? I'm on my phone, hard to browse code.

caponica: yes, `AbstractAdmin::update()` calls `ModelManager::update()` which calls `flush()`. In our project we then have all manner of doctrine events firing at that time, so it gets complicated.

Stepping back a bit, why do we care here if the object is in the UOW? Couldn't we use UnitOfWork::getEntityState() instead if some kind of state check is needed? I'm not sure of the case that @acrobat is trying to catch here, but I think all we care about is "can we get an identifier from the entity in hand", right?

If there's other logic related to detecting whether an object is new or not then I'm not sure it lives in `ModelManager::getNormalizedIdentifier()` whose purpose in life is just to "Get the identifiers for this model class as a string." So, imho, returning null just because we can't find the object in the UOW (even though the object is in the database and the ModelManager can work out an id for it) is not the best position to be in.

greg0ire: Some people set their identifiers manually, so no, getting an ID might not be good enough to guarantee the object is persisted. Moreover, on the object edit page, I think it is safe to assume the object was fetched already and is managed. I'd recommend using step by step debugging with xdebug and try to pinpoint when the problem occurs.

caponica: My point here is that I think that `UnitOfWork::isInIdentityMap()` is too unreliable. `UnitOfWork::getEntityState()` will give you one of four values, and from this you can tell if the object is persisted or not. In my case it returns `STATE_MANAGED` before the entity leaves the UOW and `STATE_DETACHED` afterwards. And both states identify that the entity is persisted in the database. Any new (unpersisted and unmanaged) object should return `STATE_NEW` and anything pending delete should return `STATE_REMOVED`.

If you're trying to test if an object has been persisted already then I think that checking for `STATE_MANAGED || STATE_DETACHED` would be more robust than simply using `isInIdentityMap()`.

However, as I said before, I don't really understand what you're trying to do in the case where identifiers are set manually, so perhaps I've missed something? However, if you just want this method to return null in cases where the entity in hand is unmanaged/unpersisted then you can do something like:

    if (!$entity) {
        return null;
    }

    $entityState = $this->getEntityManager($entity)->getUnitOfWork()->getEntityState($entity);
    if (UnitOfWork::STATE_NEW == $entityState || UnitOfWork::STATE_REMOVED == $entityState) {
        return null;
    }

Would that provide the required functionality in cases where IDs are set manually?

caponica: (sorry, in the original code, it just does a void `return` ... I added the nulls because the 'success' case returns something and my IDE was complaining about an inconsitent return point)

greg0ire: > Couldn't we use UnitOfWork::getEntityState() instead if some kind of state check is needed?

A cursory glance at [the doctrine doc article about this](http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/reference/working-with-objects.html) makes me think it is the correct method to call indeed. Maybe it wasn't available when the code was written. Can you please try that and tell us if it works?

greg0ire: Just noticed your two previous messages, I must be tired… 

> However, as I said before, I don't really understand what you're trying to do in the case where identifiers are set manually, so perhaps I've missed something?

Some people don't use auto-generated ids, which means simply relying on the presence of an id to determine whether something was persisted or not won't work.

However, using the sample code you provided should work perfectly. Could you please provide a PR for that? I think it is the way to go.

caponica: OK - will do. Thanks for your guidance!

greg0ire: Well thanks for your tenacity, keep up the good work :)

Issue #557, opened 2016-05-19
Remove knplabs/knp-menu-bundle require


Soullivaneuh:
This dependency has been added to "fix" the following issue: https://github.com/composer/composer/issues/5355

This have to be removed once resolved on a new stable release of composer.

PR #558, opened 2016-05-19
Move knplabs/knp-menu-bundle to require-dev

Soullivaneuh:
As the issue come only on Travis, this is not necessary to add this on the require section.

Ref: #557


greg0ire:
"comme" :trollface:

PR #571, opened 2016-05-28
DevKit updates for 3.x branch

SonataCI:


Makefile
```diff
@@ -8,7 +8,7 @@ all:
 	@echo "Please choose a task."
 
 test:
```
OskarStark: we should add `test-with-coverage` for this. As a developer i would like to execute `make test` without the coverage

ping @greg0ire @core23 @Soullivaneuh

Soullivaneuh: @OskarStark In this case, just do `phpunit`?

Personally  I never use `make test`.

PR #578, opened 2016-06-01
Support embedded object for mapping

claudusd:
### Changelog

```markdown
### Fixed
- Support embedded object for mapping
```

### Subject
This PR allow to use embedded object in the show mapper. 
```php
class Foo {
    public function getValue()
    {
        return 'myValue';
    }
}

class Bar {
    public function getFoo()
    {
        return $this->foo;
    }
}
```
``` php
protected function configureShowFields(ShowMapper $showMapper)
{
    $showMapper
        ->add('foo.value');
}
```
The FieldDescription method getValue ignore the embedded object and do 
``` php
$bar = new Bar();
$bar->getValue()
```
With my PR the getter call is on the embedded object and not on the parent object.
``` php
$bar->getFoo()->getValue()
```

### DIFF
--- a/Admin/FieldDescription.php
+++ b/Admin/FieldDescription.php
@@ -98,6 +98,13 @@ class FieldDescription extends BaseFieldDescription
             $object = $this->getFieldValue($object, $parentAssociationMapping['fieldName']);
         }
 
+        $fieldMapping = $this->getFieldMapping();
+        // Support embedded object for mapping
+        // Ref: http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/tutorials/embeddables.html
+        if (isset($fieldMapping['declaredField'])) {
+            $object = $this->getFieldValue($object, $fieldMapping['declaredField']);
+        }
+
         return $this->getFieldValue($object, $this->fieldName);
     }
 }
--- a/Resources/doc/reference/list_field_definition.rst
+++ b/Resources/doc/reference/list_field_definition.rst
@@ -131,7 +131,7 @@ Advance Usage
 Displaying sub entity properties
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-If you need to display only one field from a sub entity in a dedicated column, you can simply use the dot-separated notation:
+If you need to display only one field from a sub entity or embedded object in a dedicated column, you can simply use the dot-separated notation:
 
 .. note::
     This only makes sense when the prefix path is made of entities, not collections.
--- a/Resources/doc/tutorial/creating_your_first_admin_class/defining_admin_class.rst
+++ b/Resources/doc/tutorial/creating_your_first_admin_class/defining_admin_class.rst
@@ -44,6 +44,7 @@ First, you need to create an `Admin/PostAdmin.php` file:
             $showMapper
                 ->add('enabled')
                 ->add('title')
+                ->add('author.name')
                 ->add('abstract')
                 ->add('content')
                 ->add('tags')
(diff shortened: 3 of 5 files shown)


Soullivaneuh:
I can't evaluate the real benefit because I tried this on 3.x:

```php
// BankAccountAdmin
protected function configureShowFields(ShowMapper $showMapper)
{
    $showMapper
        ->add('customer.email')
    ;
}
```

Where `customer` is a `User` instance and it already works:

image

greg0ire:
@claudusd : are you referring to [this](http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/tutorials/embeddables.html) ? It's unclear.

OskarStark:
I think he mean if class `Foo` has no Admin class, does this work in your case @Soullivaneuh ?

claudusd:
@greg0ire Yes i referring to the doctrine Embeddables.

greg0ire:
> @greg0ire Yes i referring to the doctrine Embeddables.

Then I am :+1: for your PR! We don't have any support for Embeddables yet, so you might run into other problems later, but we would love PRs to make it work.

greg0ire:
LGTM

Soullivaneuh:
I would like to finish my test before merge.

greg0ire:
Also : it would be nice to document the support for this feature, once we are sure it works in most situations (so maybe in a later PR).

Admin/FieldDescription.php
```diff
@@ -98,6 +98,10 @@ public function getValue($object)
             $object = $this->getFieldValue($object, $parentAssociationMapping['fieldName']);
         }
 
+        if (isset($this->getFieldMapping()['declaredField'])) {
```
Soullivaneuh: Could you please add a comment about why this condition is here?

Something simple like this:

```php
// Support embedded object for mapping
// Ref: http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/tutorials/embeddables.html
```

Soullivaneuh:
@claudusd Your PR does not fix only show mapper but also list.

Could you please change the commit title to: 

> Support embedded object for mapping

Thanks.

claudusd:
Ok, i will change the commit message and add documentation.

Soullivaneuh:
Good! BTW I tested your PR manually, works great, good job! :+1:

claudusd:
Don't merge, I will add more documentation this evening.

Admin/FieldDescription.php
```diff
@@ -98,6 +98,12 @@ public function getValue($object)
             $object = $this->getFieldValue($object, $parentAssociationMapping['fieldName']);
         }
 
+        // Support embedded object for mapping
+        // Ref: http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/tutorials/embeddables.html
+        if (isset($this->getFieldMapping()['declaredField'])) {
```
EmmanuelVella: Hi, beware this syntax (`Function array dereferencing`) is only supported from php 5.4 (http://php.net/manual/en/migration54.new-features.php)

greg0ire: We have been having a problem with Travis, that's why the build wasn't red, but the next commit will probably report this.

claudusd:
I fix php 5.3 issue, rebase and add doc.

Soullivaneuh:
Thank you @claudusd! :+1:

Soullivaneuh:
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/releases/tag/3.0.2

PR #587, opened 2016-06-09
Replace deprecated 'form' type by FormType class

Soullivaneuh:
Replaces #587 

### Changelog
```markdown
### Fixed
- Deprecated usage of `form` type name
```

### DIFF
--- a/Tests/Builder/DatagridBuilderTest.php
+++ b/Tests/Builder/DatagridBuilderTest.php
@@ -1,0 +1,73 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DoctrineORMAdminBundle\Tests\Builder;
+
+use Sonata\AdminBundle\Datagrid\Pager;
+use Sonata\AdminBundle\Filter\FilterFactoryInterface;
+use Sonata\AdminBundle\Guesser\TypeGuesserInterface;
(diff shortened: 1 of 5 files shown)


OskarStark:
Good job @Soullivaneuh 👍

Tests/Builder/DatagridBuilderTest.php
```diff
@@ -0,0 +1,73 @@
...
+     * @var TypeGuesserInterface|\PHPUnit_Framework_MockObject_MockObject
+     */
+    protected $typeGuesser;
+
+    /**
+     * @var DatagridBuilder
+     */
+    protected $datagridBuilder;
+    /**
+     * @var FormFactoryInterface|\PHPUnit_Framework_MockObject_MockObject
```
greg0ire: The mock implements the interface, no need to do this IMO

Soullivaneuh: > The mock implements the interface, no need to do this IMO.

IDE does not recognize it if you don't specify it.

> Plus it would be &, not |

Where did you find that? It was always `|` on phpdoc comment AFAIK.

OskarStark: > Where did you find that? It was always | on phpdoc comment AFAIK.

AFAIK too

greg0ire: > Where did you find that? It was always | on phpdoc comment AFAIK.

Nowhere, but you don't mean that it's one or the other, you mean it is both, don't you ? Dunno what the symbol is for that.

> IDE does not recognize it if you don't specify it.

Oh so it's for autocompletion… I see.

Soullivaneuh: > Oh so it's for autocompletion… I see.

Not only. Type should be specified IMO.

Mock are special, because both types are acceptable.

> Dunno what the symbol is for that.

I will let it as is for the moment.

PR #588, opened 2016-06-09
Fix wrong property name on FormContractor

Soullivaneuh:
### Changelog

```markdown
### Fixed
- Fix wrong property name on FormContractor
```

### Subject

The other is never used. As a protected one, I keep it for BC.

Because of the deprecation, this should be delivered as a new minor.

### DIFF
--- a/Builder/FormContractor.php
+++ b/Builder/FormContractor.php
@@ -20,9 +20,16 @@ use Symfony\Component\Form\FormFactoryInterface;
 class FormContractor implements FormContractorInterface
 {
     /**
+     * @deprecated since version 3.x, to be removed in 4.0
+     *
      * @var FormFactoryInterface
      */
     protected $fieldFactory;
+
+    /**
+     * @var FormFactoryInterface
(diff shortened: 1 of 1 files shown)


Builder/FormContractor.php
```diff
@@ -20,11 +20,18 @@
...
+     * @deprecated since version 3.x, to be removed in 4.0
+     *
      * @var FormFactoryInterface
      */
     protected $fieldFactory;
 
     /**
+     * @var FormFactoryInterface
+     */
+    protected $formFactory;
```
greg0ire: Does it need to be protected ?

Soullivaneuh:
Please wait the patch release before merge this one.

Builder/FormContractor.php
Soullivaneuh: I don't know, but this was done like this for the `fake` one. We can discuss it later.

greg0ire: ok

PR #590, opened 2016-06-11
Append row instead of replacing all rows

auipga:
Closes https://github.com/sonata-project/SonataAdminBundle/issues/2383
### Changelog

``` markdown
### Changed
- behavior when appending a new empty row template
```
### Subject

The old code does reset all changes made before clicking the "new / add" button.
The new code leaves all other rows untouched.
### To do

- [ ] Update the tests
- [ ] Update the documentation
- [ ] Add an upgrade note

### DIFF
--- a/Resources/views/CRUD/edit_orm_one_association_script.html.twig
+++ b/Resources/views/CRUD/edit_orm_one_association_script.html.twig
@@ -47,8 +47,12 @@ This code manage the one-to-many association field popup
                     return;
                 }
 
-                var $appendRow = jQuery(jQuery.parseHTML(html)).find('tr:last');
-                jQuery('#field_container_{{ id }}').find('tr:last').after($appendRow);
+                if (jQuery('#field_widget_{{ id }}').children().length) {
+                    var $newRow = jQuery(jQuery.parseHTML(html)).find('tr:last');
+                    jQuery('#field_container_{{ id }}').find('tr:last').after($newRow);
+                } else {
+                    jQuery('#field_container_{{ id }}').replaceWith(html); // replace the html
+                }
 
                 Admin.shared_setup(jQuery('#field_container_{{ id }}'));
 
--- a/Resources/views/CRUD/edit_orm_one_association_script.html.twig
+++ b/Resources/views/CRUD/edit_orm_one_association_script.html.twig
@@ -47,7 +47,8 @@ This code manage the one-to-many association field popup
                     return;
                 }
 
-                jQuery('#field_container_{{ id }}').replaceWith(html); // replace the html
+                var $appendRow = jQuery(jQuery.parseHTML(html)).find('tr:last');
+                jQuery('#field_container_{{ id }}').find('tr:last').after($appendRow);
 
                 Admin.shared_setup(jQuery('#field_container_{{ id }}'));
 


Resources/views/CRUD/edit_orm_one_association_script.html.twig
```diff
@@ -47,7 +47,8 @@ This code manage the one-to-many association field popup
                     return;
                 }
 
-                jQuery('#field_container_{{ id }}').replaceWith(html); // replace the html
+                var $appendRow = jQuery(jQuery.parseHTML(html)).find('tr:last');
```
greg0ire: Why prefix the variable with a `$` ? cc @ju1ius

auipga: old habits to indicate that this is a jQuery object

Resources/views/CRUD/edit_orm_one_association_script.html.twig
```diff
@@ -47,7 +47,12 @@ This code manage the one-to-many association field popup
                     return;
                 }
 
-                jQuery('#field_container_{{ id }}').replaceWith(html); // replace the html
+                if (jQuery('#field_widget_{{ id }}').children().length) {
+                    var $newRow = jQuery(jQuery.parseHTML(html)).find('tr:last');
```
ju1ius: @auipga Have you tested that this is correct when the collection is sortable ?

ju1ius: @auipga The code does submit the whole form to the server in order to rebuild the collection field.
So the real issue here is select2 not synchronizing changes to the underlying input, because the submit event is stopped before select2 has a chance to get at it.

So your first move would be to try to remove the `event.stopPropagation()` call. I don't think it's needed anyway.
If you're lucky then great !
If not, then you need to find a way to force select2 to update the underlying input _before_ the ajax call.
Something like this (not tested but should give you the idea):
```js
// panzer division method, so this might even work with other plugins
$('#field_container_{{ id }}').find(':input').change();
```

Soullivaneuh:
Is that a bugfix? In this case you are targeting the wrong branch. Please reopen it on `3.x`.

Also the changelog should be on the `Fixed` section and the empty part should be removed.

Soullivaneuh:
Note: Have to be manually tested to reproduce the issue before merge.

Resources/views/CRUD/edit_orm_one_association_script.html.twig
greg0ire: Oh ok, might be handy indeed

ju1ius: Especially when you start using both "raw" dom elements and jQuery objects.

Resources/views/CRUD/edit_orm_one_association_script.html.twig
core23: Please have a look at the @ju1ius comments

auipga: I understand the problem with my solution and a sorted collection. But the problem with recreating the whole form is, that file upload fields can't be filled by javascript or value attribute, so this information will get lost. :(

sujayjaju:
@auipga #720 addresses the file input issue that you were trying to solve.

sujayjaju:
@SonataCI The already merged commits from #720 resolves this issue.

greg0ire:
Let's close then.

PR #591, opened 2016-06-13
Fix not working filters on model list popup

jontorrado:
### Changelog

```markdown
### Fixed
- Create form is shown instead of filters on `sonata_type_model_list` popup 
```

### Subject

When using `sonata_type_model_list`, dropdown filters are not working inside the list popup due to an AJAX call. This function is catching that website click, so we must assure that the click href contains something different than *empty* and *#* to make the AJAX request. `return` is needed so other JS files are able to catch that event. For example: Bootstrap dropdown (filters, as seen in previous lines).

### DIFF
--- a/Resources/views/CRUD/edit_orm_many_association_script.html.twig
+++ b/Resources/views/CRUD/edit_orm_many_association_script.html.twig
@@ -34,6 +34,10 @@ This code manage the many-to-[one|many] association field popup
         initialize_popup_{{ id }}();
 
         var target = jQuery(this);
+        
+        if (target.attr('href').length === 0 || target.attr('href') === '#') {
+            return;
+        }
 
         event.preventDefault();
         event.stopPropagation();


greg0ire:
Keeped => kept

Resources/views/CRUD/edit_orm_many_association_script.html.twig
```diff
@@ -34,6 +34,10 @@ This code manage the many-to-[one|many] association field popup
         initialize_popup_{{ id }}();
 
         var target = jQuery(this);
+        
+        if (target.attr('href').length === 0 || target.attr('href') === '#') {
```
greg0ire: What if it is `#some-id` ? Or `javascript()` ? Isn't there some better way to detect valid urls? Cc @ju1ius

jontorrado: Sure, I just checked for that option but it should be quite better to check other variations

ju1ius: @greg0ire @jontorrado  We're not trying to detect valid urls here, we're trying to detect if the link is an anchor.
An anchor points to the either the exact same url as the current page or the same url with a different hash.
So basically to do that you have two solutions, the long one:
```js
// HTMLElement -> boolean
const isAnchor = a => (
  a.tagName === 'A'
  && a.origin === location.origin
  && a.pathname === location.pathname
  && a.search === location.search
)
```
or the short one:
```js
// HTMLElement -> boolean
const isAnchor = a => {
  const href = (a.getAttribute('href') || '').trim()
  // notice the href[0] === '#' and not href === '#' 
  return a.tagName === 'A' && (!href || href[0] === '#')
}
```
So the shortest way gives us, with jQuery:
```js
// jQuery -> boolean
function isAnchor ($a) {
    var href = ($a.attr('href') || '').trim();
    return $a.is('a') && (!href || href[0] === '#');
}
```

As for the `href="javascript:alert(this)"` case, nobody should use that, so don't bother.

greg0ire: Stellar answer as always, thanks @ju1ius

ju1ius: That's kind of you ! However my answer was partly wrong...
In this particular context, we're checking the url for a link that has been loaded from a different url.
As the location object always refers to the "top level" location, the first solution won't work. If the loaded html fragment contains a link to the current "top level" page, it is valid to open it in a new dialog.
So the second method is actually the only valid option.

Soullivaneuh:
Could you please details the steps to reproduce? Maybe with some screenshots?

jontorrado:
Here are the steps:
1. Install the last stable Sonata Admin release
2. Create two different admins with a Many to One relation.
3. Use a 'sonata_type_model_list' widget in the relation form
4. In the create form, click "list" in that field and, in the popup, click "filters"

Image below:
image

Resources/views/CRUD/edit_orm_many_association_script.html.twig
greg0ire: What do you mean by "top level" exactly? Are you thinking iframes or something like that?

Soullivaneuh:
When I click on the filter, I have the main entity create form:

anim

Is this issue you are trying to fix on this PR?

jontorrado:
Sure, thats because the href contains a "#", and it makes an AJAX request loading the same page (which is the actual form, as you see in that gif). Try the patch and see that it fixes this issue.

Soullivaneuh:
I updated your changelog to be more user comprehensive.

Soullivaneuh:
@jontorrado Indeed, works like a charm, thank you! :+1: 

I update the title to be more accurate, could you please do the same for your commit title?

@greg0ire @ju1ius is this PR LGTM except my comment?

greg0ire:
The grammar in the commit subject needs to be fixed : Keeped => kept

greg0ire:
And assure => ensure I think. Apart from this, great commit message :+1:

greg0ire:
And then there is @ju1ius 's comment of course.

Soullivaneuh:
> Apart from this, great commit message

The message yes, not the title.

> Kept default functionality

Which one? This is not enough clear.

If the PR is OK for @ju1ius I can modify it on a squash merge anyway. :+1:

jontorrado:
I cannot amend this commit, it was done with the Github.com edit feature, not with a fork :( You can close this PR and I can do it again with the correct message:

> Fix not working filters on model list popup

greg0ire:
> I cannot amend this commit, it was done with the Github.com edit feature, not with a fork

Of course you can.

Soullivaneuh:
> I cannot amend this commit, it was done with the Github.com edit feature, not with a fork :(

You can even clone your project and get the `patch-2` branch

> You can close this PR and I can do it again with the correct message:

Please don't do this. At least we can edit the message while merging.

greg0ire:
Getting your code locally : 

```
git checkout -b jontorrado-patch-2 3.x
git pull https://github.com/jontorrado/SonataDoctrineORMAdminBundle.git patch-2
```

Then `git commit --amend` and `git push --force -u my_newly_created_fork the_branch`

jontorrado:
Sorry, I didn't see that fork, haha. I've just changed the commit message.

greg0ire:
The commit message is worse now…

Soullivaneuh:
@jontorrado The commit title is good, but you can keep the message on it. :+1:

Soullivaneuh:
@ju1ius what is your status for this PR?

Resources/views/CRUD/edit_orm_many_association_script.html.twig
ju1ius: > What do you mean by "top level" exactly?

[window.top](https://developer.mozilla.org/en-US/docs/Web/API/Window/top)

Url-related properties of anchors (a.href, a.origin...) are resolved against it's window's location (`a.ownerDocument.defaultView.location`).

If the page were loaded in an iframe, solution 1. would work by doing this:
```js
const isAnchor = a => {
  const loc = a.ownerDocument.defaultView.location
  return a.tagName === 'A'
    && a.origin === loc.origin
    && a.pathname === loc.pathname
    && a.search === loc.search
}
```

But again, since we load the page via ajax, a.href will always be resolved against the top-level window.

Now to be even more accurate, solution n°2 could look more like:
```js
// String -> String
const withoutHash = url => url.replace(/#(.*)$/, '')
// String -> HTMLElement -> Boolean
const isAnchorToUrl = (url, a) => {
  const href = (a.getAttribute('href') || '').trim()
  return a.tagName === 'A' && (
    !href // empty href
    || href[0] === '#' // hash only
    // same urls, hash-insensitive
    || withoutHash(url) === withoutHash(href)
    // after that, things start getting hairy...
  )
}
```
Which still doesn't handle all possible cases...
However I think the short version of n°2 can be considered good enough for the use-case.

Soullivaneuh: Should @jontorrado update his code?

ju1ius:
> @ju1ius what is your status for this PR?

In `Resources/views/CRUD/edit_orm_many_association_script.html.twig`:

The following function should be added at the beginning of the `<script>` tag:
```js
function isAnchor ($a) {
  var href = ($a.attr('href') || '').trim()
  return $a.is('a') && (!href || href[0] === '#')
}
```
And [line 38](https://github.com/jontorrado/SonataDoctrineORMAdminBundle/blob/5f53871b8d5ffef746faa2f70be01537e8435dce/Resources/views/CRUD/edit_orm_many_association_script.html.twig#L38) should be changed to:
```js
if (isAnchor(target)) {
```

Soullivaneuh:
I don't like the fact to add a generic `isAnchor` under a script tag, this could be added somewhere else.

Or a extra lib could help.

Soullivaneuh:
BTW, his condition work for this link. I don't think it should be changed, am I wrong?

jontorrado:
Any updates?

Soullivaneuh:
I will merge this one as this fixes a quite critical issue.

@ju1ius I agree for the `isAnchor` method, but not directly on this part. I'll be glad to see a PR to improve it! :wink:

Soullivaneuh:
Thank you @jontorrado

ju1ius:
>  I'll be glad to see a PR to improve it! 

You [already did](https://github.com/sonata-project/SonataAdminBundle/pull/3647/files#diff-71a75de45c915dd9990a4da7d9b5dc0aR34) :wink: 

> I will merge this one as this fixes a quite critical issue.

As a side note, you should keep in mind that templates for CRUD _associations_ have been **copy/pasted** from the original admin bundle to the persistence bundles.
It implies that this issue is very likely to be present also in the mongo, phpcr and propel bundles. Let me check... yes it is !
I just started work to bring them all back to the admin bundle, but until this is finished and merged, a proper fix should result in **4** PRs, not one.

Soullivaneuh:
> You [already did](https://github.com/sonata-project/SonataAdminBundle/pull/3647/files#diff-71a75de45c915dd9990a4da7d9b5dc0aR34) :wink: 

Well, not a PR with 149 modified files. :stuck_out_tongue: 

> As a side note, you should keep in mind that templates for CRUD associations have been copy/pasted from the original admin bundle to the persistence bundles.
> It implies that this issue is very likely to be present also in the mongo, phpcr and propel bundles. Let me check... yes it is !

Yes I know... Waiting your work, it should be fixed on each ones...

> I just started work to bring them all back to the admin bundle, but until this is finished and merged, a proper fix should result in **4** PRs, not one.

4 or more, be sure they will be precise and enough small for fast review. I prefer a step by step long evolution, but reliable instead of merging one or 4 PR if they are too big.

PR #596, opened 2016-06-23
Moves crud association templates to admin-bundle

ju1ius:
This PR is related to [SonataAdminBundle#3971](https://github.com/sonata-project/SonataAdminBundle/pull/3971)
I am targetting this branch, because there are BC breaks
## Changelog

``` markdown
### Added
- `FieldDescription` implements `FieldDescriptionInterface::describesSingleValuedAssociation` and `FieldDescriptionInterface::describesCollectionValuedAssociation` methods

### Changed
- `FormContractor` now extends `Sonata\AdminBundle\Admin\AbstractFormContractor`

### Removed
- removed ALL crud association templates and form widget blocks. They now belong to the admin bundle

```

PR #599, opened 2016-07-02
Check if property exists and throw exception if not

ebuildy:
I am targetting this branch, because this is the version I use in different projects.
## Changelog

``` markdown
### Added
- Throw exception if propertyName doesnot exist in fieldMappings
```

This could help a lot for debuging, throw exception with propertyName and class name instead of a Php Notice that is hard to debug.

### DIFF
--- a/Guesser/FilterTypeGuesser.php
+++ b/Guesser/FilterTypeGuesser.php
@@ -60,6 +60,11 @@ class FilterTypeGuesser extends AbstractTypeGuesser
                     return new TypeGuess('doctrine_orm_model', $options, Guess::HIGH_CONFIDENCE);
             }
         }
+        
+        if (!isset($metadata->fieldMappings[$propertyName]))
+        {
+            throw new \Exception('Property ' . $propertyName . ' not found in class' . $class . '!');
+        }
 
         $options['field_name'] = $metadata->fieldMappings[$propertyName]['fieldName'];
 


core23:
> because this is the version I use in different projects.

That's no reason You have to explain why you choose the master or the stable branch.

For me it's a BC break, because usage changed. Code that just worked will now fail with an exception.

Soullivaneuh:
> This could help a lot for debuging, throw exception with propertyName and class name instead of a Php Notice that is hard to debug.

Could you please explain a little bit more your context? In which cases this could happen and why the actual behavior is not correct?

> For me it's a BC break, because usage changed. Code that just worked will now fail with an exception.

Quite agree, but it's depends if the missing property is a normal way or not. I mean, if this missing property finish by an error at the end or not.

Guesser/FilterTypeGuesser.php
```diff
@@ -60,6 +60,11 @@ public function guessType($class, $property, ModelManagerInterface $modelManager
                     return new TypeGuess('doctrine_orm_model', $options, Guess::HIGH_CONFIDENCE);
             }
         }
+        
+        if (!isset($metadata->fieldMappings[$propertyName]))
```
Soullivaneuh: Use `array_key_exists` instead,

Guesser/FilterTypeGuesser.php
```diff
@@ -60,6 +60,11 @@ public function guessType($class, $property, ModelManagerInterface $modelManager
...
+        {
+            throw new \Exception('Property ' . $propertyName . ' not found in class' . $class . '!');
```
Soullivaneuh: Using `\Exception` is not very accurate IMHO.

Soullivaneuh: No need for the `!`.

Soullivaneuh:
Could you please also (if PR goal is accepted):

* Fix CS: https://styleci.io/analyses/8Q35Mm
* Add a test for your modification

Soullivaneuh:
cc @ju1ius: Maybe this file is part of your refactoring.

ju1ius:
> cc @ju1ius: Maybe this file is part of your refactoring.

@Soullivaneuh Nope, #596 doesn't deal with filters. Just association forms.

Guesser/FilterTypeGuesser.php
ju1ius: > Using \Exception is not very accurate IMHO.

A custom exception would lovely! :heart: 

```php
namespace Sonata\DoctrineORMAdminBundle\Model;

class MissingPropertyMetadata extends \LogicException
{
  public function __construct($class, $property)
  {
    parent::__construct(sprintf(
      'No metadata found for property `%s::$%s`.'
      . ' Please make sure your Doctrine mapping is properly configured.',
      $class,
      $property
    ));
  }
}
```

```php
if (empty($metadata->fieldMappings[$propertyName])) {
  throw new MissingPropertyMetadata($class, $property);
}
```

core23:
Please rebase to retrigger StyleCI @ebuildy

Guesser/FilterTypeGuesser.php
core23: Please have a look at this @ebuildy

ebuildy: Yes this is much better! 

@ju1ius I think you can post a new PR, I will close this one. Thanks you

greg0ire: @ebuildy : can you please instead do what @ju1ius suggested?

OskarStark:
Closing in favor of #676

PR #611, opened 2016-08-07
Adds missing translation of 'Delete' in edit view

core23:

## Changelog

```markdown
### Fixed
- Adds missing translation of 'Delete' in edit view
```

## Subject

The `Delete` label wasn't translated.

### DIFF
--- a/Resources/views/CRUD/edit_orm_one_to_many.html.twig
+++ b/Resources/views/CRUD/edit_orm_one_to_many.html.twig
@@ -106,7 +106,7 @@ file that was distributed with this source code.
                                 </div>
 
                                 {% if nested_group_field['_delete'] is defined %}
-                                    {{ form_row(nested_group_field['_delete']) }}
+                                    {{ form_row(nested_group_field['_delete'],  { 'label': ('action_delete'|trans({}, 'SonataAdminBundle')) } ) }}
                                 {% endif %}
                             </div>
                         {% endfor %}


Resources/views/CRUD/edit_orm_one_to_many.html.twig
```diff
@@ -106,7 +106,7 @@ file that was distributed with this source code.
                                 </div>
 
                                 {% if nested_group_field['_delete'] is defined %}
-                                    {{ form_row(nested_group_field['_delete']) }}
+                                    {{ form_row(nested_group_field['_delete'],  { 'label': ('action_delete'|trans({}, 'SonataAdminBundle')) } ) }}
```
greg0ire: Please linebreak this.

core23: IMO we don't need a break here, because we have many indentions caused by the html structure. 

Btw. we have many lines in this file who have much longer lines.

greg0ire: > IMO we don't need a break here, because we have many indentions caused by the html structure. 

This does not change anything, because editors dont shift right when the whole part of the file in the view port is indented, do they? This is still hard to read. This is a problem, it could be solved with include statements. I guess we also could remove the extra level of indention that is here for no apparent reason, on [this line](https://github.com/core23/SonataDoctrineORMAdminBundle/blob/master/Resources/views/CRUD/edit_orm_one_to_many.html.twig#L17)

> Btw. we have many lines in this file who have much longer lines.

And many services have the whole container directly injected in them. There are many problems in this project, I don't think we should view the existing code as a reason to do the same.

greg0ire: I think some `if` statements can be combined.

greg0ire: See #612

OskarStark:
I merge this as it is!

We can discuss the line breaking stuff later

greg0ire:
> We can discuss the line breaking stuff later

Yeah, it's not like this is the only long line in this template anyway…

PR #614, opened 2016-08-13
Improved support for symfony3

core23:

## Changelog

```markdown
### Changed
 - Use class names for filters when using symfony 2.8+
```
## Subject

This will reduce some deprecation warnings and supports the new class usage for forms.

### DIFF
--- a/Filter/AbstractDateFilter.php
+++ b/Filter/AbstractDateFilter.php
@@ -165,6 +165,17 @@ abstract class AbstractDateFilter extends Filter
             $name .= '_range';
         }
 
+        // NEXT_MAJOR: Remove this line when drop Symfony <2.8 support
+        if (method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')) {
+            $classnames = array(
+                'sonata_type_filter_date' => 'Sonata\AdminBundle\Form\Type\Filter\DateType',
+                'sonata_type_filter_date_range' => 'Sonata\AdminBundle\Form\Type\Filter\DateRangeType',
+                'sonata_type_filter_datetime' => 'Sonata\AdminBundle\Form\Type\Filter\DateTimeType',
+                'sonata_type_filter_datetime_range' => 'Sonata\AdminBundle\Form\Type\Filter\DateTimeRangeType',
+            );
+            $name = $classnames[$name];
+        }
+
         return array($name, array(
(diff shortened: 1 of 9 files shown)


Filter/AbstractDateFilter.php
```diff
@@ -165,6 +165,24 @@ public function getRenderSettings()
...
+                    break;
+                case 'sonata_type_filter_datetime':
+                    $name = 'Sonata\AdminBundle\Form\Type\Filter\DateTimeType';
+                    break;
+                case 'sonata_type_filter_datetime_range':
+                    $name = 'Sonata\AdminBundle\Form\Type\Filter\DateTimeRangeType';
+                    break;
+            }
+        }
+
```
greg0ire: Maybe use an inflected here?

greg0ire: Inflector damn autocorrect

core23: Can you please elaborate?

greg0ire: There are classes to convert snake case to camel case and so on. Doctrine inflector or maybe sf has  its own class

core23: Not sure if we can improve this with an inflector class. We can't convert everything from snake to camel case.

greg0ire: Then maybe use an associative array type => class?

Filter/BooleanFilter.php
```diff
@@ -66,7 +66,12 @@ public function getDefaultOptions()
      */
     public function getRenderSettings()
     {
-        return array('sonata_type_filter_default', array(
+        // NEXT_MAJOR: Remove this line when drop Symfony <2.8 support
+        $type = method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')
```
OskarStark: can we move this to a method in `abstract class Filter extends BaseFilter` ?

in this case we avoid code duplication for the default stuff

core23: 👎 Introducing a method only for BC handling is a bad idea

cc @Soullivaneuh

OskarStark: ok, fine for me

lets merge this as it is

OskarStark:
Thank you @core23 👍

PR #616, opened 2016-08-25
[DO NOT MERGE]Avoid two last versions of phpunit mock objects

greg0ire:
I am targetting this branch, because this is a patch

They are buggy. This implies using phpunit from the vendors.
See
https://github.com/sebastianbergmann/phpunit-mock-objects/pull/324/files
and
https://github.com/sebastianbergmann/phpunit-mock-objects/issues/322

This is a POC, do not merge, it is just to show how the build can be fixed, but should not be merged because some files are supposed to come from dev-kit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ lint:
 	git diff --exit-code
 
 test:
-	phpunit -c phpunit.xml.dist --coverage-clover build/logs/clover.xml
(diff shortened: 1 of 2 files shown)


greg0ire:
To be closed in favor #617 , this is here for science.

PR #617, opened 2016-08-25
Mock class that exists

greg0ire:
I am targetting this branch, because this is a patch

## Subject

This is a workaround for a phpunit mock objects bug, see https://github.com/sebastianbergmann/phpunit-mock-objects/pull/324

To be merged ASAP :rotating_light:


greg0ire:
This makes the tests a bit uglier, but allows not to delay the decision about [phpunit from the vendors](https://github.com/sonata-project/dev-kit/issues/197)

PR #613, opened 2016-08-10
fix for FilterTypeGuesser for ManyToOne relations, it conflicts with …

victordevaid:
I am targetting this branch, because it causes conflicts with current versions of symfony.

Fixes https://github.com/sonata-project/SonataAdminBundle/issues/4052

## Changelog

```markdown

### Changed
- The way `FilterTypeGuesser` resolves ManyToOne relations, so it supports the new way symfony defines form types (FQCN)
```

## Subject

It changes  the form FilterTypeGuesser.php resolves ManyToOne relations, so it supports the new way symfony defines types (FQCN).

PR #621, opened 2016-09-02
Replaced form type names with classnames

deguif:
I am targetting this branch, because it's about avoiding deprecation warnings when using Sf >= 2.8, while keeping the old behavior.

## Changelog
```markdown
## Fixed
- Use class name when referencing `Form Type` to be compatible with Symfony 2.8+
```
Closes #613

## Subject

Remove deprecation warnings when using Sf >= 2.8

### DIFF
--- a/Guesser/FilterTypeGuesser.php
+++ b/Guesser/FilterTypeGuesser.php
@@ -45,11 +45,16 @@ class FilterTypeGuesser extends AbstractTypeGuesser
                 case ClassMetadataInfo::ONE_TO_MANY:
                 case ClassMetadataInfo::MANY_TO_ONE:
                 case ClassMetadataInfo::MANY_TO_MANY:
-
-                    $options['operator_type'] = 'sonata_type_equal';
+                    // NEXT_MAJOR: Remove ternary (when requirement of Symfony is >= 2.8)
+                    $options['operator_type'] = method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')
+                        ? 'Sonata\CoreBundle\Form\Type\EqualType'
+                        : 'sonata_type_equal';
                     $options['operator_options'] = array();
 
-                    $options['field_type'] = 'entity';
+                    // NEXT_MAJOR: Remove ternary (when requirement of Symfony is >= 2.8)
+                    $options['field_type'] = method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')
+                        ? 'Symfony\Bridge\Doctrine\Form\Type\EntityType'
+                        : 'entity';
                     $options['field_options'] = array(
                         'class' => $mapping['targetEntity'],
                     );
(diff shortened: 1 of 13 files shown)


Guesser/FilterTypeGuesser.php
```diff
@@ -77,18 +85,21 @@ public function guessType($class, $property, ModelManagerInterface $modelManager
                 return new TypeGuess('doctrine_orm_date', $options, Guess::HIGH_CONFIDENCE);
             case 'decimal':
             case 'float':
-                $options['field_type'] = 'number';
```
OskarStark: can you explain why you remove it?

Guesser/FilterTypeGuesser.php
```diff
@@ -77,18 +85,21 @@ public function guessType($class, $property, ModelManagerInterface $modelManager
...
-
-                return new TypeGuess('doctrine_orm_number', $options, Guess::MEDIUM_CONFIDENCE);
```
OskarStark: and this?

Guesser/FilterTypeGuesser.php
deguif: Because this is the same than the case just after.

Guesser/FilterTypeGuesser.php
deguif: Same as last response ;)

greg0ire: There is no break statement, which means lines 92 - 96 will be executed, and they do the same. So this was unnecessarily duplicated code if I understand correctly. @deguif : next time, do that in a separate commit (you can use `git add --patch` to do that easily).

greg0ire:
LGTM

Guesser/FilterTypeGuesser.php
OskarStark: indeed

Guesser/FilterTypeGuesser.php
OskarStark: you both areright, i saw number != doctrine_orm_number at the first glympse, whcih confused me a bit ;-)

deguif: For sure, I will do it in another PR next time.

greg0ire: Another PR would be maybe a bit overkill, a separate commit is enough in this case IMO.

deguif:
I just saw thereis another PR that has started to do the same: #613

greg0ire:
Added "Closes #613" to your PR body

greg0ire:
@OskarStark , I'll let you merge if you're ok with this.

OskarStark:
> @OskarStark , I'll let you merge if you're ok with this.

I am, thank you @deguif 👍

greg0ire:
Yeah thanks for your contribution(s)!

PR #622, opened 2016-09-09
Fix formcontractor typecheck for classes

nlzet:
I am targetting this branch, because there are no BC breaks.

## Changelog

```markdown
### Changed
- Changed how `FormContractor::getDefaultOptions` checks which type is used. Instead of checking for an array of available types, we improve this by checking for the class instance or parents. 
```

This fixes an issue, when you for example try to extend ModelAutocompleteType, you will receive a fatal error: `Catchable Fatal Error: Argument 1 passed to Sonata\AdminBundle\Form\DataTransformer\ModelToIdPropertyTransformer::__construct() must implement interface Sonata\AdminBundle\Model\ModelManagerInterface, null given, called in /app/vendor/sonata-project/admin-bundle/Form/Type/ModelAutocompleteType.php on line 37 and defined.`

That was because it only checks for the ModelAutocompleteType, and not types extending it.

### DIFF
--- a/Builder/FormContractor.php
+++ b/Builder/FormContractor.php
@@ -99,17 +99,18 @@ class FormContractor implements FormContractorInterface
         $options['sonata_field_description'] = $fieldDescription;
 
         // NEXT_MAJOR: Check only against FQCNs when dropping support for Symfony <2.8
-        if (in_array($type, array(
-            'sonata_type_model',
-            'Sonata\AdminBundle\Form\Type\ModelType',
-            'sonata_type_model_list',
-            'Sonata\AdminBundle\Form\Type\ModelTypeList',
-            'Sonata\AdminBundle\Form\Type\ModelListType',
-            'sonata_type_model_hidden',
-            'Sonata\AdminBundle\Form\Type\ModelHiddenType',
-            'sonata_type_model_autocomplete',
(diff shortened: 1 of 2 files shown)


nlzet:
Do I need to do something for the coverage, since I can't seem to find out what decreased it (0.2%).

Builder/FormContractor.php
```diff
@@ -176,4 +171,29 @@ private function hasAssociation(FieldDescriptionInterface $fieldDescription)
             ClassMetadataInfo::ONE_TO_ONE,
         ));
     }
+
+    /**
+     * @param string $type
+     * @param array  $types
+     *
+     * @return bool
+     *              // NEXT_MAJOR: See next major comments above, this method should be removed when dropping support for Symfony <2.8
```
OskarStark: please put this line to line 176 and remove the `//`

OskarStark:
LGTM 👍 

you can ignore the coverage for now, but please apply my comment

OskarStark:
awesome, can you please squash your commits into one and give it a good commit message?

nlzet:
@OskarStark squashed with a new commit message and applied StyleCI fixes.

OskarStark:
Thank you @nlzet, good job 👍

PR #627, opened 2016-09-26
Entity persistence detection

caponica:
the changes in this PR make the existing code more robust and address possible bugs caused by the UnitOfWork not carrying already-persisted entities in some scenarios

See discussion:
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/547#r80505888

### Changed
- Changed how persisted entities are detected, now using UnitOfWork::getEntityState() instead of isInIdentityMap()

## Subject
- Instead of looking at the UnitOfWork identity map (which may detach objects after a flush), this revision uses the reported entity state which should be a more robust indicator of whether or not a given entity has been persisted


caponica:
Sorry, this is based on the wrong branch.... let me see if I can remember how to fix that!

greg0ire:
You can change base branches by clicking the edit button I think ;)

PR #631, opened 2016-09-30
Improve detection of entity persistence:

caponica:
the changes in this PR make the existing code more robust and address possible bugs caused by the UnitOfWork not carrying already-persisted entities in some scenarios

See discussion:
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/547#r80505888

## Changelog

```markdown
### Changed
- Changed how persisted entities are detected, now using `UnitOfWork::getEntityState()` instead of `isInIdentityMap()`
```

## Subject
Instead of looking at the UnitOfWork identity map (which may detach objects after a flush), this revision uses the reported entity state which should be a more robust indicator of whether or not a given entity has been persisted

### DIFF
--- a/Model/ModelManager.php
+++ b/Model/ModelManager.php
@@ -19,6 +19,7 @@ use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\OptimisticLockException;
 use Doctrine\ORM\QueryBuilder;
+use Doctrine\ORM\UnitOfWork;
 use Exporter\Source\DoctrineORMQuerySourceIterator;
 use Sonata\AdminBundle\Admin\FieldDescriptionInterface;
 use Sonata\AdminBundle\Datagrid\DatagridInterface;
@@ -385,8 +386,12 @@ class ModelManager implements ModelManagerInterface, LockInterface
             throw new \RunTimeException('Invalid argument, object or null required');
         }
 
-        // the entities is not managed
-        if (!$entity || !$this->getEntityManager($entity)->getUnitOfWork()->isInIdentityMap($entity)) {
+        if (!$entity) {
+            return;
+        }
+
+        $entityState = $this->getEntityManager($entity)->getUnitOfWork()->getEntityState($entity);
+        if (UnitOfWork::STATE_NEW == $entityState || UnitOfWork::STATE_REMOVED == $entityState) {
             return;
         }
 


OskarStark:
Thank you @caponica 👍

Model/ModelManager.php
```diff
@@ -385,8 +386,12 @@ public function getNormalizedIdentifier($entity)
...
         }
 
-        // the entities is not managed
-        if (!$entity || !$this->getEntityManager($entity)->getUnitOfWork()->isInIdentityMap($entity)) {
+        if (!$entity) {
+            return;
+        }
+
+        $entityState = $this->getEntityManager($entity)->getUnitOfWork()->getEntityState($entity);
+        if (UnitOfWork::STATE_NEW == $entityState || UnitOfWork::STATE_REMOVED == $entityState) {
```
Soullivaneuh: Comparison should be strict.

caponica: I have no problem with this but, in my defence:
 - `getEntityState()` says that is will only return integers, and
 - would we really care if it returns '2' as a string for some reason instead of 2 as an int? (or '4')?

Secondly, I've always been told to avoid in_array unless it's really necessary (as used in #633 ). I think it's for performance reasons, but maybe I've just been doing this too long and it's not an issue any more!

Soullivaneuh: > would we really care if it returns '2' as a string for some reason instead of 2 as an int? (or '4')?

Yes we really care of this. If it returns integers, let's compare integers.

Strict comparison is more reliable. For example, if this method returns also a boolean for some reasons on a next releases, this can have an impact on our comparison if not strict.

So if we can do a strict comparison, we have to do a strict comparison. :wink: 

For the `in_array` usage, I never been told about performance or anything else. But I would love to read an article about this if you have it!

greg0ire: I second the strict comparison argument, as it can be seen as unnecessary, distracting noise. As for the performance argument, I think we should optimize for readability, and this piece of code is not called intensively, AFAIK

caponica: All fine by me!

(On reflection, I agree with you about the strict comparison. But not entirely convinced that the `in_array` version is more readable!)

P.S. Dredging my memory for in_array performance, I think it's something to do with the way in_array scans the whole array... but it's almost certainly a non-issue on small arrays like this, it's just that, over the years, I've become hard-coded myself to think twice about using it!

Issue #210, opened 2013-04-29
Error in multiple sorting (PosgreSQL)


ZhukV:
I have a error in multiple sorting.

My code for create query:

``` php
    public function createQuery($context = 'list')
    {
        /** @var \Doctrine\ORM\EntityRepository $er */
        $er = $this->getConfigurationPool()->getContainer()
            ->get('doctrine')
            ->getRepository('FindCapacity\Entity\IndustryActivity');

        $qb = $er->createQueryBuilder('ia')
            ->orderBy('ia.root', 'ASC')
            ->addOrderBy('ia.lft', 'ASC')
        ;

        return new ProxyQuery($qb);
    }
}
```

Database: postgress 9.1.6

Error:

```
SQLSTATE[42P10]: Invalid column reference: 7 ERROR: for SELECT DISTINCT, ORDER BY expressions must appear in select list
```

Solution this bug:
Reset "orderBy" dql part in countable and load `idx` queries.

Patchs:

``` patch
Index: vendor/sonata-project/doctrine-orm-admin-bundle/Sonata/DoctrineORMAdminBundle/Datagrid/ProxyQuery.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- vendor/sonata-project/doctrine-orm-admin-bundle/Sonata/DoctrineORMAdminBundle/Datagrid/ProxyQuery.php   (revision )
+++ vendor/sonata-project/doctrine-orm-admin-bundle/Sonata/DoctrineORMAdminBundle/Datagrid/ProxyQuery.php   (revision )
@@ -80,7 +80,7 @@

         // step 3 : retrieve the different subjects id
         $select = sprintf('%s.%s', $queryBuilderId->getRootAlias(), $idName);
-        $queryBuilderId->resetDQLPart('select');
+        $queryBuilderId->resetDQLParts(array('select', 'orderBy'));
         $queryBuilderId->add('select', 'DISTINCT ' . $select);

         //for SELECT DISTINCT, ORDER BY expressions must appear in select list
```

``` patch
Index: vendor/sonata-project/doctrine-orm-admin-bundle/Sonata/DoctrineORMAdminBundle/Datagrid/Pager.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
[...]

PR #544, opened 2016-05-10
Added missing default sort by primary key(s)

greg0ire:

PR #639, opened 2016-10-31
reset orderBy DQL part in datagrid query

### LINKED ISSUE #458: OrderBy with postgresql
I have a problem while trying to custom the createQuery in my admin class : 

An exception has been thrown during the rendering of a template ("An exception occurred while executing 'SELECT DISTINCT c0_.id AS id0, c0_.id AS id1 FROM crm_product_goodplan c1_ INNER JOIN crm_product_product c0_ ON c1_.id = c0_.id LEFT JOIN crm_product_establishment c2_ ON c0_.establishment_id = c2_.id LEFT JOIN crm_product_agent c3_ ON c0_.advisor_id = c3_.id WHERE c0_.endProductionDate >= ? ORDER BY c0_.endProductionDate ASC, c0_.id ASC LIMIT 32 OFFSET 0' with params ["2015-06-02"]:

SQLSTATE[42P10]: Invalid column reference: 7 ERREUR: pour SELECT DISTINCT, ORDER BY, les expressions doivent apparaître dans la
liste SELECT
LINE 1: ...c3_.id WHERE c0_.endProductionDate >= $1 ORDER BY c0_.endPro...
^") in SonataAdminBundle:CRUD:base_list.html.twig at line 33.

this is my function : 

```
{
    $today = new \DateTime(); 
    $query->andWhere('o.endProductionDate >= :today')->setParameter('today', $today->format('Y-m-d'))->orderBy('o.endProductionDate', 'ASC');

    return $query;
}
```

relating to the issue #210

axzx:
I am targetting this branch, because I found bug in this version.

Fixes #458, #210

## Changelog

```markdown
### Changed
- added `orderBy` part to resetDQLparts
```

## Subject

To retrieve the different subjects ids, is not need set orderBy for sorting.
Without this fix database returns an error:
```
SQLSTATE[42P10]: Invalid column reference: 7 ERROR: for SELECT DISTINCT, ORDER BY expressions must appear in select list
LINE 1
```
Full db query:
```
SELECT DISTINCT p0_.id AS id_0, p0_.id AS id_1 FROM product_part_code p0_ LEFT JOIN product_part p1_ ON p0_.part_id = p1_.id ORDER BY p1_.name ASC, p0_.is_active DESC, p0_.id DESC, p0_.id DESC LIMIT 32 OFFSET 0
```

The problem occurs in two cases:
- in `configureFormFields`, when use field `sonata_type_model_autocomplete` and set 'callback'. like this:
```
'callback' => function ($admin, $property, $value) {
    $datagrid = $admin->getDatagrid();
    $queryBuilder = $datagrid->getQuery();
    $queryBuilder
      ->addSelect('p')
      ->join($queryBuilder->getRootAlias() . '.part', 'p')
      ->where('lower(' . $queryBuilder->getRootAlias() . '.name) LIKE :name or lower(p.name) LIKE :name')
      ->orderBy('p.name', 'asc')
      ->setParameter('name', '%' . mb_strtolower($value) . '%')
    ;
},
```
- When I want to manually determine the sorting (`orderBy`). 
```
{
    $query
      ->orderBy('p.name')
      ->addOrderBy('pc.name')
    ;
    return $query;
}
```

### DIFF
--- a/Datagrid/ProxyQuery.php
+++ b/Datagrid/ProxyQuery.php
@@ -309,7 +309,7 @@ class ProxyQuery implements ProxyQueryInterface
             }
             $idxSelect .= ($idxSelect !== '' ? ', ' : '').$idSelect;
         }
-        $queryBuilderId->resetDQLPart('select');
+        $queryBuilderId->resetDQLParts(array('select', 'orderBy'));
         $queryBuilderId->add('select', 'DISTINCT '.$idxSelect);
 
         // for SELECT DISTINCT, ORDER BY expressions must appear in idxSelect list


axzx:
@greg0ire could you check it?

greg0ire:
So you want to get rid of the ORDER BY clause? This means the on two subsequent calls, the list might be completely different. This won't work well with pagination, see https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/544/files#diff-409a051b11d25b4355f7ce9ca60eabd5R79

axzx:
Yes, but only from $queryBuilderId.

greg0ire:
So? What does this query builder do? If it returns the ids of the objects you can view on the current page, there's a problem, because if you don't have an order by clause, you can and should consider this as a random list of ids.

axzx:
So... 
Do you know, how can I solve the problem occurs in two cases?;)

greg0ire:
@axzx : I answered on both cases. Closing this.

axzx:
@greg0ire where?
I still have problem with
```
{
    $query
      ->orderBy($alias.'.confirmed')
      ->addOrderBy($alias.'.status')
    ;
    return $query;
}
```
error:
```
SQLSTATE[HY000]: General error: 3065 Expression #1 of ORDER BY clause is not in SELECT list, references column 'table.s0_.confirmed' which is not in SELECT list; this is incompatible with DISTINCT") in SonataAdminBundle:CRUD:base_list.html.twig at line 30. 
```

greg0ire:
Only in #210 actually

axzx:
so, should be:
```
/**
     *
     * @param string $context
     * @return type
     */
    {

        $query
          ->andWhere($query->expr()->isNotNull($query->getRootAliases()[0] . '.confirmed'))
          ->andWhere($query->expr()->isNotNull($query->getRootAliases()[0] . '.status'))
          ->addSelect($query->getRootAliases()[0] . '.confirmed')
          ->addSelect($query->getRootAliases()[0] . '.status')
          ->orderBy($query->getRootAliases()[0] . '.confirmed', 'asc')
          ->addOrderBy($query->getRootAliases()[0] . '.status', 'asc')
        ;

        return $query;

    }
```
?

This code makes the same error:
```
SQLSTATE[HY000]: General error: 3065 Expression #1 of ORDER BY clause is not in SELECT list, references column 'onelog_cors.s0_.confirmed' which is not in SELECT list; this is incompatible with DISTINCT") in SonataAdminBundle:CRUD:base_list.html.twig at line 30. 
```

I think because select part is reset.

greg0ire:
Look at [the code after the SQL part is reset](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/Datagrid/ProxyQuery.php#L315), there is an interesting comment that says that there is going to be an attempt to fix your problem, but based on what `getSortBy` returns.

greg0ire:
ping @axzx

axzx:
for now, I give up:(

greg0ire:
Oooh :disappointed:

PR #649, opened 2016-12-08
TYPE_NOT_CONTAINS doesn't consider NULL values

samuel4x4:
I am targeting this branch because it's a bugfix.

## Changelog

```markdown
### Fixed
- Consider NULL values when using 'does not contain' advanced string filter
```

## To do

- [ ] Update the tests
- [ ] Update the documentation
- [ ] Add an upgrade note

## Subject

TYPE_NOT_CONTAINS advanced filter should consider also the NULL values. Instead of

$alias.$field NOT LIKE :$parameterName

you should have

$alias.$field NOT LIKE :$parameterName OR $alias.$field IS NULL

If you have the following values for a field 

['Susoo', NULL, 'Bibee']

If you filter by << does not contain 'Susoo' >>, it should return [NULL, 'Bibee'], not only ['Bibee'] as it is implemented right now.

### DIFF
--- a/Filter/StringFilter.php
+++ b/Filter/StringFilter.php
@@ -50,7 +50,7 @@ class StringFilter extends Filter
             $or->add($queryBuilder->expr()->isNull(sprintf('%s.%s', $alias, $field)));
         }
 
-        $this->applyWhere($queryBuilder, $or);        
+        $this->applyWhere($queryBuilder, $or);
 
         if ($data['type'] == ChoiceType::TYPE_EQUAL) {
             $queryBuilder->setParameter($parameterName, $data['value']);
--- a/Filter/StringFilter.php
+++ b/Filter/StringFilter.php
@@ -41,17 +41,17 @@ class StringFilter extends Filter
 
         // c.name > '1' => c.name OPERATOR :FIELDNAME
         $parameterName = $this->getNewParameterName($queryBuilder);
-        
+
         $or = $queryBuilder->expr()->orX();
-        
+
         $or->add(sprintf('%s.%s %s :%s', $alias, $field, $operator, $parameterName));
-        
+
         if ($data['type'] == ChoiceType::TYPE_NOT_CONTAINS) {
             $or->add($queryBuilder->expr()->isNull(sprintf('%s.%s', $alias, $field)));
         }
-        
+
         $this->applyWhere($queryBuilder, $or);        
-        
+
         if ($data['type'] == ChoiceType::TYPE_EQUAL) {
             $queryBuilder->setParameter($parameterName, $data['value']);
         } else {
--- a/Filter/StringFilter.php
+++ b/Filter/StringFilter.php
@@ -41,8 +41,15 @@ class StringFilter extends Filter
 
         // c.name > '1' => c.name OPERATOR :FIELDNAME
         $parameterName = $this->getNewParameterName($queryBuilder);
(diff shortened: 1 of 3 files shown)


Filter/StringFilter.php
```diff
@@ -41,8 +41,15 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
 
         // c.name > '1' => c.name OPERATOR :FIELDNAME
         $parameterName = $this->getNewParameterName($queryBuilder);
-        $this->applyWhere($queryBuilder, sprintf('%s.%s %s :%s', $alias, $field, $operator, $parameterName));
 
+        $whereCondition = sprintf('%s.%s %s :%s', $alias, $field, $operator, $parameterName);
+
+        if ($data['type'] == ChoiceType::TYPE_NOT_CONTAINS) {
+            $whereCondition .= sprintf(' OR %s.%s IS NULL', $alias, $field);
```
Koc: Should we wrap in brackets condition with `OR`?

samuel4x4: What about now? I replaced the OR.

samuel4x4: @Koc What about now? I replaced the OR.

samuel4x4:
#650

core23:
Can you name the BC breaks? IMHO it's a bugfix.

greg0ire:
Please rebase this one too.

samuel4x4:
@greg0ire rebase done

greg0ire:
Good job!

greg0ire:
Please fix StyleCI, ask for help if you don't know how.

Filter/StringFilter.php
```diff
@@ -41,8 +41,17 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
 
         // c.name > '1' => c.name OPERATOR :FIELDNAME
         $parameterName = $this->getNewParameterName($queryBuilder);
-        $this->applyWhere($queryBuilder, sprintf('%s.%s %s :%s', $alias, $field, $operator, $parameterName));
-
+        
+        $or = $queryBuilder->expr()->orX();
```
greg0ire: `orX` does not seem to be recognized, see the Travis build.

greg0ire:
Thanks @samuel4x4 !

greg0ire:
I read a bit more about null because it bothered me that this is not the default, and it means "unknown value", so in fact. When you ask for rows with a property not in a particular set, you don't know if null should be returned or not… so I'm not sure if merging that was the right call after all.

samuel4x4:
Hi @greg0ire,

I understand your concerns, I think it can be different when you compare technical and functional perspectives.

On the database level, I think it's right to not have NULL values in results when your query is "table.field NOT LIKE value". If you want to include NULL values, you should add also "OR table.field IS NULL". But, from a functional point of view, when a client filter in a GUI with "NOT CONTAINS", I see that he should receive everything excepting the rows with "table.field LIKE value".

From the functional perspective, the client has right now these options:

- Contains
- Does not contain
- Is equal to

It can be this way:

- Contains
- Does not contain or is null
- Does not contain and is not null
- Is equal to

but this can confuse the client and actually in my experience I never saw this need. They just want to receive all rows, it doesn't matter if they are null or not, the only condition is to not contain a specific value on a specific field.

greg0ire:
I think you are right. `null` means unknow, and the client needs to be 100% sure that the condition is fulfilled.

PR #654, opened 2017-01-08
Fixed wrong entity link in list view

core23:
I am targetting this branch, because this is a bug fix.

## Changelog

```markdown
### Fixed
 - Fixed wrong link generation when calling `DatagridMapper::addIdentifier` on mapped field

```

## Subject

When calling `DatagridMapper::addIdentifier` on a mapped field, the link is generated to the fields admin edit page instead of the current admin edit page.

### DIFF
--- a/Resources/views/CRUD/list_orm_many_to_one.html.twig
+++ b/Resources/views/CRUD/list_orm_many_to_one.html.twig
@@ -14,7 +14,11 @@ file that was distributed with this source code.
 {% block field %}
     {% if value %}
         {% set route_name = field_description.options.route.name %}
-        {% if field_description.hasAssociationAdmin and field_description.associationadmin.hasRoute(route_name) and field_description.associationadmin.hasAccess(route_name, value) %}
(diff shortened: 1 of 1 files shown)


Resources/views/CRUD/list_orm_many_to_one.html.twig
```diff
@@ -14,7 +14,7 @@ file that was distributed with this source code.
 {% block field %}
     {% if value %}
         {% set route_name = field_description.options.route.name %}
-        {% if field_description.hasAssociationAdmin and field_description.associationadmin.hasRoute(route_name) and field_description.associationadmin.hasAccess(route_name, value) %}
+        {% if not field_description.options.identifier|default(false) and field_description.hasAssociationAdmin and field_description.associationadmin.hasRoute(route_name) and field_description.associationadmin.hasAccess(route_name, value) %}
```
greg0ire: Please linebreak this, it still wouldn't fit if the GH editor was twice as large.

PR #666, opened 2017-02-04
Fixed wrong translation in one_to_many template

core23:
I am targetting this branch, because this is a patch.

## Changelog

```markdown
### Fixed
 - Fixed wrong translation in delete checkbox in `edit_orm_one_to_many_inline_table.html.twig`
```
## Subject

When adding a collection field to the admin form with a sortable position, the `Delete` button will be translated by mistake. The label isn't visible at all in this view.

bildschirmfoto 2017-02-04 um 18 40 35


OskarStark:
Thank you @core23 👍

PR #670, opened 2017-03-25
Remove exceptions from phpunit whitelist

jordisala1991:
I am targeting this branch, because this is pedantic.

## Subject

I think coverage is not calculated correctly, lets see


greg0ire:
Thanks @jordisala1991 !

PR #675, opened 2017-04-03
Edit Inline when using sonata_type_collection with SubClasses is throwing error.

twify93:
When using setSubClasses and using tabs, error is thrown for fields that are not defined in one of the classes.

I am targeting this branch, because using setSubClasses with the inline and sonata_type_collection which uses `edit_orm_one_to_many_inline_tabs.html.twig` throws an error `Key "...." for array with keys "_delete, title, content ...." does not exist `.

## Changelog

```markdown
### Fixed
- don't display fields that are missing in child classes
```

### DIFF
--- a/src/Resources/views/CRUD/edit_orm_one_to_many_inline_tabs.html.twig
+++ b/src/Resources/views/CRUD/edit_orm_one_to_many_inline_tabs.html.twig
@@ -35,18 +35,20 @@ file that was distributed with this source code.
                             <fieldset>
                                 <div class="sonata-ba-collapsed-fields">
                                     {% for field_name in form_group.fields %}
-                                        {% set nested_field = nested_group_field.children[field_name] %}
-                                        <div class="sonata-ba-field-{{ id }}-{{ field_name }}">
-                                            {% if associationAdmin.formfielddescriptions[field_name] is defined %}
-                                                {{ form_row(nested_field, {
-                                                    'inline': 'natural',
-                                                    'edit'  : 'inline'
-                                                }) }}
-                                                {% set dummy = nested_group_field.setrendered %}
-                                            {% else %}
-                                                {{ form_row(nested_field) }}
-                                            {% endif %}
-                                        </div>
+                                        {% if nested_group_field.children[field_name] is defined %}
+                                            {% set nested_field = nested_group_field.children[field_name] %}
+                                            <div class="sonata-ba-field-{{ id }}-{{ field_name }}">
+                                                {% if associationAdmin.formfielddescriptions[field_name] is defined %}
+                                                    {{ form_row(nested_field, {
+                                                        'inline': 'natural',
+                                                        'edit'  : 'inline'
+                                                    }) }}
+                                                    {% set dummy = nested_group_field.setrendered %}
+                                                {% else %}
+                                                    {{ form_row(nested_field) }}
+                                                {% endif %}
+                                            </div>
+                                        {% endif %}
                                     {% endfor %}
                                 </div>
                             </fieldset>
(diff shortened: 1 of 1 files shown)


Resources/views/CRUD/edit_orm_one_to_many_inline_tabs.html.twig
```diff
@@ -35,18 +35,20 @@ file that was distributed with this source code.
...
-                                                {{ form_row(nested_field, {
-                                                    'inline': 'natural',
-                                                    'edit'  : 'inline'
-                                                }) }}
-                                                {% set dummy = nested_group_field.setrendered %}
-                                            {% else %}
-                                                {{ form_row(nested_field) }}
-                                            {% endif %}
-                                        </div>
+                                        {% if nested_group_field.children[field_name] is defined %}
```
greg0ire: I think you can combine the `for` and the `if`, like this : http://twig.sensiolabs.org/doc/2.x/tags/for.html#adding-a-condition

greg0ire:
Fixed the markdown for you.

twify93:
Yeah, maybe don't merge it yet. I still experience problems.
The Add new button in the inline tab view is not a dropdown, meaning that I can only create one of the classes there... Any clue where/how to fix this?

greg0ire:
> The Add new button in the inline tab view is not a dropdown, meaning that I can only create one of the classes there..

Sorry, I have no idea what you mean, maybe a screenshot would help?

twify93:
So I have 2 sub classes in this sortable tab view (edit=inline), but now I need to button to have a dropdown so I can select what type of class to add next.

screen shot 2017-04-03 at 14 18 49

greg0ire:
So… you're creating a collection of pages?

twify93:
No, I have a Page that has different types of Page blocks, made using the Doctrine Discriminator and I want to be able to re-arrange them and add more blocks with the default 'Add new' button either rendering it as a dropdown or adding a select in front of it so you can choose what block you want to add next.

You can also see this issue: https://github.com/sonata-project/SonataAdminBundle/issues/2876

greg0ire:
Well it's a tough one… Maybe this can be solved with a lot of js / ajax?

jordisala1991:
This one is not easy tho, I worked on a project that needed this feature and I managed to get that dropdown showing subclasses, but it didn't work well.

I found another solution using subclasses that works great.

But it would be super nice if this solution works also.

twify93:
If it's not a secret I want to explore that option. Maybe if I add a tab but with separate list view inside somehow..?

jordisala1991:
I remember now. When adding a new element all the children are created again. But add button only sends info about new item, sonata does not know subclass of previously created ones and fails.

twify93:
I've almost got it working, made a lot of changes so I can show a dropdown `Add new` button, now I have this url: `admin/core/append-form-field-element?code=.xxxxx.admin.page&elementId=s58e35b92648b0_blocks&objectId=7&uniqid=s58e35b92648b0&subClass=simpleBlock` which doesn't work as the `subject` isn't correctly set in the Parent `Admin` called `PageBlockAdmin` which leads to the issue that the chosen subclass wont be appended... 

A guy in the issue I referenced before pointed to a commit, which apparently should fix this issue, but it doesn't as the subclass GET parameter isn't used at all in `HelperController::appendFormFieldElementAction` ... Any ideas?
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/commit/501e0dcec17f8d847623b3c73310f258c052e5e2

Tofandel:
Have the same problem with the parent. The parent id is not even sent when the ajax request create the entity. (I'm using a modal with subClasses so i don't actually need the inline feature with the dropdown but if you got the parent to be set and appended in that page, I'm interested in knowing how.

twify93:
I've just rebased, I think it's ready for merge again.

OskarStark:
Thank you @twify93 👍

PR #688, opened 2017-04-25
Update `AddAuditEntityCompilerPass` in order to allow use of decoupled `simplethings/entity-audit-bundle:^2.0@dev`

phansys:
I am targeting this branch, because currently there is an issue when trying to use `simplethings/entity-audit-bundle:^2.0@dev`.

### Fixed
- Fixed `AddAuditEntityCompilerPass::process()` when definition `simplethings.entityaudit.audited_entities` is not present, as of `2.x` version for `simplethings/entity-audit-bundle`.

## To do

- [ ] Validate with core members if this change must be replaced with a `conflict` definition at `composer.json`;
- [ ] Check if this requires an extra build pass using the constraint `simplethings/entity-audit-bundle:^2.0@dev` at `composer.json`.

### DIFF
--- a/DependencyInjection/Compiler/AddAuditEntityCompilerPass.php
+++ b/DependencyInjection/Compiler/AddAuditEntityCompilerPass.php
@@ -11,7 +11,6 @@
 
 namespace Sonata\DoctrineORMAdminBundle\DependencyInjection\Compiler;
 
-use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 
@@ -26,7 +25,7 @@ class AddAuditEntityCompilerPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container)
     {
-        if (!$container->hasDefinition('simplethings_entityaudit.config')) {
+        if (!$container->hasDefinition('simplethings_entityaudit.config') || !$container->hasDefinition('simplethings.entityaudit.audited_entities')) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


core23:
The `2.x` branch isn't supported anymore. 

Also, we don't support unstable dependencies like `simplethings/entity-audit-bundle:^2.0@dev`

phansys:
@core23, I don't adding support for unstable dependencies in this bundle, I'm making possible to use this bundle alongside the unstable dependency in 3rd party projects requiring this bundle.

jordisala1991:
Why do you want to make this PR for 2.x?

phansys:
Because I think there must be no conflicts if I decide to use `sonata-project/doctrine-orm-admin-bundle` without any integration with `simplethings/entity-audit-bundle`, while I need to use `simplethings/entity-audit-bundle:^2.0` separated from the SonataAdmin scope.

jordisala1991:
Yes, but why 2.x of doctrine-orm-admin-bundle?

phansys:
Because I think this fix must be available for the oldest maintained version, and I think the oldest maintained version is `2.x`. Please, correct me if I'm wrong and I could retarget `3.x` or `master`.

jordisala1991:
2.x is not supported anymore, that's why this PR got closed at first.

You should target 3.x

phansys:
Thank you for your feedback @jordisala1991, the target branch is now updated to `3.x`.

jordisala1991:
@core23 I cannot reopen this one, can you? Does it makes sense? I dont use entity-audit so I dont know what are the implications on adding this changes.

phansys:
@core23, have you any concern about this issue?

core23:
Can reopen this either.

If the `simplethings/entity-audit-bundle` 2.x gets stable, you could open a PR for this. Unless this won't make sense to me.

phansys:
Thank you for your concerns @core23.
IMHO, we should then ensure that the current "forced" integration with `simplethings/entity-audit-bundle` just applies to versions "whitelisted" (approved) by `sonata-project/doctrine-orm-admin-bundle`. Otherwise, any user specific requirement outside the built in integration (like the one I have) can not be satisfied.

Issue #655, opened 2017-01-10
add select and having to createQuery in admin class


safwanghamrawi:
#### Sonata packages

```bash
$ composer show sonata-project/*
sonata-project/admin-bundle              dev-develop 4565bdc The missing Symfony Admin Generator (forked from 3.10.2)
sonata-project/block-bundle              3.2.0               Symfony SonataBlockBundle
sonata-project/cache                     1.0.7               Cache library
sonata-project/core-bundle               3.1.2               Symfony SonataCoreBundle
sonata-project/doctrine-orm-admin-bundle 3.1.2               Symfony Sonata / Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/exporter                  1.7.0               Lightweight Exporter library
```

#### Symfony packages

```bash
$ composer show symfony/*
symfony/monolog-bundle     3.0.1  Symfony MonologBundle
symfony/phpunit-bridge     v3.2.1 Symfony PHPUnit Bridge
symfony/polyfill-apcu      v1.3.0 Symfony polyfill backporting apcu_* functions to lower PHP versions
symfony/polyfill-intl-icu  v1.3.0 Symfony polyfill for intl's ICU-related data and classes
symfony/polyfill-mbstring  v1.3.0 Symfony polyfill for the Mbstring extension
symfony/polyfill-php56     v1.3.0 Symfony polyfill backporting some PHP 5.6+ features to lower PHP versions
symfony/polyfill-php70     v1.3.0 Symfony polyfill backporting some PHP 7.0+ features to lower PHP versions
symfony/polyfill-util      v1.3.0 Symfony utilities for portability of PHP codes
symfony/security-acl       v3.0.0 Symfony Security Component - ACL (Access Control List)
symfony/swiftmailer-bundle v2.4.0 Symfony SwiftmailerBundle
symfony/symfony            v3.2.1 The Symfony PHP framework
```

#### PHP version

```bash
$ php -v
PHP 5.6.22-1~dotdeb+7.1 (cli) (built: May 26 2016 22:11:02)
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2016, by Zend Technologies
    with blackfire v1.13.0, https://blackfire.io, by Blackfireio Inc.
```

## Subject

Adding have clause to query in createQuery of my admin class

[...]

PR #694, opened 2017-06-12
Reset order by in ProxyQuery::getFixedQueryBuilder

coyl:
Fix for problems like 
https://github.com/sonata-project/SonataAdminBundle/issues/4332 
and
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/655

`ProxyQuery::getFixedQueryBuilder` resets select to `DISTINCT` and pay no attention to existing `OrderBy` in query builder.
The result – if there was `addOrderBy()` in  `Admin::createQuery`, it would lead to an error like `ORDER BY clause is not in SELECT list, references column 't.s0_.cl' which is not in SELECT list; this is incompatible with DISTINCT`

### DIFF
--- a/Datagrid/ProxyQuery.php
+++ b/Datagrid/ProxyQuery.php
@@ -310,6 +310,7 @@ class ProxyQuery implements ProxyQueryInterface
             $idxSelect .= ($idxSelect !== '' ? ', ' : '').$idSelect;
         }
         $queryBuilderId->resetDQLPart('select');
+        $queryBuilderId->resetDQLPart('orderBy');
         $queryBuilderId->add('select', 'DISTINCT '.$idxSelect);
 
         // for SELECT DISTINCT, ORDER BY expressions must appear in idxSelect list


greg0ire:
@coyl you are pointing to #4332, but you do not explain why you disregard [this comment](https://github.com/sonata-project/SonataAdminBundle/issues/4332#issuecomment-280657778)

coyl:
I couldn't find where the result of this method used, except `base_list.html.twig` just for `admin.datagrid.results|length > 0`

greg0ire:
What does this have to do with the comment above?

coyl:
So orderBy doesn't matter

greg0ire:
> So orderBy doesn't matter

Why? As soon as you have pagination, it matters, right? Otherwise how can you ensure page 2 "follows" page 1, has no intersection with it, etc? If you don't specify an order by, your sql database may return a completely different order on 2 subsequent queries.

greg0ire:
ping @coyl

PR #677, opened 2017-04-03
use createMock() instead of deprecated getMock()

OskarStark:
I am targeting this branch, because this is BC.

PR #696, opened 2017-06-26
Fixed exception on missing field order by 

eugenekurasov:
Closes https://github.com/sonata-project/SonataAdminBundle/issues/4333

## Changelog
```markdown
### Fixed
- Fixed `ModelManager::getDataSourceIterator` when` getSortBy` is empty
```

## To do

- [x] Update the tests

## Subject
If you try use export functionality without ordering and disabled sort primary key you got exception.

### Entity
```php
<?php

namespace AppBundle\Entity;

use Doctrine\ORM\Mapping as ORM;

/**
 * Post
 *
 * @ORM\Table(name="post")
 * @ORM\Entity(repositoryClass="AppBundle\Repository\PostRepository")
 */
class Post
{
    /**
     * @var int
     *
     * @ORM\Column(name="id", type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private $id;

    /**
     * @var string
     *
     * @ORM\Column(name="name", type="string", length=255)
     */
    private $name;


    /**
     * Get id
     *
     * @return int
     */
    public function getId()
    {
        return $this->id;
    }

    /**
     * Set name
     *
     * @param string $name
     *
     * @return Post
     */
    public function setName($name)
    {
        $this->name = $name;

        return $this;
    }

    /**
     * Get name
     *
     * @return string
     */
    public function getName()
    {
        return $this->name;
    }
}


```
### Admin

```php
<?php

namespace AppBundle\Admin;

use Sonata\AdminBundle\Admin\AbstractAdmin;
use Sonata\AdminBundle\Show\ShowMapper;
use Sonata\AdminBundle\Form\FormMapper;
use Sonata\AdminBundle\Datagrid\ListMapper;
use Sonata\AdminBundle\Datagrid\DatagridMapper;

/**
 * Class PostAdmin
 */
class PostAdmin extends AbstractAdmin
{
    /**
     * {@inheritdoc}
     */
    protected function configureFormFields(FormMapper $formMapper)
    {
        $formMapper
            ->add('id', 'int', array(
                'label' => 'Post Title'
            ))
            ->add('name', 'text', array(
                'label' => 'name'
            ))
        ;
    }

    /**
     * {@inheritdoc}
     */
    {
        $datagridMapper
            ->add('id')
            ->add('name')
        ;
    }

    /**
     * {@inheritdoc}
     */
    protected function configureListFields(ListMapper $listMapper)
    {
        $listMapper
            ->addIdentifier('id', null, array(
                'sortable' => false, // If you disabled sortable you got exception
            ))
            ->add('name')
        ;
    }

    /**
     * {@inheritdoc}
     */
    protected function configureShowFields(ShowMapper $showMapper)
    {
        $showMapper
            ->add('id')
            ->add('name')
        ;
    }

}
```

### DIFF
--- a/Model/ModelManager.php
+++ b/Model/ModelManager.php
@@ -509,7 +509,11 @@ class ModelManager implements ModelManagerInterface, LockInterface
         $query->setMaxResults($maxResult);
 
         if ($query instanceof ProxyQueryInterface) {
-            $query->addOrderBy($query->getSortBy(), $query->getSortOrder());
+            $sortBy = $query->getSortBy();
+
+            if (!empty($sortBy)) {
+                $query->addOrderBy($sortBy, $query->getSortOrder());
+            }
 
             $query = $query->getQuery();
         }
--- a/Tests/Model/ModelManagerTest.php
+++ b/Tests/Model/ModelManagerTest.php
@@ -14,6 +14,7 @@ namespace Sonata\DoctrineORMAdminBundle\Tests\Model;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\OptimisticLockException;
+use Doctrine\ORM\Query;
 use Doctrine\ORM\Version;
 use Sonata\DoctrineORMAdminBundle\Admin\FieldDescription;
 use Sonata\DoctrineORMAdminBundle\Model\ModelManager;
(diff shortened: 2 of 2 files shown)


Tests/Model/ModelManagerTest.php
```diff
@@ -403,6 +404,84 @@ public function testAssociationIdentifierType()
         $this->assertSame(42, $result[0]);
     }
 
+    public function getSortableInDataSourceIteratorDataProvider()
+    {
+        return array(
+            array(
+                $sortBy = null,
```
OskarStark: IMO the variables are not needed.

@greg0ire should we remove them or use keys instead?

greg0ire: You are right, they are not needed. Please remove them. Not sure what keys would do, but comments could be used instead.

eugenekurasov: Variables is for code readability. 
Ok I remove variables and add comments.

Tests/Model/ModelManagerTest.php
```diff
@@ -403,6 +404,77 @@ public function testAssociationIdentifierType()
...
+    /**
+     * @dataProvider getSortableInDataSourceIteratorDataProvider
+     *
+     * @param string|null $sortBy
+     * @param string|null $sortOrder
+     * @param bool        $isAddOrderBy
+     */
+    public function testSortableInDataSourceIterator($sortBy, $sortOrder, $isAddOrderBy)
+    {
+        $datagrid = $this->getMockForAbstractClass('Sonata\AdminBundle\Datagrid\DatagridInterface');
```
OskarStark: could you please use `createMock` instead of `getMock` (which is deprecated?

eugenekurasov: Hi, I have installed version 4.8. For run tests I use simple-phpunit. 
Maybe we should made dependency to 5.* PHPunit?

greg0ire: Not until we drop php 5.3

eugenekurasov: I can create mock from builder. 
getMock from builder is not deprecated.

greg0ire: We have a base test class for this IIRC

greg0ire: Not merged yet :( #677

eugenekurasov: I saw this pull request. 
I think I can create mock from builder, and you don't need fixed unit test after merge pull request.  Is solution suitable?

greg0ire:
@OskarStark please review again

Tests/Model/ModelManagerTest.php
OskarStark: Oh you are right, sorry my fault!

OskarStark:
Can you please squash you commits into one? and please decide which account/email you want to use, currently you use 2 different emais

eugenekurasov:
@OskarStark  please review again

Model/ModelManager.php
```diff
@@ -509,7 +509,12 @@ public function getDataSourceIterator(DatagridInterface $datagrid, array $fields
         $query->setMaxResults($maxResult);
 
         if ($query instanceof ProxyQueryInterface) {
-            $query->addOrderBy($query->getSortBy(), $query->getSortOrder());
+            $sortBy = $query->getSortBy();
+            $sortOrder = $query->getSortOrder();
+
+            if (!empty($sortBy) && !empty($sortOrder)) {
+                $query->addOrderBy($query->getSortBy(), $query->getSortOrder());
```
Soullivaneuh: Why calling again `$query->getSortBy()` and `$query->getSortOrder()` here? You can just use the variables.

eugenekurasov:
@Soullivaneuh and @OskarStark please review again

Model/ModelManager.php
```diff
@@ -509,7 +509,12 @@ public function getDataSourceIterator(DatagridInterface $datagrid, array $fields
         $query->setMaxResults($maxResult);
 
         if ($query instanceof ProxyQueryInterface) {
-            $query->addOrderBy($query->getSortBy(), $query->getSortOrder());
+            $sortBy = $query->getSortBy();
+            $sortOrder = $query->getSortOrder();
+
+            if (!empty($sortBy) && !empty($sortOrder)) {
```
Soullivaneuh: Is that necessary to have a sort order?

eugenekurasov: Yes, this case when we sorter list before export.

eugenekurasov: Sorry, but maybe I not understand what you ask? Can you explain what you mean?

eugenekurasov: You was right sortOrder can be null.

eugenekurasov:
I have made change. @Soullivaneuh and @OskarStark please review again.

Soullivaneuh:
Code looks good. I let @sonata-project/contributors decide for the code change logic.

greg0ire:
Thanks @eugenekurasov!

OskarStark:

PR #697, opened 2017-06-26
Prevent null check when relationship is many to many

deguif:
I am targeting this branch, because this is a bug fix.

## Changelog

```markdown
### Fixed
Fixed wrong DQL generated for many to many relationship when filtering with not equals
```

## Subject

Currently when using the `ModelFilter` for a many to many relationship and choosing the not equals from advanced filters, will trigger an Exception.

```
[Semantical Error] line 0, col 134 near 'xxx) IS NUL': Error: Invalid PathExpression. Must be a SingleValuedAssociationField.
```

### DIFF
--- a/Filter/ModelFilter.php
+++ b/Filter/ModelFilter.php
@@ -100,9 +100,15 @@ class ModelFilter extends Filter
 
             $or->add($queryBuilder->expr()->notIn($alias, ':'.$parameterName));
 
-            $or->add($queryBuilder->expr()->isNull(
-                sprintf('IDENTITY(%s.%s)', $this->getParentAlias($queryBuilder, $alias), $this->getFieldName())
-            ));
+            if ($this->getOption('mapping_type') === ClassMetadataInfo::MANY_TO_MANY) {
+                $or->add(
+                    sprintf('%s.%s IS EMPTY', $this->getParentAlias($queryBuilder, $alias), $this->getFieldName())
+                );
+            } else {
+                $or->add($queryBuilder->expr()->isNull(
+                    sprintf('IDENTITY(%s.%s)', $this->getParentAlias($queryBuilder, $alias), $this->getFieldName())
+                ));
+            }
 
             $this->applyWhere($queryBuilder, $or);
         } else {


Filter/ModelFilter.php
```diff
@@ -100,9 +100,11 @@ protected function handleMultiple(ProxyQueryInterface $queryBuilder, $alias, $da
 
             $or->add($queryBuilder->expr()->notIn($alias, ':'.$parameterName));
 
-            $or->add($queryBuilder->expr()->isNull(
-                sprintf('IDENTITY(%s.%s)', $this->getParentAlias($queryBuilder, $alias), $this->getFieldName())
-            ));
+            if (ClassMetadataInfo::MANY_TO_MANY !== $this->getOption('mapping_type')) {
```
greg0ire: You can't assign anything to `$this->getOption('mapping_type')` since it isn't a variable, so the Yoda conditions is unnecessary. Should Yoda be used anyway @sonata-project/contributors ?

deguif: I personally use the filter with this configuration.

```php
    {
        $datagridMapper
            ->add('tags', ModelFilter::class, [
                'mapping_type' => ClassMetadataInfo::MANY_TO_MANY,
            ], null, [
                'class' => ArticleTag::class,
                'choice_label' => function (ArticleTag $value, $key, $index) {
                    return $value->getLabel();
                }
            ])
        ;
    }
```

I'm assigning the value `ClassMetadataInfo::MANY_TO_MANY` to the option.
I don't understand your previous point. Is it just about removing `!==` ?

greg0ire: No, it's about writing the condition with non-Yoda style:

`$this->getOption('mapping_type') !== ClassMetadataInfo::MANY_TO_MANY`

Yoda is "useful" when you risk an unwanted assignment. There is no risk here.

deguif: Got it, sorry. Making the change

greg0ire:
Not 100% sure it is the right fix... are entities w/o any relationship really filtered out after that?

deguif:
You're right, this doesn't return entities without relations.

deguif:
I fixed this wrong behavior with `IS EMPTY` instead of `IS NULL` for many to many relationship
See http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/dql-doctrine-query-language.html#dql-select-examples
Search for `IS EMPTY` on this page

greg0ire:
Nice! TIL!

greg0ire:
@sonata-project/contributors please review

greg0ire:
Thanks @deguif !

OskarStark:

Issue #695, opened 2017-06-23
configureDatagridFilters doesn't recognize fields from a trait


blixit:
Original issue : [https://github.com/sonata-project/SonataAdminBundle/issues/4524](https://github.com/sonata-project/SonataAdminBundle/issues/4524)
@jlamur suggested me to reopen the issue in this repository.

### Environment
Linux

#### Sonata packages

```bash
$ composer show sonata-project/*
sonata-project/admin-bundle              3.19.0 The missing Symfony Admin Generator
sonata-project/block-bundle              3.3.2  Symfony SonataBlockBundle
sonata-project/cache                     1.0.7  Cache library
sonata-project/core-bundle               3.4.0  Symfony SonataCoreBundle
sonata-project/datagrid-bundle           2.2.1  Symfony SonataDatagridBundle
sonata-project/doctrine-orm-admin-bundle 3.1.5  Symfony Sonata / Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/exporter                  1.7.1  Lightweight Exporter library
```

#### Symfony packages

```bash
$ composer show symfony/*
 symfony/monolog-bundle     v3.1.0 Symfony MonologBundle
symfony/phpunit-bridge     v3.3.2 Symfony PHPUnit Bridge
symfony/polyfill-apcu      v1.4.0 Symfony polyfill backporting apcu_* functions to lower PHP versions
symfony/polyfill-intl-icu  v1.4.0 Symfony polyfill for intl's ICU-related data and classes
symfony/polyfill-mbstring  v1.4.0 Symfony polyfill for the Mbstring extension
symfony/polyfill-php56     v1.4.0 Symfony polyfill backporting some PHP 5.6+ features to lower PHP versions
symfony/polyfill-php70     v1.4.0 Symfony polyfill backporting some PHP 7.0+ features to lower PHP versions
symfony/polyfill-util      v1.4.0 Symfony utilities for portability of PHP codes
symfony/security-acl       v3.0.0 Symfony Security Component - ACL (Access Control List)
symfony/swiftmailer-bundle v2.6.2 Symfony SwiftmailerBundle
symfony/symfony            v3.3.2 The Symfony PHP framework
```

#### PHP version

```bash
$ php -v
PHP 7.0.18-0ubuntu0.16.04.1 (cli) ( NTS )
Copyright (c) 1997-2017 The PHP Group
[...]

PR #676, opened 2017-04-03
throw MissingPropertyMetadataException if propertyName could not be found in mappings

OskarStark:
I am targeting this branch, because this is BC.

Closes #599
Closes #695 

## Changelog

```markdown
### Added
- Added `MissingPropertyMetadataException` if property name is not found in field mappings
```

## Subject

throw `MissingPropertyMetadataException` if propertyName could not be found in mappings

This is based on: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/599/files#r69381571

PR #698, opened 2017-06-26
making error more verbose on field's miswriting

blixit:
I am targeting this branch, because the change I'm requesting is minor and the 3.x is the dedicated one.

## Changelog
Throwing a better exception when FilterType guesser doesn't find a field

```markdown
### Changed
- Throw an error if the field doesn't exist
```

## Subject

This is a partial response to the issue I raised here : https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/695  where @jlamur suggested me to do a PR.

Suggested fixes are : 
> * The exception must be more verbose (something like No mapping found for field 'costHt'),
> * The DatagridMapper/Builder should be consistent with other Builders by accepting entity field names (and not column names).

My code doesn't fix the whole issue since the other part is still being investigated. I just offer to make the exception more verbose (something like No mapping found for field 'name')

### DIFF
--- a/Guesser/FilterTypeGuesser.php
+++ b/Guesser/FilterTypeGuesser.php
@@ -66,6 +66,9 @@ class FilterTypeGuesser extends AbstractTypeGuesser
             }
         }
 
+        if( ! $metadata->hasField($propertyName)){
+            throw new \RuntimeException("No mapping found for field '$propertyName' in $class");
+        }
         $options['field_name'] = $metadata->fieldMappings[$propertyName]['fieldName'];
 
         switch ($metadata->getTypeOfField($propertyName)) {
--- a/Guesser/FilterTypeGuesser.php
+++ b/Guesser/FilterTypeGuesser.php
@@ -66,7 +66,7 @@ class FilterTypeGuesser extends AbstractTypeGuesser
             }
         }
 
-        if( ! $metadata->hasField($propertyName)){
+        if (!$metadata->hasField($propertyName)){
             throw new \RuntimeException("No mapping found for field '$propertyName' in $class");
         }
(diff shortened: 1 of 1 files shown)


Guesser/FilterTypeGuesser.php
```diff
@@ -66,6 +66,9 @@ public function guessType($class, $property, ModelManagerInterface $modelManager
             }
         }
 
+        if (!$metadata->hasField($propertyName)) {
+            throw new \RuntimeException("No mapping found for field '$propertyName' in $class");
```
OskarStark: please use `sprintf()` here for the exception message

OskarStark:
@Soullivaneuh this sounds like a BC break to me, isn't it?

Guesser/FilterTypeGuesser.php
greg0ire: Also, please invert quotes and double quotes.

greg0ire: Also, please invert quotes and double quotes

jlamur:
@OskarStark 
> - The exception must be more verbose (something like No mapping found for field 'costHt'),

This is BC,

> - The DatagridMapper/Builder should be consistent with other Builders by accepting entity field names (and not column names).

This is not (be aware that this bug is not confirmed for now, I wasn't able to reproduce it).

jlamur:
IMO, this PR should be closed as #676 fixes the same bug, and is going to be accepted.

Another one could be open to fix the second bug (the *consistency between builders* part) when it will be confirmed.

greg0ire:
Closing in favor of #676, sorry @blixit

blixit:
@greg0ire No it's not a problem. I didn't see this PR before publising mine.

Issue #691, opened 2017-05-17
ChoiceList deprecated in Symfony 3+


fliespl:
## Steps to reproduce
Use symfony 3.2. Create class with discriminator column. Set subclasses on admin class. Add filter to admin class:

`
 ->add('badge_type', 'doctrine_orm_class', array('sub_classes' => $this->getSubClasses()))
`

## Expected results
No error.

## Actual results
Error with importing `Symfony\Component\Form\ChoiceList\ChoiceList`. I twas removed in symfony 3.0.

Should be replaced with `Symfony\Component\Form\ChoiceList\ArrayChoiceList`.

ClassFilter.php

```
Attempted to load class "ChoiceList" from namespace "Symfony\Component\Form\Extension\Core\ChoiceList".
Did you forget a "use" statement for another namespace?
```

OskarStark:
which versions are you using?

fliespl:
"sonata-project/admin-bundle": "3.10.0",
        "sonata-project/doctrine-orm-admin-bundle": "3.1.1",

OskarStark:
can you try to upgrade admin bundle?

fliespl:
```
composer update sonata-project/doctrine-orm-admin-bundle sonata-project/admin-bundle
Loading composer repositories with package information
Updating dependencies (including require-dev)
Nothing to install or update
```

Yet problem lies in ClassFilter inside Doctrine ORM Admin Bundle `Sonata\DoctrineORMAdminBundle\Filter\ClassFilter`. Rewriting getFIeldOptions method to below example works:

```
    /**
     * {@inheritdoc}
     */
    public function getFieldOptions()
    {
        return $this->getOption('choices', array(
            'required' => false,
            'choices' => $this->getOption('sub_classes')
        ));
    }
```

OskarStark:
hmm, but admin-bundle 3.18.2 is currently the latest release

fliespl:
Same happens with such versions:

        "sonata-project/admin-bundle": "3.18.2",
        "sonata-project/doctrine-orm-admin-bundle": "3.1.1",

Jeace:
I can confirm this problem with :

> "name": "sonata-project/admin-bundle",
> "version": "3.18.2",

and

> "name": "sonata-project/doctrine-orm-admin-bundle",
> "version": "3.1.4",

greg0ire:
[...]

PR #700, opened 2017-06-29
Fixed of not exist ChoiceList  in Symfony 3+

eugenekurasov:
Closes [#691](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/691
)

## Changelog

```markdown

### Fixed
Fixed ClassFilter for Symfony 3+. Remove deprecated for Symfony > 2.7

### DIFF
--- a/Filter/ClassFilter.php
+++ b/Filter/ClassFilter.php
@@ -13,7 +13,6 @@ namespace Sonata\DoctrineORMAdminBundle\Filter;
 
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
 use Sonata\CoreBundle\Form\Type\EqualType;
-use Symfony\Component\Form\Extension\Core\ChoiceList\ChoiceList;
 
 class ClassFilter extends Filter
 {
@@ -66,13 +65,20 @@ class ClassFilter extends Filter
      */
     public function getFieldOptions()
     {
-        return $this->getOption('choices', array(
+        $choiceOptions = array(
             'required' => false,
-            'choice_list' => new ChoiceList(
-                \array_values($this->getOption('sub_classes')),
-                \array_keys($this->getOption('sub_classes'))
-            ),
-        ));
+            'choices' => $this->getOption('sub_classes'),
+        );
+
+        // NEXT_MAJOR: Remove (when requirement of Symfony is >= 2.7)
+        if (!method_exists('Symfony\Component\Form\AbstractType', 'configureOptions')) {
+            $choiceOptions['choices'] = array_flip($this->getOption('sub_classes'));
+            // NEXT_MAJOR: Remove (when requirement of Symfony is >= 3.0)
+        } elseif (method_exists('Symfony\Component\Form\FormTypeInterface', 'setDefaultOptions')) {
+            $choiceOptions['choices_as_values'] = true;
+        }
+
+        return $this->getOption('choices', $choiceOptions);
     }
 
     /**

### CODE REVIEW

File: Filter/ClassFilter.php [hunk 1]
```diff
@@ -66,13 +66,24 @@ public function getFieldType()
...
     public function getFieldOptions()
     {
-        return $this->getOption('choices', array(
+        if (!class_exists('Symfony\Component\Form\Extension\Core\ChoiceList\ChoiceList')) {
+            $choiceOption = array(
+                'choices' => $this->getOption('sub_classes'),
+                'choices_as_values' => true, //NEXT_MAJOR: Remove this line when drop Symfony <3.0 support
+            );
+        } else {
+            //NEXT_MAJOR: Remove this condition when drop Symfony <2.7 support
```
greg0ire: "dropping"

Filter/ClassFilter.php
```diff
@@ -66,13 +66,24 @@ public function getFieldType()
      */
     public function getFieldOptions()
     {
-        return $this->getOption('choices', array(
+        if (!class_exists('Symfony\Component\Form\Extension\Core\ChoiceList\ChoiceList')) {
+            $choiceOption = array(
+                'choices' => $this->getOption('sub_classes'),
+                'choices_as_values' => true, //NEXT_MAJOR: Remove this line when drop Symfony <3.0 support
```
greg0ire: "dropping"

Filter/ClassFilter.php
```diff
@@ -66,13 +66,24 @@ public function getFieldType()
...
+            );
+        } else {
+            //NEXT_MAJOR: Remove this condition when drop Symfony <2.7 support
+            $choiceOption = array(
+                'choice_list' => new ChoiceList(
```
greg0ire: Could you please remove `ChoiceList` altogether and do that kind of code instead: https://github.com/sonata-project/SonataAdminBundle/blob/025e8e1092ae80d305dcbeb817efb17e4ee7a4d2/Form/Type/Filter/DateTimeRangeType.php#L77

That should rid people still using sf 2.8 from some unwanted deprecation notices.

Filter/ClassFilter.php
eugenekurasov: We have discussion about this comment in [#691](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/691#issuecomment-311638009). The goal of the comments is to make a note why this needed. You are want that I dropping this comment?

Filter/ClassFilter.php
eugenekurasov: We have discussion about this comment in [#691](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/691#issuecomment-311638009). The goal of the comments is to make a note why this needed. You are want that I dropping this comment?

Filter/ClassFilter.php
eugenekurasov: Thank you for example.
I will update pull request tomorrow.

Filter/ClassFilter.php
greg0ire: No lol. I am suggesting you replace "drop" with "dropping"

eugenekurasov:
@greg0ire Please review again

eugenekurasov:
Sorry for a lot changing of CS. I need more sleep )

jlamur:
You should not mention the issue in the commit title. I think it does not belong there, and it's polluting the issue everytime you modify your commit https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/691#ref-commit-922d33d.

greg0ire:
Please format your commit message according to our rules : 

- The commit message subject must be less than 50 characters and tell us *what* you did.
- The commit message body should tell us *why* you did it. It is optional but highly recommended.

Bad example :

```
Fixed bug #989 by removing call to defraculate()
```

Good example:

```
Remove call to defraculate()

Calling this function caused a bug because it interferes 
with calls to getSchmeckles().
Fixes #989
```

More details in CONTRIBUTING.md

Filter/ClassFilter.php
```diff
@@ -13,10 +13,11 @@
 
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
 use Sonata\CoreBundle\Form\Type\EqualType;
-use Symfony\Component\Form\Extension\Core\ChoiceList\ChoiceList;
 
 class ClassFilter extends Filter
 {
+    const OPTION_SUB_CLASSES = 'sub_classes';
```
greg0ire: Is there any benefit to making this publicly accessible (via `ClassFilter::OPTION_SUB_CLASSES`)?
Do we usually expose options with a constant?

eugenekurasov: I used literal "sub_classes" twice. So I moved "sub_classes" to constant. It is more flexibility than use literal. If you want I do remove it.

jlamur: No `const` in any other `Filter` class. It could be better to use an `OptionResolver` for options. Anyway, this does not belong in this PR.

eugenekurasov: For example [sonar](https://www.sonarqube.org/) is throw warning if he detected use twice literal.

eugenekurasov: @jlamur you are right, this code not belong in this PR. I will remove const.

greg0ire: @eugenekurasov ok, but this pollutes the global scope. If you want to avoid repeating literals, please just use a variable (technically, a variable that never changes is... a constant). Note that with php 7.1 you could make the constant private, which would be a bit better, but would still make the constant available to the rest of the class.

greg0ire: To be clear, what's not great about this is that the const will no longer be needed in next major when we simplify this, but removing it will be a BC-break. Removing a local variable isn't.

eugenekurasov:
Changed commit message and removed const.
@greg0ire and @jlamur please review again.

greg0ire:
Great job!

jlamur:
Wouldn't it be cleaner to use the `choice_loader' option for SF 3+ instead of *tricky* functions checks on different classes / interfaces? Not a requirement but IMO easier to understand.

greg0ire:
TIL about `choice_loader`, but the docs say: 

> The choice_loader can be used to only partially load the choices in cases where a fully-loaded list is not necessary. This is only needed in advanced cases and would replace the choices option.

I don't think our use case matches, because people hopefully don't have lots of subclasses.

eugenekurasov:
"choice_loader" is not "choices". I think class filter should be same in different versions. PR is for fix ClassFilter

greg0ire:
Also, bear in mind that on master, we are going to remove these tricks, and we should end up with just a `choices` option, which will be more straightforward. And the closer we are to the final state of the code, the better IMO.

jlamur:
@greg0ire @eugenekurasov You're both right. PR is ok for me. :)

eugenekurasov:
@jlamur if all ok, can you give approved

greg0ire:
Thanks @eugenekurasov !

OskarStark:

PR #702, opened 2017-07-07
Add slash to chars '%' and '_' on LIKE for StringFilter

### LINKED ISSUE #138: StringFilter does not escape % (percent sign) or _ (underscore)
It seems that Sonata\DoctrineORMAdminBundle\Filter\StringFilter does not filter % or _

I have tried something this : 

``` php
if ($data['type'] == ChoiceType::TYPE_EQUAL) {
    $this->applyWhere($queryBuilder, sprintf('%s.%s %s :%s', $alias, $field, $operator, $parameterName));
} else {
    if ($data['type'] == ChoiceType::TYPE_CONTAINS) {
        $where = $queryBuilder->expr()->like(sprintf('%s.%s', $alias, $field), sprintf(':%s', $parameterName));
    } else if ($data['type'] == ChoiceType::TYPE_NOT_CONTAINS) {
        $where = $queryBuilder->expr()->not($queryBuilder->expr()->like(sprintf('%s.%s', $alias, $field), sprintf(':%s', $parameterName)));
    }
    $this->applyWhere($queryBuilder, $where);
    $queryBuilder->setParameter($parameterName, sprintf($this->getOption('format'), $queryBuilder->expr()->literal($data['value'], 'string')));
}
```

This example uses : 
- use expr()->like to create the like
- use expr()->literal to escape _, %
  but literal does not work with the oci8 driver that I use.

Any idea ?

eugenekurasov:
StringFilter does not filter '%' or '_'

Closes #138

## Changelog
Added filter '%' or '_' for StringFilter  on LIKE operation

```markdown
### Fixed
Search with special chars '%' or '_' on StringFilter was not available
```

### DIFF
--- a/Filter/StringFilter.php
+++ b/Filter/StringFilter.php
@@ -55,7 +55,10 @@ class StringFilter extends Filter
         if ($data['type'] == ChoiceType::TYPE_EQUAL) {
             $queryBuilder->setParameter($parameterName, $data['value']);
         } else {
-            $queryBuilder->setParameter($parameterName, sprintf($this->getOption('format'), $data['value']));
+            $queryBuilder->setParameter($parameterName, sprintf(
+                $this->getOption('format'),
+                addcslashes($data['value'], '%_')
+            ));
         }
     }
 


Filter/StringFilter.php
```diff
@@ -55,7 +55,10 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
         if ($data['type'] == ChoiceType::TYPE_EQUAL) {
             $queryBuilder->setParameter($parameterName, $data['value']);
         } else {
-            $queryBuilder->setParameter($parameterName, sprintf($this->getOption('format'), $data['value']));
+            $queryBuilder->setParameter($parameterName, sprintf(
+                $this->getOption('format'),
+                addcslashes($data['value'], '%_')
```
core23: add*cs*lashes?

eugenekurasov: Thank you

eugenekurasov: http://php.net/manual/en/function.addcslashes.php

jlamur: What if someone currently uses '%' and '_' chars as a feature for searching? It's not BC.
Also, '\\%' is escaped to '\\\\%', which evaluates as *"search a backslash followed by anything else"*. The escaping strategy is wrong.

eugenekurasov: Hi, you are right, it is can used by someone like feature.
I think I can add new option which will be enable/disable escaping. 
Escaping strategy I will fix.

eugenekurasov:
I will create new pull request.

PR #706, opened 2017-07-24
[Fix] add missing sonata-ba-field-error class

dmaicher:
Closes https://github.com/sonata-project/SonataAdminBundle/issues/4533

## Changelog

```markdown
### Fixed
- added the missing `sonata-ba-field-error` class to table fields with errors
```

## Subject

The class `sonata-ba-field-error` was missing so that sub-forms were not correctly handled when inside a tab here: https://github.com/sonata-project/SonataAdminBundle/blob/3.x/Resources/public/Admin.js#L329


greg0ire:
Thanks @dmaicher !

PR #715, opened 2017-08-22
Fix getEntityManager deprecation

deguif:
I am targeting this branch, because this is about avoiding deprecation warning.

## Changelog

```markdown
### Changed
- Replaced `getEntityManager` with `getManager`
```

## Subject

Since Symfony 2.1 `Registry::getEntityManager()` method is deprecated and should be replaced by `Registry::getManager()`


OskarStark:
Thank you @deguif 👍

PR #721, opened 2017-08-25
Fix invalid FieldDescription for association embedded properties

aurelien-vtech:
I am targeting this branch, because it's BC bug fix.

## Changelog

```markdown
### Fixed
- Fixed invalid FieldDescription for association embedded properties
```

## To do

- [x] Update the tests

## Subject

Without this fix the ModelManager is not able to generate valid FieldDescription for association embedded properties.

### DIFF
--- a/Model/ModelManager.php
+++ b/Model/ModelManager.php
@@ -82,8 +82,10 @@ class ModelManager implements ModelManagerInterface, LockInterface
     {
         $nameElements = explode('.', $propertyFullName);
         $lastPropertyName = array_pop($nameElements);
+        $propertyName = $lastPropertyName;
         $class = $baseClass;
         $parentAssociationMappings = array();
+        $parentEmbeddedMappings = array();
 
         foreach ($nameElements as $nameElement) {
             $metadata = $this->getMetadata($class);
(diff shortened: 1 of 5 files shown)


greg0ire:
Please write a changelog directed at end users. Keep in mind that they probably do not care what file or class you changed, but what features you provided / what bugs you fixed. Try to answer the question "What benefit do I get from upgrading?"
More on http://keepachangelog.com

greg0ire:
Please kindly squash your commits together, the code shouldn't live w/o the tests. If you don't, we'll try to remember to do it for you but it's best if you save us this trouble.

How to do that?

1. `git rebase -i origin/3.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
2. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one
3. Close your editor, git should do its magic, and you should end up with one commit
4. Use `git push --force` to overwrite what you already pushed. Don't forget the `--force` option otherwise git will try to merge both things together.

Model/ModelManager.php
```diff
@@ -92,13 +94,26 @@ public function getParentMetadataForProperty($baseClass, $propertyFullName)
...
+                    'field' => $nameElement,
+                );
+                $class = $metadata->embeddedClasses[$nameElement]['class'];
             }
         }
 
-        return array($this->getMetadata($class), $lastPropertyName, $parentAssociationMappings);
+        // Support for multi-level embedded properties
+        if (!empty($parentEmbeddedMappings)) {
+            $class = $parentEmbeddedMappings[0]['class'];
```
greg0ire: Please use `current` over `[0]`, it is more robust.

Model/ModelManager.php
```diff
@@ -92,13 +94,26 @@ public function getParentMetadataForProperty($baseClass, $propertyFullName)
...
+
+            $propertyName = '';
+            foreach ($parentEmbeddedMappings as $embeddedMapping) {
+                $propertyName .= $embeddedMapping['field'].'.';
+            }
```
greg0ire: Please use `array_column` and `implode` instead.

greg0ire: nvm, `array_column` is only since php 5.5, we can't afford to use it yet.

Model/ModelManager.php
aurelien-vtech: Done

greg0ire:
Thanks @aurelien-vtech !

Model/ModelManager.php
```diff
@@ -92,13 +94,27 @@ public function getParentMetadataForProperty($baseClass, $propertyFullName)
...
                 $class = $metadata->getAssociationTargetClass($nameElement);
             } elseif (isset($metadata->embeddedClasses[$nameElement])) {
-                $parentAssociationMappings = array();
-                $lastPropertyName = $propertyFullName;
-                $class = $baseClass;
+                $parentEmbeddedMappings[] = array(
+                    'class' => $class,
+                    'field' => $nameElement,
+                );
+                $class = $metadata->embeddedClasses[$nameElement]['class'];
```
Andrej-in-ua: This line corrupted code, when has multi-level embedding (embedded_first.embedded_second.field).
Because impossible to obtain metadata from embedded class in doctrine.
Is this a bug, or am I mistaken?

jordisala1991: Not sure, can you create an issue, or if you have a fix for this a PR?

PR #730, opened 2017-09-26
Allow to set Query hint in createQuery

dmarkowicz:
I am targeting this branch, because it is BC feature.

Closes sonata-project/SonataTranslationBundle#162

## Changelog

```markdown
### Added
- Added `ProxyQuery::setHint` which allow to pass Query hint in createQuery
```

## Subject

Right now it is impossible to setHint on Query:
```php
{
    $proxyQuery = parent::createQuery($context);

    $proxyQuery->setHint(
        \Doctrine\ORM\Query::HINT_CUSTOM_OUTPUT_WALKER,
        'Gedmo\\Translatable\\Query\\TreeWalker\\TranslationWalker'
    );

    return $proxyQuery;
}
```
I would like to reduce number of database queries. Using the TranslationBundle it is additional query per row, as described here sonata-project/SonataTranslationBundle#162. `setHint` has the same syntax as in `Doctrine\ORM\AbstractQuery`.
Let me know what do you think about this idea.

### DIFF
--- a/Datagrid/ProxyQuery.php
+++ b/Datagrid/ProxyQuery.php
@@ -48,6 +48,13 @@ class ProxyQuery implements ProxyQueryInterface
     protected $entityJoinAliases;
 
     /**
+     * The map of query hints.
+     *
+     * @var array<string,mixed>
+     */
+    private $hints = array();
+
+    /**
      * @param QueryBuilder $queryBuilder
      */
     public function __construct($queryBuilder)
@@ -132,7 +139,12 @@ class ProxyQuery implements ProxyQueryInterface
             }
         }
 
-        return $this->getFixedQueryBuilder($queryBuilder)->getQuery()->execute($params, $hydrationMode);
+        $query = $this->getFixedQueryBuilder($queryBuilder)->getQuery();
+        foreach ($this->hints as $name => $value) {
+            $query->setHint($name, $value);
+        }
+
+        return $query->execute($params, $hydrationMode);
     }
 
     /**
@@ -270,6 +282,24 @@ class ProxyQuery implements ProxyQueryInterface
         }
 
         return $alias;
+    }
+
+    /**
+     * Sets a {@see \Doctrine\ORM\Query} hint. If the hint name is not recognized, it is silently ignored.
+     *
+     * @param string $name  the name of the hint
+     * @param mixed  $value the value of the hint
+     *
+     * @return ProxyQueryInterface
+     *
+     * @see \Doctrine\ORM\Query::setHint
+     * @see \Doctrine\ORM\Query::HINT_CUSTOM_OUTPUT_WALKER
+     */
+    final public function setHint($name, $value)
+    {
+        $this->hints[$name] = $value;
+
+        return $this;
     }
 
     /**
--- a/Tests/Datagrid/ProxyQueryTest.php
+++ b/Tests/Datagrid/ProxyQueryTest.php
@@ -16,10 +16,12 @@ use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\Query\Expr\From;
 use Doctrine\ORM\Query\Expr\OrderBy;
 use Doctrine\ORM\Tools\SchemaTool;
+use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\DoctrineType\UuidType;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Util\NonIntegerIdentifierTestClass;
 use Sonata\DoctrineORMAdminBundle\Tests\Helpers\PHPUnit_Framework_TestCase;
 use Symfony\Bridge\Doctrine\Test\DoctrineTestHelper;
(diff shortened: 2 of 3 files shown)


Datagrid/ProxyQuery.php
```diff
@@ -132,7 +139,27 @@ public function execute(array $params = array(), $hydrationMode = null)
...
+
+    /**
+     * Sets a query hint. If the hint name is not recognized, it is silently ignored.
+     *
+     * @param string $name  the name of the hint
+     * @param mixed  $value the value of the hint
+     *
+     * @return ProxyQueryInterface
+     */
+    public function setHint($name, $value)
```
core23: final

Datagrid/ProxyQuery.php
```diff
@@ -48,6 +48,13 @@ class ProxyQuery implements ProxyQueryInterface
     protected $entityJoinAliases;
 
     /**
+     * The map of query hints.
+     *
+     * @var array
+     */
+    protected $hints = array();
```
greg0ire: Should be private

dmarkowicz:
Ready, please review.

Resources/doc/reference/query_proxy.rst
```diff
@@ -27,3 +27,14 @@ The ``ProxyQuery`` object is used to add missing features from the original `Doc
     $proxyQuery->setMaxResults(10);
 
     $results = $proxyQuery->execute();
+
+* add query hint in advanced usage
```
jlamur: Doesn't belong in this list. The header of this list is:

> The `ProxyQuery` object is used to add missing features from the original `Doctrine Query` builder

But obviously this is not a missing feature.

Anyway, there is no need to document it at all IMO.

Tests/Datagrid/ProxyQueryTest.php
```diff
@@ -185,4 +188,39 @@ public function testAddOrderedColumns()
...
+        );
+        $pq->setHint('hint', 'value');
+
+        $result = $pq->execute();
+
+        $this->assertEquals(2, $result[0]['id']);
+    }
+}
+
+class TestWalker extends SqlWalker
```
jlamur: Move this class out of this file.

Datagrid/ProxyQuery.php
```diff
@@ -48,6 +48,13 @@ class ProxyQuery implements ProxyQueryInterface
     protected $entityJoinAliases;
 
     /**
+     * The map of query hints.
+     *
+     * @var array
```
greg0ire: Is it an associative array? How many dimensions? What does it contain? please describe.

Datagrid/ProxyQuery.php
```diff
@@ -273,6 +285,21 @@ public function entityJoin(array $associationMappings)
...
 
     /**
+     * Sets a query hint. If the hint name is not recognized, it is silently ignored.
+     *
+     * @param string $name  the name of the hint
+     * @param mixed  $value the value of the hint
+     *
+     * @return ProxyQueryInterface
+     */
+    final public function setHint($name, $value)
```
greg0ire: Do you think this method should be added to the `ProxyQueryInterface` (commented out until next major) ? Or is it too big and should it go to a separate interface? Or in no interface?

Tests/Fixtures/Query/FooWalker.php
```diff
@@ -0,0 +1,15 @@
...
+
+use Doctrine\ORM\Query\SqlWalker;
+
+class FooWalker extends SqlWalker
+{
+    public function walkOrderByClause($orderByClause)
+    {
+        $sql = parent::walkOrderByClause($orderByClause);
+
+        return str_replace(' ASC', ' DESC', $sql);
```
greg0ire: `return str_replace(' ASC', ' DESC', parent::walkOrderByClause($orderByClause));` is more concise.

Datagrid/ProxyQuery.php
dmarkowicz: It is ORM specific method. For example Doctrine\ODM\PHPCR\Query\Query does not have `setHint` at all. IMHO `ProxyQueryInterface` should stay generic.

Tests/Fixtures/Query/FooWalker.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Query;
+
+use Doctrine\ORM\Query\SqlWalker;
+
+class FooWalker extends SqlWalker
```
jlamur: `final` here.

Datagrid/ProxyQuery.php
dmarkowicz: I added types in phpdoc and references to `\Doctrine\ORM\Query`. Is it good enough? We can pass one of [predefined hints](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/dql-doctrine-query-language.html#query-hints) or define [custom](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/cookbook/dql-custom-walkers.html#modify-the-output-walker-to-generate-vendor-specific-sql). It is associative array, same behaviour as in `Query`. The `name` is always a string, type of `value` depends on hint.

jlamur:
Thanks @dmarkowicz!

PR #731, opened 2017-10-01
Change minimum doctrine/orm version on master

dmarkowicz:
I am targeting master branch, to prove that it works. Afterwards it should be targeted 3.x.

## Changelog

```markdown
### Changed
- Change minimum doctrine/orm version to 2.4.5
```

## To do

- [ ] Target PR to 3.x

## Subject

#730 test case revealed strange Doctrine bug in version < 2.4.5 https://github.com/doctrine/doctrine2/pull/1142. I wonder how it passed 3.x tests because now it does not pass them too (in my local env). If it pass tests on master it should be retargeted to 3.x.

<details><summary>See stack trace with --prefer-lowest</summary>
<p>

```
..................................E............................... 65 / 94 ( 69%)
......................S..WWWW.                                     94 / 94 (100%)

Time: 5.5 seconds, Memory: 38.00MB

There was 1 error:

1) Sonata\DoctrineORMAdminBundle\Tests\Datagrid\ProxyQueryTest::testSetHint
Doctrine\ORM\Query\QueryException: [Syntax Error] line 0, col 82: Error: Expected Literal, got 'ORDER'

/home/travis/build/sonata-project/SonataDoctrineORMAdminBundle/vendor/doctrine/orm/lib/Doctrine/ORM/Query/QueryException.php:44
/home/travis/build/sonata-project/SonataDoctrineORMAdminBundle/vendor/doctrine/orm/lib/Doctrine/ORM/Query/Parser.php:396
/home/travis/build/sonata-project/SonataDoctrineORMAdminBundle/vendor/doctrine/orm/lib/Doctrine/ORM/Query/Parser.php:2363
/home/travis/build/sonata-project/SonataDoctrineORMAdminBundle/vendor/doctrine/orm/lib/Doctrine/ORM/Query/Parser.php:2550
/home/travis/build/sonata-project/SonataDoctrineORMAdminBundle/vendor/doctrine/orm/lib/Doctrine/ORM/Query/Parser.php:2485
/home/travis/build/sonata-project/SonataDoctrineORMAdminBundle/vendor/doctrine/orm/lib/Doctrine/ORM/Query/Parser.php:2453
/home/travis/build/sonata-project/SonataDoctrineORMAdminBundle/vendor/doctrine/orm/lib/Doctrine/ORM/Query/Parser.php:2427
/home/travis/build/sonata-project/SonataDoctrineORMAdminBundle/vendor/doctrine/orm/lib/Doctrine/ORM/Query/Parser.php:2414
[...]

PR #732, opened 2017-10-01
Change minimum doctrine/orm version

dmarkowicz:
I am targeting this branch, because it is bugfix.

Closes #731

## Changelog

```markdown
### Changed
- Change minimum doctrine/orm version to 2.4.5 because QueryBuilder bug on PHP7 and HHVM
```

## Subject

#730 test case revealed Doctrine bug in version < 2.4.5 on master branch, see #731. But I noticed that 3.x is affected too. Test fail on my local env in 3.x.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     ],
     "require": {
         "php": "^5.3 || ^7.0",
-        "doctrine/orm": "^2.3",
+        "doctrine/orm": "^2.4.5",
         "sonata-project/admin-bundle": "^3.1",
         "sonata-project/core-bundle": "^3.0",
         "sonata-project/exporter": "^1.3.1",


greg0ire:
> - Change minimum doctrine/orm version to 2.4.5

People will read that and think: "why" ?

dmarkowicz:
Updated changelog.

core23:
Does this bug affect our bundles? So do WE need this fix?

dmarkowicz:
Yes, tests with `--prefer-lowest` flag fails on current master https://travis-ci.org/sonata-project/SonataDoctrineORMAdminBundle/builds/281103513 and on my local PHP 7.1 in 3.x branch. Normally composer use at least 2.5.11 version. 2.4.5 is pretty old but sufficient.

dmarkowicz:
To be precise. We do not need this because we will never use such old version. Travis need it to pass the tests with --prefer-lowest.

dmarkowicz:
@core23 ok, I fully understand what happened here with Travis tests.
master branch is tested on: `PHP 7.1 --prefer-lowest`
3.x branch is tested on: `PHP 5.3 --prefer-lowest`
So, this fix is a must for master branch. And is optional for 3.x branch. However it could be backported to 3.x. It is up to you.

Because I have PHP 7.1 on my development env, tests 3.x branch will always fail with --prefer-lowest, and will be passing on PHP 5.x. Because bug affects only PHP 7.x.

greg0ire:
Thanks @dmarkowicz!

PR #740, opened 2017-10-18
Reset page to first 3.1.4

mihail-intexsys:
## Changelog

```markdown

### Changed
- Added check if offset is higher or equals result count and forced to show 
results from the first page in such case.

### Fixes
- Fixes a bug when all items were deleted(with a batch deletion) from the second page on the admin list and we were redirected to the same page with empty result list and without page selector

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.1.4](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/compare/3.1.3...3.1.4) - 2017-03-23
+### Fixed
+- Fixed broken translation in oneToMany table view
+- Fixed wrong translation in delete checkbox in `edit_orm_one_to_many_inline_table.html.twig`
+
+### Security
+- Fixed view - check specific item collection, not to the whole collection.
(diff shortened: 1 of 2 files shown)


greg0ire:
I'd rather do this in the controller of the batch action, it will be easier to understand and will benefit other storage engines.

mihail-intexsys:
There is also few more cases which are not related to batch deletion. 

- You can also delete 33rd item with 32 per page (first and only item on the second page) from the edit form and you will be redirected to empty page.
- You can share a url with applied filters on it (page and per page size) from the last page with someone and after time if items from the last page were deleted your url will be broken and behave like in previous described cases.

greg0ire:
> You can share a url with applied filters on it (page and per page size) from the last page with someone and after time if items from the last page were deleted your url will be broken and behave like in previous described cases.

I'd expect this to give a 404

> You can also delete 33rd item with 32 per page (first and only item on the second page) from the edit form and you will be redirected to empty page.

I don't understand how

mihail-intexsys:
Actually we managed to fix this issue in the listAction of extended sonataAdminBundle CRUDController , just redirecting to the first page if there is 0 results on current non first page, pretty similar like @greg0ire suggested in the first comment. I think pull request can be declined.

greg0ire:
Thanks for following up on this!

PR #747, opened 2017-11-05
fix: errors field not found

ebuildy:
Inside a ``sonata_type_collection``, using a button to delete  I got the exception:

> Key "errors" for array with keys "value, attr, form, id, name, full_name, disabled, label, label_format, multipart, block_prefixes, unique_block_prefix, translation_domain, cache_key" does not exist.

This commit fix the issue. Here the ``configureFormField`` extract:

```php
->add('synonyms', 'sonata_type_collection', [
                    'label' => false,
                    'required' => false,
                    'help' => 'Synonyms or translations',
                    'type_options' => [
                        'delete' => true,
                        'delete_options' => array(
                            // You may otherwise choose to put the field but hide it
                            'type'         => 'button',
                            // In that case, you need to fill in the options as well
                            'type_options' => array(

                            )
                        )
                    ]
                ], [
                    'edit' => 'inline',
                    'inline' => 'table'
                ])
```

### DIFF
--- a/Resources/views/CRUD/edit_orm_one_to_many_inline_table.html.twig
+++ b/Resources/views/CRUD/edit_orm_one_to_many_inline_table.html.twig
@@ -38,7 +38,7 @@ file that was distributed with this source code.
                     <td class="
                         sonata-ba-td-{{ id }}-{{ field_name  }}
                         control-group
-                        {% if nested_field.vars.errors|length > 0 %} error sonata-ba-field-error{% endif %}
+                        {% if nested_field.vars.errors is defined and nested_field.vars.errors|length > 0 %} error sonata-ba-field-error{% endif %}
                         "
                         {% if (nested_field.vars['attr']['hidden'] is defined) and (nested_field.vars['attr']['hidden']) %}
                             style="display:none;"
@@ -55,7 +55,7 @@ file that was distributed with this source code.
                                 {{ form_widget(nested_field) }}
                             {% endif %}
                         {% endif %}
-                        {% if nested_field.vars.errors|length > 0 %}
+                        {% if nested_field.vars.errors is defined and nested_field.vars.errors|length > 0 %}
                             <div class="help-inline sonata-ba-field-error-messages">
                                 {{ form_errors(nested_field) }}
                             </div>


Resources/views/CRUD/edit_orm_one_to_many_inline_table.html.twig
```diff
@@ -38,7 +38,7 @@ file that was distributed with this source code.
                     <td class="
                         sonata-ba-td-{{ id }}-{{ field_name  }}
                         control-group
-                        {% if nested_field.vars.errors|length > 0 %} error sonata-ba-field-error{% endif %}
+                        {% if nested_field.vars.errors is defined and nested_field.vars.errors|length > 0 %} error sonata-ba-field-error{% endif %}
```
greg0ire: I think `if nested_field.vars.errors|default([])|length > 0` would be better.

Resources/views/CRUD/edit_orm_one_to_many_inline_table.html.twig
```diff
@@ -55,7 +55,7 @@ file that was distributed with this source code.
                                 {{ form_widget(nested_field) }}
                             {% endif %}
                         {% endif %}
-                        {% if nested_field.vars.errors|length > 0 %}
+                        {% if nested_field.vars.errors is defined and nested_field.vars.errors|length > 0 %}
```
greg0ire: Same here

greg0ire:
Hey there!

It looks like you did not fill the pull request template properly. Maybe you
were confused, so let me walk you through this step by step.

# Explaining why you target this branch

> I am targeting this branch, because {reason}.

Here, you are supposed to replace `{reason}` with something like:

- "there is a BC-break" if you picked the master branch
- "there is no BC-break" if you picked the stable branch

# Searching and finding related issues

> Closes #{put_issue_number_here}

Please have a look at the list of issues and pull requests, and look for issues
that will be fixed, and pull requests that will become irrelevant should we
merge your PR. Each issue / pull request has a number shown right next to the
title. Kindly replace `{put_issue_number_here}` with whatever issue or PR is
relevant. Of course, you may specify several numbers.
On merge, [these issues and PRs will automatically get
closed](https://help.github.com/articles/closing-issues-using-keywords/).
If there is no issue or pull request that fits there, just remove the sentence
entirely.

# Filling the changelog

That one is really important to get right, because it will be used to craft a
release changelog such as [this
one](https://github.com/sonata-project/SonataAdminBundle/releases/tag/3.23.0).
The first thing you need to know about this is that if your pull request is
pedantic, meaning it does not fix a bug for the user, or adds a new feature, it
is not needed. This is typically the case for PRs that fix the build or add
some documentation.

3 important things:

1. Make sure to fill the changelog section with something directed at the end
user, that answers the question "What benefit do I get from upgrading?".
2. Make sure to remove empty sections.
3. Make sure to keep the markdown markup.

## Filling the TODO list

If you think your PR is ready, just remove it. If you have still some things
left to do, customize it.

## Filling the Subject

Last but not least, describe your PR as well as you can. Bear in mind that some
reviewers do not even use the library they are maintaining, so try to be
crystal clear.

Resources/views/CRUD/edit_orm_one_to_many_inline_table.html.twig
ebuildy: It's not a problem in performance point of view? default will create an empty array for nothing, and this code run into a loop.

greg0ire: I think you won't have any issue finding code that dwarves this potential problem.

OskarStark:
Thank you for your contribution @ebuildy 

If you would like to finish this PR, please reopen.

PR #750, opened 2017-11-08
Fixed deprecated method calls

core23:

## Changelog

```markdown
### Fixed
- Fixed deprecated method calls
```

## Subject

This should fix https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/745


greg0ire:
You can cast to array to avoid the error messages you are getting :)

PR #761, opened 2017-11-27
Drop old Symfony code

jordisala1991:

### Subject

Removing old code that never gets executed because we dropped SF 2.3 to 2.7


src/Builder/DatagridBuilder.php
```diff
@@ -126,8 +128,7 @@ public function addFilter(DatagridInterface $datagrid, $type, FieldDescriptionIn
 
         $fieldDescription->mergeOption('field_options', ['required' => false]);
 
-        // NEXT_MAJOR: Remove first check (when requirement of Symfony is >= 2.8)
```
OskarStark: which check was removed?

jordisala1991: None, it cannot be removed, will add this comment back

jordisala1991: Fixed!

greg0ire:
Thanks @jordisala1991 !

PR #708, opened 2017-07-29
sonata_type_model_list edit button in dialog

kunicmarko20:
I am targeting this branch, because I am adding an additional feature.

Closes #707

## Changelog
```markdown
### Added
- Added edit button that opens in dialog instead of add if there is object already in sonata type model list
```

## To do

- [x] Update the tests
- [x] Update the documentation
- [x] Add `btn_edit` option to sonata_type_model_list form type in SonataAdminBundle

## Subject

Instead of add button that would replace the current object with a new one, add edit button that would open the current object in dialog instead of redirecting to edit screen of that object.

### Before
deepinscreenshot_select-area_20170815210554
deepinscreenshot_select-area_20170815210950

### After
deepinscreenshot_select-area_20170815205438
deepinscreenshot_select-area_20170815205456

PR #765, opened 2017-12-01
sonate_admin.value shouldn't be a condition for the add button

n3o77:
I am targeting this branch, because it's a non BC bugfix.

## Changelog

```markdown
### Fixed
Allow to `add` a new Model even if one is already selected
```

## To do
- [x] Update the tests
- [ ] Add an upgrade note

## Subject

Introduced with #708 was an additional condition for the `add` button which shouldn't be there. This prevents creating new models if one is already selected. If this behaviour is still needed it can be archived in the Admin::configureFormFields method simply by setting the `btn_add` option to `false`.

### DIFF
--- a/src/Resources/views/Form/form_admin_fields.html.twig
+++ b/src/Resources/views/Form/form_admin_fields.html.twig
@@ -85,7 +85,6 @@ file that was distributed with this source code.
                 {% if sonata_admin.field_description.associationadmin.hasroute('create')
                     and sonata_admin.field_description.associationadmin.isGranted('CREATE')
                     and btn_add
-                    and sonata_admin.value == null
                 %}
                     <a  href="{{ sonata_admin.field_description.associationadmin.generateUrl('create') }}"
                         onclick="return start_field_dialog_form_add_{{ id }}(this);"
@@ -95,7 +94,8 @@ file that was distributed with this source code.
                         <i class="fa fa-plus-circle"></i>
                         {{ btn_add|trans({}, btn_catalogue) }}
                     </a>
-                {% elseif sonata_admin.field_description.associationadmin.hasroute('edit')
+                {% endif %}
+                {% if sonata_admin.field_description.associationadmin.hasroute('edit')
                     and sonata_admin.field_description.associationadmin.isGranted('EDIT')
                     and btn_edit
                     and sonata_admin.value != null


greg0ire:
@kunicmarko20 please review this

kunicmarko20:
@greg0ire this is something I mentioned 2 times ([here](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/708#issuecomment-322562327), [here](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/708#issuecomment-345464166)) in the PR as I knew something like this would happen because the edit is replacing add button always (I mean that is not bad but you can not disable it).

kunicmarko20:
Hi @n3o77,

The #708 introduced the edit button that now replaces the add button if an object is already set. The confusing part about add button is that if you already have an object it will replace the existing one but the label is still `"Add"`. From this PR I can see that this is something you need so we can move that check to the label instead. We would then replace `{{ btn_add|trans({}, btn_catalogue) }}` with `{{ (sonata_admin.value == null ? btn_add :  btn_replace) |trans({}, btn_catalogue) }}` and we can have both the add and the edit button but instead of "Add" we can have "Replace" (maybe "Replace with New") label, we only need to add a new PR to `SonataAdminBundle` that would add this string to translations.

n3o77:
Hi @kunicmarko20,

i guess it should only replace the linked entity if it has the delete cascade and|or orphanRemoval. I have the case that the linked entity will stay in the database and just want to link a new one. Or to be more precise, i want to create a new one and link to that, but unlinking doesn't mean the entity is going to be deleted.

Not sure if i get stuck on semantics here, just want to make sure we're on the same page.

But i guess i understand your point. The button label can be confusing in some scenarios. In my scenario it is more `Add new entity to the DB and link to it`.  But this can already be taken care of by the developer, either he can use the translations to rename it everywhere, or can use the `btn_add` option in the formBuilder to set the label dynamically.

The other big problem I noticed with the changes in #708 is that if you select a new entity the `edit` button href isn't changed in the JS events. So you link to a new one (with the `list` button i.e.) but if you click the `edit` button you will always edit the initially linked entity, same will be true if we have different labels for the button in different states.

IMHO not having both buttons linked in any way and keep them as separated as possible is the best solution. We already have all the options in the formBuilder to cover all of the scenarios and gives the dev the most options and control over the behavior.

kunicmarko20:
I totally agree with @n3o77 that label can be handled in formBuilder and that this should be separated, so this can be merged and it will be on the user to choose if he wants to have add or edit button and what label should be shown.

The other big problem with edit not refreshing I was able to reproduce so I will make a new PR about that.

PR #767, opened 2017-12-02
Manual merge

greg0ire:


src/Builder/FormContractor.php
```diff
@@ -109,7 +107,14 @@ public function getDefaultOptions($type, FieldDescriptionInterface $fieldDescrip
     {
         $options = [];
         $options['sonata_field_description'] = $fieldDescription;
-        if ($this->checkFormClass($type, [
+
+        // NEXT_MAJOR: Check only against FQCNs when dropping support for form mapping
```
greg0ire: I fixed these misleading comments on the way and restored form mapping compat

PR #770, opened 2017-12-04
Refresh edit button URL if item is replaced with list/add action

kunicmarko20:
I am targeting this branch, because I am adding fix for edit button that was always opening initial item in popup even if you replace that item with list/add action. 

## Changelog

```markdown
### Added
- Added refresh of edit button URL if item was replaced by add/list action

### Fixed
- Issue with edit button always showing initial item in popup
```

## To do

- [x] Update the tests

## Subject

As pointed out to me in https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/765#issuecomment-348928269 by n3o77 if you replace the item with list/add action edit button will still open the initial item in the popup and that is wrong. I added javascript code that is loaded if `btn_edit` is enabled, and it will change the URL of edit button when updating the label.

### DIFF
--- a/src/Resources/views/CRUD/edit_orm_many_association_script.html.twig
+++ b/src/Resources/views/CRUD/edit_orm_many_association_script.html.twig
@@ -473,6 +473,14 @@ This code manage the many-to-[one|many] association field popup
                     jQuery('#field_widget_{{ id }}').html(html);
                 }
             });
+
+            {% if btn_edit %}
+                var edit_button_url = '{{
+                    associationadmin.generateUrl('edit', {'id' : 'OBJECT_ID'})
+                }}'.replace('OBJECT_ID', jQuery(this).val());
+
+                jQuery('#field_actions_{{ id }} a.btn-warning').attr('href', edit_button_url);
+            {% endif %}
         });
 
     {% endif %}


src/Resources/views/CRUD/edit_orm_many_association_script.html.twig
```diff
@@ -473,6 +473,11 @@ This code manage the many-to-[one|many] association field popup
                     jQuery('#field_widget_{{ id }}').html(html);
                 }
             });
+
+            {% if btn_edit %}
+                var edit_button_url = '{{ sonata_admin.field_description.associationadmin.generateUrl('edit', {'id' : 'OBJECT_ID'}) }}'.replace('OBJECT_ID', jQuery(this).val());
```
greg0ire: Hard to review from phone, please linebreak

kunicmarko20: @greg0ire my bad, fixed now.

ElectricMaxxx:
@greg0ire didn't we move the association files to SonataAdminBundles? At the end they have to be fixed there also: https://github.com/sonata-project/SonataAdminBundle/tree/3.x/src/Resources/views/CRUD/Association

greg0ire:
Uuuuuh if we did, the current file should be deprecated...

greg0ire:
Oh @ElectricMaxxx you did this in https://github.com/sonata-project/SonataAdminBundle/pull/4589, but there should be follow-up PRs to deprecate the templates in the persistence bundles, right?

ElectricMaxxx:
no clue what i forgot there :-)

ElectricMaxxx:
but at the end, we whanted to have them together in main bundle and remove them from persistence bundles like this one here.

kunicmarko20:
@greg0ire so what now? as we are still using `SonataDoctrineORMAdminBundle` templates and I added the edit button to this bundle templates only. Should I make a PR and move the templates I edited to `SonataAdminBundle` also?

ElectricMaxxx:
I would say yes, and we should deprecate the templates here.

greg0ire:
@n3o77 please review

greg0ire:
Thanks @kunicmarko20 !

PR #780, opened 2017-12-13
Replace FQCN strings with ::class constants

ZRazor:
I am targeting this branch, because it's BC.

## Changelog

```markdown
### Fixed
 - Replaced FQCN strings with `::class` constants
```


greg0ire:
Thanks @ZRazor !

PR #783, opened 2017-12-13
WIP 7.1 signatures

greg0ire:
I am targeting this branch, because this is a huge BC-break.

## Changelog

```markdown
### Changed
- most method signatures now have php 7.1 type hints
```

## To do

- [ ] Change signatures in interfaces that are in the admin bundle :/
- [ ] Change more signatures
- [ ] Add an upgrade note

### DIFF
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -56,12 +56,7 @@ class ModelManager implements ModelManagerInterface, LockInterface
         $this->registry = $registry;
     }
 
-    /**
-     * @param string $class
-     *
-     * @return ClassMetadata
-     */
-    public function getMetadata($class)
+    public function getMetadata(string $class): ClassMetadata
     {
         return $this->getEntityManager($class)->getMetadataFactory()->getMetadataFor($class);
     }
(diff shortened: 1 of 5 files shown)


core23:
Ping @greg0ire

src/Model/ModelManager.php
```diff
@@ -363,11 +357,14 @@ public function getIdentifierValues($entity)
...
     }
 
-    public function getIdentifierFieldNames($class)
+    public function getIdentifierFieldNames($class): array
     {
         return $this->getMetadata($class)->getIdentifierFieldNames();
     }
 
+    /**
+     * @return string|void
```
core23: string or **void**? This looks wrong

src/Model/ModelManager.php
```diff
@@ -399,6 +396,8 @@ public function getNormalizedIdentifier($entity)
      *
      * The ORM implementation does nothing special but you still should use
      * this method when using the id in a URL to allow for future improvements.
+     *
+     * @return string|void
```
core23: Same here

greg0ire:
> Ping @greg0ire

no need to work on this until the first element of the todo list is resolved I'm afraid :sweat_smile:

core23:
I was just looking at the creation date of this PR:  Dec. 2017 oO

greg0ire:
Yeah me too, I was a bit suprised, so I started working on it, and then I saw that there were many cases where I just could not add type declarations because we rely on interfaces in the admin bundle that don't have type declarations yet.

PR #792, opened 2018-01-14
Manual merge

greg0ire:
Just approve, don't merge.

### DIFF
--- a/src/Builder/FormContractor.php
+++ b/src/Builder/FormContractor.php
@@ -160,7 +160,7 @@ class FormContractor implements FormContractorInterface
 
             $options['data_class'] = $fieldDescription->getAssociationAdmin()->getClass();
             $fieldDescription->setOption('edit', $fieldDescription->getOption('edit', 'admin'));
-            // NEXT_MAJOR: Check only against FQCNs when dropping support for form mapping
+        // NEXT_MAJOR: Check only against FQCNs when dropping support for Symfony 2.8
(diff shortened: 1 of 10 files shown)


greg0ire:
The build is broken because of doctrine type hints, and I think this should be fixed on the stable branch

jordisala1991:
Type hints or because of the strict type declaration?

greg0ire:
unsure yet, but doctrine 2.6 is out, and I think it might have `string` type hints.

greg0ire:
Extracting the interesting part: `/home/travis/build/sonata-project/SonataDoctrineORMAdminBundle/vendor/sonata-project/admin-bundle/src/Admin/BaseFieldDescription.php:265`

greg0ire:
Leads to this: https://github.com/sonata-project/SonataAdminBundle/blob/3.x/src/Admin/BaseFieldDescription.php#L265

greg0ire:
might be fixed in this bundle since its a passthrough

jordisala1991:
Ohh yes.. I see better now. We have to fix it on stable indeed, add a cast should fix it right?

greg0ire:
Not sure yet, I think it might just be the tests that need fixing

greg0ire:
Updating because I don't reproduce it locally, I must have an old version of doctrine

greg0ire:
Ok now I reproduce the issue

greg0ire:
See #793

PR #805, opened 2018-02-03
Hide selects added by OrderByToSelectWalker from hydration

dmarkowicz:
I am targeting this branch, because this is BC fix.

## Changelog

```markdown
### Fixed
- Hide selects added by OrderByToSelectWalker from hydration
```

## Subject
Select expressions added by `OrderByToSelectWalker` should be marked as hidden to prevent pollute query result field set. This class works as expected in ProxyQuery but it can be use in other contexts like https://github.com/sonata-project/SonataAdminBundle/issues/4924.


jordisala1991:
Does this fix the second issue @kunicmarko20 ?

kunicmarko20:
yes, this fixes the problem

greg0ire:
Thanks @dmarkowicz !

PR #807, opened 2018-02-04
Add missing tests

kunicmarko20:

## Subject

Adding tests that are missing, as much as I can.

### DIFF
--- a/tests/Block/AuditBlockServiceTest.php
+++ b/tests/Block/AuditBlockServiceTest.php
@@ -1,0 +1,84 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DoctrineORMAdminBundle\Tests\Block;
+
+use Prophecy\Argument;
+use SimpleThings\EntityAudit\AuditReader as SimpleThingsAuditReader;
+use SimpleThings\EntityAudit\Revision;
+use Sonata\BlockBundle\Block\BlockContext;
+use Sonata\BlockBundle\Model\Block;
+use Sonata\BlockBundle\Test\AbstractBlockServiceTestCase;
+use Sonata\DoctrineORMAdminBundle\Block\AuditBlockService;
+
+/**
+ * @author Marko Kunic <kunicmarko20@gmail.com>
+ */
+class AuditBlockServiceTest extends AbstractBlockServiceTestCase
+{
+    private $simpleThingsAuditReader;
+    private $blockService;
+
+    protected function setUp()
+    {
+        parent::setUp();
+        $this->simpleThingsAuditReader = $this->prophesize(SimpleThingsAuditReader::class);
+
+        $this->blockService = new AuditBlockService(
+            'block.service',
+            $this->templating,
+            $this->simpleThingsAuditReader->reveal()
+        );
+    }
+
+    /**
+     * @group legacy
+     */
+    public function testExecute()
+    {
+        $blockContext = $this->prophesize(BlockContext::class);
+
+        $blockContext->getBlock()->willReturn($block = new Block())->shouldBeCalledTimes(1);
+        $blockContext->getSetting('limit')->willReturn($limit = 10)->shouldBeCalledTimes(1);
+
+        $this->simpleThingsAuditReader->findRevisionHistory($limit, 0)
+            ->willReturn([$revision = new Revision('test', '123', 'test')])
+            ->shouldBeCalledTimes(1);
+
+        $this->simpleThingsAuditReader->findEntitesChangedAtRevision(Argument::cetera())
+            ->willReturn([])
+            ->shouldBeCalledTimes(1);
+
+        $blockContext->getTemplate()->willReturn('template')->shouldBeCalledTimes(1);
+        $blockContext->getSettings()->willReturn([])->shouldBeCalledTimes(1);
+
+        $this->blockService->execute($blockContext->reveal());
+
+        $this->assertSame('template', $this->templating->view);
+        $this->assertInternalType('array', $this->templating->parameters['settings']);
+        $this->assertSame($revision, $this->templating->parameters['revisions'][0]['revision']);
+        $this->assertSame($block, $this->templating->parameters['block']);
+    }
+
+    public function testDefaultSettings()
+    {
+        $blockContext = $this->getBlockContext($this->blockService);
+
+        $this->assertSettings([
+            'attr' => [],
+            'extra_cache_keys' => [],
+            'limit' => 10,
+            'template' => '@SonataDoctrineORMAdmin/Block/block_audit.html.twig',
+            'ttl' => 0,
+            'use_cache' => true,
+        ], $blockContext);
+    }
+}
--- a/tests/Builder/DatagridBuilderTest.php
+++ b/tests/Builder/DatagridBuilderTest.php
@@ -11,69 +11,159 @@
 
 namespace Sonata\DoctrineORMAdminBundle\Tests\Builder;
 
+use Doctrine\ORM\Mapping\ClassMetadata;
 use PHPUnit\Framework\TestCase;
+use Prophecy\Argument;
 use Sonata\AdminBundle\Admin\AbstractAdmin;
 use Sonata\AdminBundle\Admin\FieldDescriptionCollection;
 use Sonata\AdminBundle\Datagrid\Datagrid;
+use Sonata\AdminBundle\Datagrid\DatagridInterface;
 use Sonata\AdminBundle\Datagrid\Pager;
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
+use Sonata\AdminBundle\Datagrid\SimplePager;
 use Sonata\AdminBundle\Filter\FilterFactoryInterface;
 use Sonata\AdminBundle\Guesser\TypeGuesserInterface;
-use Sonata\AdminBundle\Model\ModelManagerInterface;
+use Sonata\AdminBundle\Translator\FormLabelTranslatorStrategy;
+use Sonata\DoctrineORMAdminBundle\Admin\FieldDescription;
 use Sonata\DoctrineORMAdminBundle\Builder\DatagridBuilder;
+use Sonata\DoctrineORMAdminBundle\Filter\ModelAutocompleteFilter;
+use Sonata\DoctrineORMAdminBundle\Model\ModelManager;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\Form\FormFactoryInterface;
+use Symfony\Component\Form\Guess\TypeGuess;
 
 /**
  * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ * @author Marko Kunic <kunicmarko20@gmail.com>
  */
 final class DatagridBuilderTest extends TestCase
 {
     /**
-     * @var TypeGuesserInterface|\PHPUnit_Framework_MockObject_MockObject
-     */
-    protected $typeGuesser;
-
-    /**
      * @var DatagridBuilder
      */
-    protected $datagridBuilder;
-    /**
-     * @var FormFactoryInterface|\PHPUnit_Framework_MockObject_MockObject
-     */
+    private $datagridBuilder;
+    private $typeGuesser;
     private $formFactory;
-
-    /**
-     * @var FilterFactoryInterface|\PHPUnit_Framework_MockObject_MockObject
-     */
     private $filterFactory;
+    private $admin;
+    private $modelManager;
 
     protected function setUp()
     {
-        $this->formFactory = $this->createMock(FormFactoryInterface::class);
-        $this->filterFactory = $this->createMock(FilterFactoryInterface::class);
-        $this->typeGuesser = $this->createMock(TypeGuesserInterface::class);
+        $this->formFactory = $this->prophesize(FormFactoryInterface::class);
+        $this->filterFactory = $this->prophesize(FilterFactoryInterface::class);
+        $this->typeGuesser = $this->prophesize(TypeGuesserInterface::class);
 
-        $this->datagridBuilder = new DatagridBuilder($this->formFactory, $this->filterFactory, $this->typeGuesser);
+        $this->datagridBuilder = new DatagridBuilder(
+            $this->formFactory->reveal(),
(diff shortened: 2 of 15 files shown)


kunicmarko20:
Should probably move to prophesize (https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/807/files#diff-6dfd8ead8e135a17ae2f1cec987ee09cR55) :sweat_smile:

greg0ire:
This is great, but pedantic since there is zero benefits when upgrading. The benefit come when developing :) Removed the changelog. prophesize is allowed and recommended for new tests, so if you want to change, you can.

jordisala1991:
Yes, if you can move to prophecy that would be great

tests/Block/AuditBlockServiceTest.php
```diff
@@ -0,0 +1,100 @@
...
+     * @group legacy
+     */
+    public function testExecute()
+    {
+        $block = $this->prophesize(Block::class);
+        $revision = $this->prophesize(Revision::class);
+        $blockContext = $this->prophesize(BlockContext::class);
+
+        $blockContext->getBlock()
+            ->willReturn($block->reveal())
```
jordisala1991: this can go one lined and won't be more than 100 chars

tests/SonataDoctrineORMAdminBundleTest.php
```diff
@@ -25,17 +26,21 @@ class SonataDoctrineORMAdminBundleTest extends TestCase
...
-        $containerBuilder->expects($this->exactly(3))
-            ->method('addCompilerPass')
-            ->withConsecutive(
-                [$this->isInstanceOf(AddGuesserCompilerPass::class)],
-                [$this->isInstanceOf(AddTemplatesCompilerPass::class)],
-                [$this->isInstanceOf(AddAuditEntityCompilerPass::class)]
-            );
+        $containerBuilder
+            ->addCompilerPass(Argument::type(AddGuesserCompilerPass::class))
+            ->shouldBeCalledTimes(1);
```
jordisala1991: All this can be simplified to shouldBeCalled().
And could be done in one line probably (have to check chars)

kunicmarko20: Well, I can do it in one line and check for `CompilerPassInterface` instead of specific class.

jordisala1991: That's another option. There are certain things that can be done with phpunit mocks that are not direct translated to how you do it with prophecy, but there is always a way in which you get a cleaner code. atleast IMO.

tests/Block/AuditBlockServiceTest.php
```diff
@@ -0,0 +1,86 @@
...
+ */
+class AuditBlockServiceTest extends AbstractBlockServiceTestCase
+{
+    private $simpleThingsAuditReader;
+    private $blockService;
+
+    protected function setUp()
+    {
+        parent::setUp();
+        $this->simpleThingsAuditReader = $this->prophesize(SimpleThingsAuditReader::class);
```
greg0ire: Can you avoid mocking this and use the real thing instead or is it too cumbersome?

kunicmarko20: Then I would need to mock things inside of `SimpleThingsAuditReader`?

greg0ire: I don't know what dependencies it has, that's why I'm asking. If there are too many, just don't do it.

kunicmarko20:
```     */
    public function __construct(EntityManagerInterface $em, AuditConfiguration $config, MetadataFactory $factory)
    {
        $this->em = $em;
        $this->config = $config;
        $this->metadataFactory = $factory;
        $this->platform = $this->em->getConnection()->getDatabasePlatform();
        $this->quoteStrategy = $this->em->getConfiguration()->getQuoteStrategy();
    }
```

Too many :sweat_smile: , wouldn't that basically be testing 3rd party code?

greg0ire: Yeah drop it for that particular case, it's way too complicated. The goal is to avoid mocking internal calls, catch `trigger_error` and deprecations if any

tests/Builder/ShowBuilderTest.php
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    public function testAddFieldNoType()
+    {
+        $fieldDescriptionCollection = $this->prophesize(FieldDescriptionCollection::class);
+        $fieldDescription = $this->prophesize(FieldDescriptionInterface::class);
+        $admin = $this->prophesize(AdminInterface::class);
+        $typeGuess = $this->prophesize(TypeGuess::class);
+        $modelManager = $this->prophesize(ModelManager::class);
+        $typeGuess = $this->prophesize(TypeGuess::class);
```
jordisala1991: duplicated line with 60

kunicmarko20: nice catch!

tests/Block/AuditBlockServiceTest.php
```diff
@@ -0,0 +1,86 @@
...
+        );
+    }
+
+    /**
+     * @group legacy
+     */
+    public function testExecute()
+    {
+        $block = $this->prophesize(Block::class);
+        $revision = $this->prophesize(Revision::class);
```
greg0ire: This one you can easily avoid mocking

greg0ire: Both, in fact.

kunicmarko20: even if  I assert them down?

tests/Builder/ShowBuilderTest.php
jordisala1991: I am looking at what scrutinizer says, there are a few errors out there, can you check it?

kunicmarko20: errors on `AuditReaderTest` do not make sense?

tests/Guesser/FilterTypeGuesserTest.php
```diff
@@ -43,4 +50,162 @@ public function testThrowsOnMissingField()
...
+            ],
+            [
+                'smallint',
+                $numberType,
+                Guess::MEDIUM_CONFIDENCE,
+                NumberType::class
+            ],
+            [
+                'string',
+                $strngType = 'doctrine_orm_string',
```
OskarStark:
```diff
- $stringType
+ $stringType
```

tests/Guesser/FilterTypeGuesserTest.php
```diff
@@ -43,4 +50,162 @@ public function testThrowsOnMissingField()
...
+                Guess::MEDIUM_CONFIDENCE,
+                TextType::class
+            ],
+            [
+                'text',
+                $strngType,
```
OskarStark:
```diff
- $stringType
+ $stringType
```

tests/Guesser/FilterTypeGuesserTest.php
```diff
@@ -43,4 +50,162 @@ public function testThrowsOnMissingField()
...
+                TextType::class
+            ],
+            [
+                'time',
+                'doctrine_orm_time',
+                Guess::HIGH_CONFIDENCE,
+            ],
+            [
+                'somefake',
+                $strngType,
```
OskarStark:
```diff
- $stringType
+ $stringType
```

tests/Builder/ShowBuilderTest.php
jordisala1991: It is because of docblocks declaring types

tests/Builder/ShowBuilderTest.php
```diff
@@ -0,0 +1,198 @@
...
+use Sonata\DoctrineORMAdminBundle\Model\ModelManager;
+use Symfony\Component\Form\Guess\TypeGuess;
+
+/**
+ * @author Marko Kunic <kunicmarko20@gmail.com>
+ */
+class ShowBuilderTest extends TestCase
+{
+    /**
+     * @var TypeGuesserInterface
```
jordisala1991: Adding this makes scrutinizer think this is a TypeGuesserInterface, but it is a ObjectProphecy (or something like that). IMO I would remove all those comments from test classes, because they do not give a lot of information, and just make noise

tests/Guesser/FilterTypeGuesserTest.php
kunicmarko20: thank you, fixed.

tests/Builder/ShowBuilderTest.php
kunicmarko20: my bad, fixed.

tests/Block/AuditBlockServiceTest.php
greg0ire: For `$revision` it will simplify things, and for `$block`, you did not write any assertions. The only thing lost is the number of times `getRev` is called, but that's an implementation detail, and we only care about the outcome of the test, right?

tests/Builder/ShowBuilderTest.php
```diff
@@ -175,13 +178,35 @@ public function testFixFieldDescription()
...
-        $fieldDescription->setTemplate('@SonataAdmin/CRUD/Association/show_one_to_one.html.twig')
+        $fieldDescription->setTemplate($template)
             ->shouldBeCalledTimes(1);
 
         $admin->attachAdminClass($fieldDescription)->shouldBeCalledTimes(1);
 
         $this->showBuilder->fixFieldDescription($admin->reveal(), $fieldDescription->reveal());
     }
+
+    public function testFixFieldDescriptionData()
```
greg0ire: A data provide that starts with `test`? Is phpunit ok with this?

tests/Block/AuditBlockServiceTest.php
```diff
@@ -43,19 +43,12 @@ protected function setUp()
      */
     public function testExecute()
     {
-        $block = $this->prophesize(Block::class);
-        $revision = $this->prophesize(Revision::class);
```
greg0ire: I'd thought you were going to do `$revision = new Revision()`, why did you remove it?

tests/Builder/ShowBuilderTest.php
kunicmarko20: I did not have any errors, so it should be ok but I will rename it as this was a mistake.

tests/Block/AuditBlockServiceTest.php
kunicmarko20: My bad, I understood `avoid mocking` as removing, will be fixed.

greg0ire:
> +17.6%

That's impressive!

kunicmarko20:
My goal is at least `70%` coverage, probably today or tomorrow 😄

kunicmarko20:
I guess it is done if you have time review it. Should I squash commits? also, any idea what to do about the test that is failing on prefer lowest?

dmarkowicz:
Lowest version passing test is:
`"simplethings/entity-audit-bundle": "0.5 - 0.9 || ^1.0",`
What about simplify it to:
`"simplethings/entity-audit-bundle": "^0.9 || ^1.0",`

tests/Builder/DatagridBuilderTest.php
```diff
@@ -11,69 +11,159 @@
...
+        return [
+            [
+                Pager::TYPE_SIMPLE,
+                SimplePager::class,
+            ],
+            [
+                Pager::TYPE_DEFAULT,
+                Pager::class,
+            ],
+        ];
```
greg0ire: It's good practice to specify a description for each item of a data provider as its key
```
return [
   'simple' => [
…
]
]
```

kunicmarko20: will add keys, any idea about the issue with the test, what @dmarkowicz suggested make sense but would that be a BC break if we move to 0.9?

jordisala1991: Increase dependencies is not a BC break, IMO let's do it

tests/Filter/DateTimeFilterTest.php
```diff
@@ -0,0 +1,43 @@
...
+ * @author Marko Kunic <kunicmarko20@gmail.com>
+ */
+class DateTimeFilterTest extends TestCase
+{
+    public function testEmpty()
+    {
+        $filter = new DateTimeFilter();
+        $filter->initialize('field_name', ['field_options' => ['class' => 'FooBar']]);
+
+        $builder = new ProxyQuery(new QueryBuilder());
```
jordisala1991: This queryBuilder should have a `use` right?

tests/Filter/DateTimeRangeFilterTest.php
```diff
@@ -0,0 +1,42 @@
...
+ * @author Marko Kunic <kunicmarko20@gmail.com>
+ */
+class DateTimeRangeFilterTest extends TestCase
+{
+    public function testEmpty()
+    {
+        $filter = new DateTimeRangeFilter();
+        $filter->initialize('field_name', ['field_options' => ['class' => 'FooBar']]);
+
+        $builder = new ProxyQuery(new QueryBuilder());
```
jordisala1991: same here

tests/Filter/TimeFilterTest.php
```diff
@@ -0,0 +1,43 @@
...
+ * @author Marko Kunic <kunicmarko20@gmail.com>
+ */
+class TimeFilterTest extends TestCase
+{
+    public function testEmpty()
+    {
+        $filter = new TimeFilter();
+        $filter->initialize('field_name', ['field_options' => ['class' => 'FooBar']]);
+
+        $builder = new ProxyQuery(new QueryBuilder());
```
jordisala1991: and here

tests/Filter/DateTimeFilterTest.php
kunicmarko20: There is a QueryBuilder class i that folder https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/tests/Filter/QueryBuilder.php

kunicmarko20:
Test are finally passing, could you review this and lets merge it :smile:

jordisala1991:
Great work BTW

kunicmarko20:
Thank you, I am happy with +26%, we are over 70% on this bundle :tada:

kunicmarko20:
Also started to like prophecy a lot, I think I can't (and don't want to) go back to `createMock` way :smile:

jordisala1991:
Once you try, you never go back ;)

Those tests could probably be ported to MongoDBAdminBundle and PHPCRAdminBundle without much trouble. We should probably give it a try some day :)

kunicmarko20:
You are right, some of the tests could be ported, will look into it.

jordisala1991:
Thank you @kunicmarko20

PR #810, opened 2018-02-10
Increase dependency of simplethings audit bundle

kunicmarko20:
I am targeting this branch, because this is a patch.

## Subject

Currently, new tests that I added https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/807 are failing because the minimum version is `0.1`, raising the minimum version of `SimpleThingsEntityAuditBundle` fixes the problem.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
     },
     "require-dev": {
         "knplabs/knp-menu-bundle": "^2.1.1",
-        "simplethings/entity-audit-bundle": "0.1 - 0.9 || ^1.0",
+        "simplethings/entity-audit-bundle": "^0.9 || ^1.0",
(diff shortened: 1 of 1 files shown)


jordisala1991:
This will be pedantic I think

kunicmarko20:
Should I remove the changelog then?

jordisala1991:
I did comment your changelog since it is pedantic (not removed, because if I am wrong, we can recover it)

jordisala1991:
One question, this change makes your tests pass?

kunicmarko20:
Yes, it seems that earlier version do not even have that `diff` method that we use in our Reader, that is why it is failing on prefer lowest

jordisala1991:
Thank you @kunicmarko20

PR #811, opened 2018-02-10
Make composer dependencies explicit

jordisala1991:

## Subject

Just declared some dependencies that are already on the code, and removed dependency on security, because the only reference to security is ACL. It was added here: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/320 just to replace the main symfony package.


jordisala1991:
I did update phpunit-bridge while I am changing some composer things :P. Please review again

greg0ire:
Thanks @jordisala1991 !

PR #815, opened 2018-02-21
Pager: reset also groupBy dql part

egircys:
I am targeting this branch, because it affects also existing functionality.

## Changelog
Pager: reset also groupBy dql part

## Subject
When in the Admin class, the `createQuery()` method has a `->groupBy` specified,
then the admin filter functionality with an empty result set fails.

```php
    {
        /** @var \Sonata\AdminBundle\Datagrid\ProxyQueryInterface|QueryBuilder $query */

        $alias = current($query->getRootAliases());

        $query->select("NEW NS\AdminBundle\Dto\InterfaceDto($alias.id, $alias.name, COUNT(pim.config))")
            ->leftJoin("{$alias}.interface", 'pi')
            ->leftJoin(Configuration::class, 'pic', Join::WITH, 'pi.id = pic.interface')
            ->innerJoin(InterfaceMapping::class, 'pim', Join::WITH, 'pim.configuration = pic.id')
            ->setMaxResults(1)
            ->groupBy("$alias.id")
        ;

        return $query;
    }
```

### DIFF
--- a/src/Datagrid/Pager.php
+++ b/src/Datagrid/Pager.php
@@ -42,7 +42,7 @@ class Pager extends BasePager
             current($this->getCountColumn())
         ));
 
-        return $countQuery->resetDQLPart('orderBy')->getQuery()->getSingleScalarResult();
+        return $countQuery->resetDQLPart('orderBy')->resetDQLPart('groupBy')->getQuery()->getSingleScalarResult();
     }
 
     public function getResults($hydrationMode = Query::HYDRATE_OBJECT)


src/Datagrid/Pager.php
```diff
@@ -42,7 +42,7 @@ public function computeNbResult()
             current($this->getCountColumn())
         ));
 
-        return $countQuery->resetDQLPart('orderBy')->getQuery()->getSingleScalarResult();
+        return $countQuery->resetDQLPart('orderBy')->resetDQLPart('groupBy')->getQuery()->getSingleScalarResult();
```
greg0ire: Ok so let's admit this provides a better result in case of empty set. What about other cases? Doesn't it provide a wrong result because there's no grouping and you are counting rows with several relationships more than once?

egircys: Thank you for reviewing @greg0ire !

Actually, without removing the groupBy part it provides wrong results (count always 1). I guess, because it's overriding the select part `->select("NEW NS\AdminBundle..`

With my current data set, the counting works well with empty, filtered or not filtered data.

greg0ire: are you sure? what if you have 5 configurations for one interface?

egircys: pretty sure, it's grouping by the interface id.. But I see your point, that it might not work for all cases. I'll will check for another solution. Thank you!

PR #822, opened 2018-03-11
Entities with construct can be used on Collection and AdminType

jordisala1991:

Closes https://github.com/sonata-project/SonataAdminBundle/issues/5001

## Changelog

```markdown
### Fixed
- Now it is possible to use entities with arguments on the constructor on the Collection and Admin types.
```

## To do

- [x] Update the tests

## Subject

`empty_data` can be used to override the default behavior when an entity does have a constructor with arguments.

http://symfony.com/doc/master/form/use_empty_data.html#option-1-instantiate-a-new-class

We are using the closure because otherwise it will fail on the collectionType. Probably adminType can use it without closure, have to check.

@Soullivaneuh can you check if this fixes your issue?

This is WIP, because we have to check we don't break anything with it.

### DIFF
--- a/src/Builder/FormContractor.php
+++ b/src/Builder/FormContractor.php
@@ -157,6 +157,9 @@ class FormContractor implements FormContractorInterface
             $options['delete'] = false;
 
             $options['data_class'] = $fieldDescription->getAssociationAdmin()->getClass();
+            $options['empty_data'] = function () use ($fieldDescription) {
+                return $fieldDescription->getAssociationAdmin()->getNewInstance();
+            };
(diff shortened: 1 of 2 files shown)


jordisala1991:
I did some testing and it works for me, but I would like to hear @Soullivaneuh opinion on this.

So, ping.

jordisala1991:
I will try to add a test to see if the 'empty_data' resolves to a new instance of the admin, more than that don't know how to do it.

jordisala1991:
Tests added, please review again

OskarStark:
@Soullivaneuh please review again and hit the merge button 😄

jordisala1991:
Ping @Soullivaneuh

PR #827, opened 2018-04-13
 #473. Available disable distinct for ProxyQuery and Pager.

### LINKED ISSUE #473: List view for big tables very slow due to DISTINCT query

eugenekurasov:
List view for big tables very slow due to DISTINCT query.
Also we should have available to enable distinct for query like with left join.

I am targeting this branch, because this is Feature.

Closes #473

## Changelog

```markdown
### Added
- Added `ProxyQuery::setDistinct` and `ProxyQuery::isDistinct`.

### Changed
- `Pager` use `CountWalker` for get count.
```

## To do

- [x] Update the tests

### DIFF
--- a/docs/reference/query_proxy.rst
+++ b/docs/reference/query_proxy.rst
@@ -9,7 +9,15 @@ The ``ProxyQuery`` object is used to add missing features from the original `Doc
 
 * ``execute`` method - no need to call the ``getQuery()`` method,
 * add sort by and sort order options,
-* add preselect id query on left join query, so a limit query will be only applied on the left statement and not on the full select statement. This simulates the original Doctrine 1 behavior.
+* add preselect id query on left join query, so a limit query will be only
+  applied on the left statement and not on the full select statement.
+  This simulates the original Doctrine 1 behavior.
+* By default, Sonata will use the ``DISTINCT`` SQL keyword when fetching
+  the identifiers of the entities that will be displayed in the listing,
+  to avoid duplicates in some cases. Sonata cannot detect whether or not
+  you need ``DISTINCT``, but lets you remove that keyword in case it
+  causes performance issues and you are sure there will be no duplicates.
+  To do so, simply call ``setDistinct(false)``.
 
 
 .. code-block:: php
--- a/src/Datagrid/Pager.php
+++ b/src/Datagrid/Pager.php
@@ -11,7 +11,9 @@
 
 namespace Sonata\DoctrineORMAdminBundle\Datagrid;
 
+use Doctrine\ORM\AbstractQuery;
 use Doctrine\ORM\Query;
+use Doctrine\ORM\Tools\Pagination\CountWalker;
 use Sonata\AdminBundle\Datagrid\Pager as BasePager;
 
 /**
@@ -36,13 +38,14 @@ class Pager extends BasePager
             $countQuery->setParameters($this->getParameters());
         }
 
-        $countQuery->select(sprintf(
-            'count(DISTINCT %s.%s) as cnt',
-            current($countQuery->getRootAliases()),
-            current($this->getCountColumn())
-        ));
+        $query = $countQuery->getQuery();
+        $query->setHint(
+            CountWalker::HINT_DISTINCT,
+            $countQuery instanceof ProxyQuery ? $countQuery->isDistinct() : true
+        );
+        $this->appendTreeWalker($query, CountWalker::class);
 
-        return $countQuery->resetDQLPart('orderBy')->getQuery()->getSingleScalarResult();
+        return $query->getSingleScalarResult();
     }
 
     public function getResults($hydrationMode = Query::HYDRATE_OBJECT)
@@ -78,5 +81,22 @@ class Pager extends BasePager
             $this->getQuery()->setFirstResult($offset);
             $this->getQuery()->setMaxResults($this->getMaxPerPage());
         }
+    }
+
+    /**
+     * Appends a custom tree walker to the tree walkers hint.
+     *
+     * @param string $walkerClass
+     */
+    private function appendTreeWalker(AbstractQuery $query, $walkerClass)
+    {
+        $hints = $query->getHint(Query::HINT_CUSTOM_TREE_WALKERS);
+
+        if (false === $hints) {
+            $hints = [];
+        }
+
+        $hints[] = $walkerClass;
+        $query->setHint(Query::HINT_CUSTOM_TREE_WALKERS, $hints);
     }
 }
--- a/src/Datagrid/ProxyQuery.php
+++ b/src/Datagrid/ProxyQuery.php
@@ -48,6 +48,13 @@ class ProxyQuery implements ProxyQueryInterface
     protected $entityJoinAliases;
 
     /**
+     * For BC reasons, this property is true by default.
+     *
+     * @var bool
+     */
+    private $distinct = true;
+
+    /**
      * The map of query hints.
      *
      * @var array<string,mixed>
@@ -77,6 +84,33 @@ class ProxyQuery implements ProxyQueryInterface
     public function __clone()
     {
         $this->queryBuilder = clone $this->queryBuilder;
+    }
+
+    /**
+     * Optimize queries with a lot of rows.
+     * It is not recommended to use "false" with left joins.
+     *
+     * @param bool $distinct
+     *
+     * @return self
+     */
+    final public function setDistinct($distinct)
+    {
+        if (!is_bool($distinct)) {
+            throw new \InvalidArgumentException('$distinct is not a boolean');
+        }
+
+        $this->distinct = $distinct;
+
+        return $this;
+    }
+
+    /**
+     * @return bool
+     */
+    final public function isDistinct()
+    {
+        return $this->distinct;
     }
 
     public function execute(array $params = [], $hydrationMode = null)
@@ -300,7 +334,7 @@ class ProxyQuery implements ProxyQueryInterface
             $idxSelect .= ('' !== $idxSelect ? ', ' : '').$idSelect;
         }
         $queryBuilderId->select($idxSelect);
-        $queryBuilderId->distinct();
+        $queryBuilderId->distinct($this->isDistinct());
 
         // for SELECT DISTINCT, ORDER BY expressions must appear in idxSelect list
         /* Consider
--- a/tests/Datagrid/ProxyQueryTest.php
+++ b/tests/Datagrid/ProxyQueryTest.php
@@ -72,21 +72,17 @@ class ProxyQueryTest extends TestCase
     public function dataGetFixedQueryBuilder()
     {
         return [
-            ['aaa', 'bbb', 'id', 'id_idx', 33, Type::INTEGER],
-            ['aaa', 'bbb', 'associatedId', 'associatedId_idx', 33, null],
-            ['aaa', 'bbb', 'id.value', 'id_value_idx', 33, Type::INTEGER],
-            ['aaa', 'bbb', 'id.uuid', 'id_uuid_idx', new NonIntegerIdentifierTestClass('80fb6f91-bba1-4d35-b3d4-e06b24494e85'), UuidType::NAME],
+            ['aaa', 'bbb', 'id', 'id_idx', 33, Type::INTEGER, true],
+            ['aaa', 'bbb', 'associatedId', 'associatedId_idx', 33, null, true],
+            ['aaa', 'bbb', 'id.value', 'id_value_idx', 33, Type::INTEGER, false],
+            ['aaa', 'bbb', 'id.uuid', 'id_uuid_idx', new NonIntegerIdentifierTestClass('80fb6f91-bba1-4d35-b3d4-e06b24494e85'), UuidType::NAME, false],
         ];
     }
 
     /**
      * @dataProvider dataGetFixedQueryBuilder
-     *
-     * @param $class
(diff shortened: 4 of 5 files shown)


eugenekurasov:
I will add tests for `Pager`.

src/Datagrid/ProxyQuery.php
```diff
@@ -79,6 +86,29 @@ public function __clone()
...
 
+    /**
+     * Optimize query with a lot rows.
+     * Not recommended set "false" with left joins.
+     *
+     * @param bool $distinct
+     *
+     * @return ProxyQuery
+     */
+    public function setDistinct($distinct)
```
core23: final please. Same for the other methods

eugenekurasov:
Ping @core23 @kunicmarko20

src/Datagrid/Pager.php
```diff
@@ -36,13 +38,12 @@ public function computeNbResult()
             $countQuery->setParameters($this->getParameters());
         }
 
-        $countQuery->select(sprintf(
-            'count(DISTINCT %s.%s) as cnt',
-            current($countQuery->getRootAliases()),
-            current($this->getCountColumn())
-        ));
+        $query = $countQuery->getQuery();
+        $isDistinct = $countQuery instanceof ProxyQuery ? $countQuery->isDistinct() : true;
```
kunicmarko20: One time variable, can you move ir down and remove `$isDistinct`

eugenekurasov: Ok

kunicmarko20:
You have a todo add documentation, please finish that.

eugenekurasov:
All done. 
@kunicmarko20

kunicmarko20:
How can we disable this in our admin class? I think we should add more details to docs. Cc @sonata-project/contributors

eugenekurasov:
I thought about add new tag for disabled distinct. This PR should be in other repository. 

I added information about disable distinct in proxyQuery in documentation. If you have idea how I can help more - I will do.

``For disable distinct call method setDistinct(false). By default distinct is enable ``

kunicmarko20:
Let's wait for someone else.

eugenekurasov:
I will create new PR in other repository, but this PR will break BC. So I waiting when this PR will merge and then I do it.

kunicmarko20:
If there is any chance, try to avoid BC break.

eugenekurasov:
Ok, I will try don't break BC but I think better need improve ``ProxyQueryInterface`` by add new methods (``setDistinct``, ``distinct``). But if I will not add this methods - I don't break BC.

What you think? @kunicmarko20

greg0ire:
Create a separate interface maybe?

eugenekurasov:
@greg0ire - Hm, good idea. But Then I need create one more PR for this repository but it avoid break BC.

eugenekurasov:
@core23 waiting your approve

docs/reference/query_proxy.rst
```diff
@@ -10,6 +10,7 @@ The ``ProxyQuery`` object is used to add missing features from the original `Doc
 * ``execute`` method - no need to call the ``getQuery()`` method,
 * add sort by and sort order options,
 * add preselect id query on left join query, so a limit query will be only applied on the left statement and not on the full select statement. This simulates the original Doctrine 1 behavior.
+* For disable distinct call method ``setDistinct(false)``. By default distinct is enable.
```
greg0ire: This requires way more explanation IMO. I don't fully understand this PR yet, but here is what I think could be here, tell me if/where I'm wrong, and please fill in the blanks.

By default Sonata will use the `DISTINCT` SQL keyword, because <insert good reason here>. This is not required for queries that <insert description of queries that do not need DISTINCT (or maybe it's necessary on some RDBMS' and not others? I don't know)>. Sonata cannot detect whether or not you need `DISTINCT`, but lets you remove that keyword in case it causes performance issues. To do so, simply call ``setDistinct(false)``

src/Datagrid/Pager.php
```diff
@@ -79,4 +82,22 @@ public function init()
             $this->getQuery()->setMaxResults($this->getMaxPerPage());
         }
     }
+
+    /**
+     * Appends a custom tree walker to the tree walkers hint.
+     *
+     * @param AbstractQuery $query
```
greg0ire: Please remove this line

src/Datagrid/ProxyQuery.php
```diff
@@ -47,6 +47,13 @@ class ProxyQuery implements ProxyQueryInterface
      */
     protected $entityJoinAliases;
 
+    /**
+     * For BC distinct true by default.
```
greg0ire: For BC reasons, this property is true by default.

src/Datagrid/ProxyQuery.php
```diff
@@ -79,6 +86,29 @@ public function __clone()
         $this->queryBuilder = clone $this->queryBuilder;
     }
 
+    /**
+     * Optimize query with a lot rows.
```
greg0ire: Optimize queries with a lot of rows.

src/Datagrid/ProxyQuery.php
```diff
@@ -79,6 +86,29 @@ public function __clone()
...
+     * Not recommended set "false" with left joins.
```
greg0ire: It is not recommended to use "false" with left joins.

src/Datagrid/ProxyQuery.php
```diff
@@ -79,6 +86,29 @@ public function __clone()
...
+     *
+     * @param bool $distinct
+     *
+     * @return ProxyQuery
```
greg0ire: `@return self`

src/Datagrid/ProxyQuery.php
```diff
@@ -79,6 +86,29 @@ public function __clone()
...
+     */
+    final public function setDistinct($distinct)
+    {
+        $this->distinct = (bool) $distinct;
```
greg0ire: Please throw an `\InvalidArgumentException` if `$distinct` is not a boolean. We are going to add a `bool` in next version, which means this will result in an error, which will mean a BC-break.

greg0ire: Alternatively, you can just remove the argument and rename this to `disableDistinct`

tests/Datagrid/ProxyQueryTest.php
```diff
@@ -85,8 +85,12 @@ public function dataGetFixedQueryBuilder()
      * @param $class
      * @param $alias
      * @param $id
+     * @param $expectedId
+     * @param $value
+     * @param $identifierType
+     * @param $distinct
```
greg0ire: Please remove this (and the others above if you want to)

src/Datagrid/Pager.php
eugenekurasov: Why would you want me to delete this line?

greg0ire: Because all the information provided by it is already provided by the type hint.

docs/reference/query_proxy.rst
```diff
@@ -10,6 +10,7 @@ The ``ProxyQuery`` object is used to add missing features from the original `Doc
 * ``execute`` method - no need to call the ``getQuery()`` method,
 * add sort by and sort order options,
 * add preselect id query on left join query, so a limit query will be only applied on the left statement and not on the full select statement. This simulates the original Doctrine 1 behavior.
+* By default Sonata will use the ``DISTINCT`` SQL keyword, because ProxyQuery going to fetch identifiers. Sonata cannot detect whether or not you need ``DISTINCT``, but lets you remove that keyword in case it causes performance issues. To do so, simply call ``setDistinct(false)``.
```
greg0ire: Please comply with the sf docs standards:

> [Each line should break approximately after the first word that crosses the 72nd character (so most lines end up being 72-78 characters);](http://symfony.com/doc/current/contributing/documentation/standards.html)

greg0ire:
```patch
- because ProxyQuery going to fetch identifiers
+ because ``ProxyQuery`` is going to fetch identifiers
```

greg0ire: I don't understand how this justifies the use of distinct, maybe someone else does?

src/Datagrid/Pager.php
```diff
@@ -79,4 +82,22 @@ public function init()
...
+     * @param string        $walkerClass
```
greg0ire: This one says it's supposed to be a `string`, so it's useful.

docs/reference/query_proxy.rst
greg0ire: Is it maybe because when you do a join, you may have several rows for one id, and thus you won't be able to get the exact number of ids you want unless you use `DISTINCT` or `GROUP BY`?

eugenekurasov: https://github.com/sonata-project/SonataAdminBundle/commit/c7f454ed86b0df5f636af90cff5cbac6cad45953#diff-21e8075cdf497941b782a6a3311751cb  - PR when DISTINCT was added.

greg0ire: Ok I think I got it:
```
By default, Sonata will use the ``DISTINCT`` SQL keyword when fetching the identifiers of the entities that will be displayed in the listing, to avoid duplicates in some cases. Sonata cannot detect whether or not you need ``DISTINCT``, but lets you remove that keyword in case it causes performance issues and you are sure there will be no duplicates. To do so, simply call ``setDistinct(false)``.
```

eugenekurasov:
@greg0ire please review again

eugenekurasov:
@kunicmarko20 can you merge this PR?

kunicmarko20:
@eugenekurasov yes but I want someone else to take look, just to be sure. cc @jordisala1991 / @core23

eugenekurasov:
@core23 and @jordisala1991 please review.

src/Datagrid/Pager.php
```diff
@@ -36,13 +38,14 @@ public function computeNbResult()
...
-            current($this->getCountColumn())
-        ));
+        $query = $countQuery->getQuery();
+        $query->setHint(
+            CountWalker::HINT_DISTINCT,
+            $countQuery instanceof ProxyQuery ? $countQuery->isDistinct() : true
+        );
+        $this->appendTreeWalker($query, CountWalker::class);
 
-        return $countQuery->resetDQLPart('orderBy')->getQuery()->getSingleScalarResult();
```
jordisala1991: are we removing the reset of the orderBy?

eugenekurasov: We copy query before manipulation 
``$countQuery = clone $this->getQuery();``

kunicmarko20:
Thank you @eugenekurasov !

greg0ire:
Thanks a lot @eugenekurasov , this bug has been here for literal years!

zhil:
in case if somebody confused about how can it be used (just like me) - you can add to your admin class such a code

```
    {
        $query->setDistinct(false);
        return $query;
    }
```

Probably, it have sense to add that sample code in docs? not sure where should it be added.

PR #829, opened 2018-04-18
Manual merge

greg0ire:
Don't merge, just approve.


src/Datagrid/ProxyQuery.php
```diff
@@ -49,6 +49,13 @@ class ProxyQuery implements ProxyQueryInterface
      */
     protected $entityJoinAliases;
 
+    /**
+     * For BC reasons, this property is true by default.
```
kunicmarko20: should we add something like "Remove on next major?

greg0ire: Yes, but in a separate PR.

PR #843, opened 2018-07-05
Use getNormalizedIdentifier() to retrieve object identifier

leonboot:
I am targeting this branch, because this PR fixes an issue introduced in commit c961daec5a37f0965d39b33c3b95ef43a60184e2.

## Changelog

```markdown
### Fixed
- Block sonata_type_model_list_widget in template Form/form_admin_fields.html.twig now determines the object identifier correctly when building a link to the associated admin
```

## Subject

The sonata_type_model_list form type allows for editing the objects associated by providing an edit link for the associated item. However, the code assumes that the identifier field of the object is always called `id`, resulting in exceptions when objects have an identifier with a different name. Since Admin classes have a `getNormalizedIdentifier($object)` method, it seems to make sense to use this instead of hard-coding the identifier field name.

### DIFF
--- a/src/Resources/views/Form/form_admin_fields.html.twig
+++ b/src/Resources/views/Form/form_admin_fields.html.twig
@@ -100,7 +100,7 @@ file that was distributed with this source code.
                     and btn_edit
                     and sonata_admin.value != null
                 %}
-                    <a  href="{{ sonata_admin.field_description.associationadmin.generateUrl('edit', {'id' : sonata_admin.value.id}) }}"
+                    <a  href="{{ sonata_admin.field_description.associationadmin.generateUrl('edit', {'id' : sonata_admin.field_description.associationAdmin.normalizedIdentifier(sonata_admin.value)}) }}"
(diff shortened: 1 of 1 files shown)


leonboot:
Also, I tried disabling the edit button to work around the issue. Looking at the `if` statement in the template, it should be possible to pass the `btn_edit => false` option to prevent it from being rendered. This seems to work for the ModelListType, but the CollectionType only has the `btn_add` variable as a configurable option. I've tried adding the btn_edit option just as it is implemented in ModelListType, but somehow the value in the template always ends up being set to `link_edit`, instead of the value passed in the `configureFormFields()` method of the Admin class.

leonboot:
@kunicmarko20 @greg0ire I figured it was a temporary glitch, unit tests ran fine locally. Seems the restart worked?

greg0ire:
Thanks @leonboot !

leonboot:
Any estimate on when the next release is due?

OskarStark:
I asked for it internally

leonboot:
@OskarStark @greg0ire Awesome! Thanks for that.

PR #854, opened 2018-10-24
Correct checkbox Class call documentation

franckentien:
I am targeting this branch, because i made a minor fix in documentation 

Based on issue : https://github.com/sonata-project/SonataAdminBundle/issues/4903#issuecomment-383885345

### DIFF
--- a/docs/reference/filter_field_definition.rst
+++ b/docs/reference/filter_field_definition.rst
@@ -250,7 +250,7 @@ In this example, ``getWithOpenCommentField`` and ``getWithOpenCommentFilter`` im
 
                         return true;
                     },
-                    'field_type' => 'checkbox'
+                    'field_type' => CheckboxType::class
                 ))
             ;
         }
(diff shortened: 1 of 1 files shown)


franckentien:
The other field who need to be checked is input_type on this part

```
class PostAdmin extends Admin
{
    protected function configureDatagridFilters(DatagridMapper $datagrid)
    {
        $datagrid
            ->add('timestamp', 'doctrine_orm_datetime_range', array('input_type' => 'timestamp'));
    }
}
```

But i'm not sure

OskarStark:
Could you replace `'doctrine_orm_datetime_range'`, too?

franckentien:
After verification I think the modification is not needed.
Because the `doctrine_orm_datetime_range` was a **filter types**

and in another hand the `field_type` is a **Form Types**
https://symfony.com/doc/current/reference/forms/types.html#other-fields.

PR #868, opened 2018-12-10
Pager::computeNbResult support Doctrine 2.5

qRoC:
https://github.com/sonata-project/SonataAdminBundle/issues/5341

### DIFF
--- a/src/Datagrid/Pager.php
+++ b/src/Datagrid/Pager.php
@@ -12,7 +12,10 @@
 namespace Sonata\DoctrineORMAdminBundle\Datagrid;
 
 use Doctrine\ORM\Query;
+use Doctrine\ORM\Tools\Pagination\CountOutputWalker;
+use Doctrine\ORM\Tools\Pagination\CountWalker;
 use Sonata\AdminBundle\Datagrid\Pager as BasePager;
+use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
 
 /**
  * Doctrine pager class.
@@ -30,29 +33,22 @@ class Pager extends BasePager
 
     public function computeNbResult()
     {
-        $countQuery = clone $this->getQuery();
+        $countQuery = $this->getClonedQuery()->getQuery();
 
-        if (\count($this->getParameters()) > 0) {
-            $countQuery->setParameters($this->getParameters());
-        }
+        $platform = $countQuery->getEntityManager()->getConnection()->getDatabasePlatform();
+        $rsm = new Query\ResultSetMapping();
+        $rsm->addScalarResult($platform->getSQLResultCasing('dctrn_count'), 'count');
+        $countQuery->setHint(Query::HINT_CUSTOM_OUTPUT_WALKER, CountOutputWalker::class);
+        $countQuery->setResultSetMapping($rsm);
 
-        $countQuery->select(sprintf(
-            'count(%s %s) as cnt',
-            $countQuery instanceof ProxyQuery && !$countQuery->isDistinct() ? null : 'DISTINCT',
-            $this->getColumnsForQuery(current($countQuery->getRootAliases()))
-        ));
+        $countQuery->setFirstResult(null)->setMaxResults(null);
 
-        return $countQuery->resetDQLPart('orderBy')->getQuery()->getSingleScalarResult();
+        return $countQuery->getSingleScalarResult();
     }
 
     public function getResults($hydrationMode = Query::HYDRATE_OBJECT)
     {
         return $this->getQuery()->execute([], $hydrationMode);
-    }
-
-    public function getQuery()
-    {
-        return $this->query;
     }
 
     public function init()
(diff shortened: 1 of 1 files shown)


src/Datagrid/Pager.php
```diff
@@ -30,31 +32,24 @@ class Pager extends BasePager
...
-        return $countQuery->resetDQLPart('orderBy')->getQuery()->getSingleScalarResult();
+        return $countQuery->getSingleScalarResult();
     }
 
     public function getResults($hydrationMode = Query::HYDRATE_OBJECT)
     {
         return $this->getQuery()->execute([], $hydrationMode);
     }
 
-    public function getQuery()
```
OskarStark: removing this method is a BC break

OskarStark:
do we need to bump some versions in composer.json?

qRoC:
> do we need to bump some versions in composer.json?

CountOutputWalker added in Doctrine 2.3

qRoC:
> Hi, do you sure this work with composite primary keys?

This generate query:
`SELECT COUNT(*) AS dctrn_count FROM (SELECT DISTINCT keyField1, keyField2, ..., keyFieldN FROM (SELECT allFields... FROM table o0_ WHERE ...) dctrn_result) dctrn_table`

eugenekurasov:
Why you remove usage of ``isDistinct()``?

qRoC:
What cases can be false?

eugenekurasov:
[Issue](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/473).

Here [PR](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/827) where is has been added. Is for stop use distinct for big tables.

qRoC:
>> is tests for composite primary keys.

`$this->assertEquals(0, $count);`

Query:
`SELECT COUNT(*) AS dctrn_count FROM (SELECT DISTINCT store_id_1, product_id_2 FROM (SELECT s0_.name AS name_0, s0_.store_id AS store_id_1, s0_.product_id AS product_id_2 FROM StoreProduct s0_) dctrn_result) dctrn_table`

eugenekurasov:
Why you do it PR?
In [migration_2_5.rst](https://github.com/doctrine/doctrine2/blob/2.6/docs/en/changelog/migration_2_5.rst#dql-improvements) is allow use Expressions inside COUNT. 

> Expressions inside COUNT() now -> allowed <-

`SELECT count (DISTINCT concat (...)) as cnt`  - is valid syntax

qRoC:
Sorry, bad title, error if used v2.5 :)

`composer require doctrine/dbal:v2.5.10 doctrine/common:v2.7.2 doctrine/doctrine-bundle:1.6.7 sonata-project/doctrine-orm-admin-bundle:3.3.0`

ok
___
`composer require doctrine/dbal:v2.6.3 doctrine/common:v2.8.1 doctrine/doctrine-bundle:1.10.0 sonata-project/doctrine-orm-admin-bundle:3.3.0`

ok
___
`composer require doctrine/orm:v2.5.14 doctrine/dbal:v2.6.3 doctrine/common:v2.8.1 doctrine/doctrine-bundle:1.10.0 sonata-project/doctrine-orm-admin-bundle:3.6.3`

`[Semantical Error] line 0, col 31 near 'key1,'-',o.key2))': Error: Invalid PathExpression. Must be a StateFieldPathExpression.`

Pseudo query:

`SELECT count(DISTINCT concat(o.key1,'-',o.key2)) as cnt FROM TestEntity o WHERE ... `
___

eugenekurasov:
Hi @qRoC, Can you check fix for your exception with changes #882

qRoC:
Hi @eugenekurasov, I'll check before the end of the week. Thanks.

qRoC:
@eugenekurasov 
Works!

src/Datagrid/Pager.php
```diff
@@ -80,16 +83,18 @@ public function init()
...
-        if (1 === \count($countColumns = $this->getCountColumn())) {
-            return $rootAlias.'.'.current($countColumns);
-        }
+        /** @var ProxyQueryInterface $cloneQuery */
+        $cloneQuery = clone $this->getQuery();
 
-        $columns = implode(",'-',", array_map(function ($column) use ($rootAlias) {
-            return "$rootAlias.$column";
-        }, $countColumns));
+        if (\count($this->getParameters()) > 0) {
```
phansys: Since `$this->getParameters()` returns an array, I think there is no need for `count()`.

kunicmarko20: I think this is a check if the array is not empty

phansys: In a [loose check](https://www.php.net/manual/en/types.comparisons.php), an empty array already is evaluated as `false`.

qRoC: implicit conversion is not good idea

phansys: AFAIK, there is no conversion, just comparison.

OskarStark: I agree with both of you, but this sounds minor to me and not worth the discussion, as both ways are ok with the same result

phansys: My concern is just about saving a non required function call. Is just for performance reasons. But since the call was already there, it's fine for me if we decide to keep it.

src/Datagrid/Pager.php
```diff
@@ -80,16 +83,18 @@ public function init()
         }
     }
 
-    private function getColumnsForQuery($rootAlias)
+    /**
+     * @return ProxyQueryInterface
```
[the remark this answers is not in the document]
greg0ire: same above

src/Datagrid/Pager.php
```diff
@@ -80,16 +83,18 @@ public function init()
...
+     */
+    private function getClonedQuery()
     {
-        if (1 === \count($countColumns = $this->getCountColumn())) {
-            return $rootAlias.'.'.current($countColumns);
-        }
+        /** @var ProxyQueryInterface $cloneQuery */
```
greg0ire: This should be uneeded after you fix the above comments

PR #875, opened 2018-12-21
Fix deprecated_collection_type

franmomu:
## Subject

If you use the new `Sonata\Form\Type\CollectionType` it doesn't show the fields properly in the form because the new namespace it is not checked in `FormContractor`. I guess the same happens in the MongoDB and Phpcr Bundle.

I am targeting this branch, because is this is a patch.

## Changelog

```markdown
### Fixed
- Fixed using the new collection type namespace
```


franmomu:
It looked like it failed when using a version of CoreBundle where the `Sonata\Form\Type\CollectionType` class didn't exist.

PR #876, opened 2018-12-28
Resolve id object by __toString() if provided

### LINKED ISSUE #842: Breadcrumbs and binary string
### Environment
Linux
#### Sonata packages
```
"sonata-project/doctrine-extensions":       "~1.0",
"sonata-project/cache-bundle":              "2.2.5",
"sonata-project/jquery-bundle":             "1.8.*",
"sonata-project/admin-bundle":              "3.0.0",
"sonata-project/user-bundle":               "3.0.0",
"sonata-project/doctrine-orm-admin-bundle": "3.0.0",
"sonata-project/easy-extends-bundle":       "~2.1",
```

```
#
sonata-project/admin-bundle              3.0.0 3.35.2 The missing Symfony Admin Generator
sonata-project/block-bundle              3.1.1 3.12.1 Symfony SonataBlockBundle
sonata-project/cache                     1.0.7 1.0.7  Cache library
sonata-project/cache-bundle              2.2.5 2.4.2  This bundle provides caching services
sonata-project/core-bundle               3.1.1 3.10.0 Symfony SonataCoreBundle
sonata-project/datagrid-bundle           2.2   2.3.1  Symfony SonataDatagridBundle
sonata-project/doctrine-extensions       1.0.2 1.0.2  Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.0.0 3.6.1  Symfony Sonata / Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/easy-extends-bundle       2.1.5 2.5.0  Symfony SonataEasyExtendsBundle
sonata-project/exporter                  1.3.3 1.9.0  Lightweight Exporter library
sonata-project/google-authenticator      1.0.1 2.0.0  Library to integrate Google Authenticator into a PHP project
sonata-project/jquery-bundle             1.8.3 1.10.2 Symfony SonatajQueryBundle
sonata-project/user-bundle               3.0.0 3.6.0  Symfony SonataUserBundle
```

#### Symfony packages

```
# 
symfony/assetic-bundle     dev-master 4282ddd dev-master 7ffdefe Integrates Assetic into Symfony2
symfony/icu                v1.0.0             v1.2.2             Contains an excerpt of the ICU data and classes to load it.
symfony/monolog-bundle     v2.5.1             v3.3.0             Symfony MonologBundle
symfony/polyfill-apcu      v1.2.0             v1.8.0             Symfony polyfill backporting apcu_* functions to lower PHP versions
symfony/polyfill-mbstring  v1.2.0             v1.8.0             Symfony polyfill for the Mbstring extension
symfony/swiftmailer-bundle v2.4.2             v3.2.2             Symfony SwiftmailerBundle
symfony/symfony            v2.7.19            v3.3.6             The Symfony PHP framework
```

#### PHP version

```
$ php -v
# PHP 7.0.30 (cli) (built: Apr 24 2018 21:41:09) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.30, Copyright (c) 1999-2017, by Zend Technologies
    with Xdebug v2.6.0, Copyright (c) 2002-2018, by Derick Rethans
    with blackfire v1.20.1~linux-x64-non_zts70, https://blackfire.io, by Blackfire
```

## Subject

Related to #547
Using UUID as PK and for InnoDB optimization we're storing it in BINARY, and when doing edit we see binary string in breadcrumbs

https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Model/ModelManager.php#L343

As workaround for this issue we've added:

```
$identifiers = $type->convertToDatabaseValue($value, $platform);
// workaround
if (\strpos($identifier, "\0") !== false) { // binary found
    $identifier = (string) $value;
}
// end workaround
$identifiers[] = $identifier;
```

But it's not a great solution.

torreytsui:
## Subject

When `$entity->getId()` returns an object where its database value is not readable, for example in Bytes format, Admin fails to display the id or generate correct entity url from it.

This is the use case of using UuidBinaryType (as described in https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/842)

I am targeting this branch, because it is BC compatible and it fixes the use case mention above

Closes #842

## Changelog

```markdown
### Added
- Offer support of string resolution for entity id object where `__toString()` will be used if provided. (This also fixes the issue where the id fails to display when the binary / byte datatype is used as primary key, e.g., UUIDBinaryType (aka: InnoDB Optimised Binary UUIDs))

### Changed

### Deprecated

### Removed

### Fixed

### Security
```


src/Model/ModelManager.php
```diff
@@ -337,6 +337,12 @@ public function getIdentifierValues($entity)
                 continue;
             }
 
+            if (method_exists($value, '__toString')) {
```
greg0ire: I'm not sure about this, `__toString` could already be used for other things, is magic, and you cannot throw inside it. How about introducing a new interface that models could implement? That would be a bit more explicit.

kunicmarko20: new interface sounds good, not sure if people want their entities depending on sonata interface?

torreytsui: Interesting, `__toString` indeed could already be used for other things. I have checked the official doc, the magic method however matches our intention - get a string representation of it

http://php.net/manual/en/language.oop5.magic.php#object.tostring

If a __toString is used for purpose other than "treated like a string", it is likely that the implementation does not follow the official interface.

What do you think?

If we can agree on the expected purpose of this magic method, it comes down to an issue where should we consider a BC breaking change.

kunicmarko20: In my entities, I would mostly use `__toString` to return name (or something else, not the id. So, if I had a `Category` entity, my `__toString` would be returning its name. I agree with @greg0ire that `__toString` is not a good solution

torreytsui: Are we talking about the same thing? it is not an `$entity->__toString()`. In this PR, the context is referring to `$entity->getId()->__toString()`

greg0ire: Indeed, I overlooked this too, this looks a bit safer.

greg0ire:
Please document this.

torreytsui:
Thanks for the review. I have documented the usage briefly now. Can you please have a look again?

docs/tutorial/creating_your_first_admin_class/defining_entities.rst
```diff
@@ -256,6 +256,49 @@ Comment
...
+    .. code-block:: php
+
+        <?php
+
+        // Retrieved from \Ramsey\Uuid\Uuid
+        class Uuid implements UuidInterface
+        {
+            // ...
+
+            public function __toString() {
```
OskarStark:
```suggestion
            public function __toString()
            {
```

torreytsui: Thanks! resolved now

docs/tutorial/creating_your_first_admin_class/defining_entities.rst
```diff
@@ -256,6 +256,50 @@ Comment
...
+        class Comment
+        {
+            /**
+             * @var \Ramsey\Uuid\UuidInterface
+             * @Id
+             * @Column(type="uuid_binary_ordered_time", unique=true)
+             * @GeneratedValue(strategy="CUSTOM")
+             * @CustomIdGenerator(class="Ramsey\Uuid\Doctrine\UuidOrderedTimeGenerator")
+             */
+            protected $id;
```
kunicmarko20: why protected?

docs/tutorial/creating_your_first_admin_class/defining_entities.rst
```diff
@@ -256,6 +256,50 @@ Comment
...
+            // ...
+        }
+
+
+    .. code-block:: php
+
+        <?php
+
+        // Retrieved from \Ramsey\Uuid\Uuid
+        class Uuid implements UuidInterface
```
kunicmarko20: what don't you add `use ....` instead of comment?

src/Model/ModelManager.php
kunicmarko20: this should be `\method_exists`, FlintCI slacking again.

docs/tutorial/creating_your_first_admin_class/defining_entities.rst
torreytsui: To be honest, I don't know. The original documentation uses protected. To follow the context, used the same way it does.

https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blame/3.x/docs/tutorial/creating_your_first_admin_class/defining_entities.rst#L220

src/Model/ModelManager.php
torreytsui: Agree, I will update it now

docs/tutorial/creating_your_first_admin_class/defining_entities.rst
OskarStark: please use `private`, we should change the docs you mentioned

docs/tutorial/creating_your_first_admin_class/defining_entities.rst
torreytsui: Like using Trait? Not sure if I understand.

Looks like my documentation here is a bit confusing. Any suggestion to approach this? it is a copy from `\Ramsey\Uuid\Uuid` to show that it is the implementation of `\Ramsey\Uuid\UuidInterface` and has offered the __toString method.

torreytsui: I have given it another attempt now. This might clarify a bit?

kunicmarko20: I meant as import the interface instead of commenting about it

docs/tutorial/creating_your_first_admin_class/defining_entities.rst
```diff
@@ -217,45 +217,91 @@ Comment
...
+            // ...
+        }
+
+    For more details, ``\Ramsey\Uuid\Uuid`` implements ``\Ramsey\Uuid\UuidInterface`` and has offered ``__toString``
+    methods as shown in below pseudo code:
+
+    .. code-block:: php
+
+        <?php
+        namespace Ramsey\Uuid;
```
kunicmarko20: this should really be something like `App\...` and just import the Ramsey uuid

torreytsui: This part supposes to be offered by ramsey/uuid package in the example. Seems it is causing confusion as I show its implementation here. Should I remove this part instead?

docs/tutorial/creating_your_first_admin_class/defining_entities.rst
```diff
@@ -217,45 +217,74 @@ Comment
...
+             * @CustomIdGenerator(class="Ramsey\Uuid\Doctrine\UuidOrderedTimeGenerator")
+             */
+            private $id;
+
+            // ...
+        }
+
+    As ``$comment->getId()`` returns an object of ``\Ramsey\Uuid\UuidInterface`` and the bundle recognizes
+    that it has offered a ``__toString`` method, ``$comment->getId()->__toString()`` is called to resolve
+    the ID string value as part of the entity url generation.
```
torreytsui: @kunicmarko20 Thanks for the feedback. For simplicity, I have rewritten the doc now. Should be less confusing now. Mind to give it another read?

PR #889, opened 2019-01-21
[Proof PR] enable additional rules for PHP-CS-Fixer

OskarStark:


kunicmarko20:
can you check travis?

greg0ire:
The real problem is not about assertions not being performed, though, right?

PR #895, opened 2019-01-24
Applied fixes from FlintCI

Soullivaneuh:
```markdown
# Fixed
- `Sonata\DoctrineORMAdminBundle\Datagrid\Pager::computeNbResult()` now returns an integer, not a string
```

### DIFF
--- a/src/Datagrid/Pager.php
+++ b/src/Datagrid/Pager.php
@@ -45,7 +45,7 @@ class Pager extends BasePager
             current($this->getCountColumn())
         ));
 
-        return $countQuery->resetDQLPart('orderBy')->getQuery()->getSingleScalarResult();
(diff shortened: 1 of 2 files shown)


src/Datagrid/Pager.php
```diff
@@ -45,7 +45,7 @@ public function computeNbResult()
             current($this->getCountColumn())
         ));
 
-        return $countQuery->resetDQLPart('orderBy')->getQuery()->getSingleScalarResult();
+        return (int) ($countQuery->resetDQLPart('orderBy')->getQuery()->getSingleScalarResult());
```
greg0ire: Technically a BC-break, but I think people actually convert this themselves to int if they figured out this was a string, or use implicit conversion.

PR #909, opened 2019-02-27
Update Pager.php

Selion05:
## Subject

I am targeting this branch, because bug fix.

Closes #908

## Changelog

```markdown
### Fixed
908
```

PR #910, opened 2019-02-28
3.8.3

greg0ire:
Just approve

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.8.3](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/compare/3.8.2...3.8.3) - 2019-02-28
+### Fixed
+- Deprecations about core bundle forms
+- Exception on `StringFilter` with null values
+- autocomplete action no longer advertises for more items when there are actually none
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.8.3](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/compare/3.8.2...3.8.3) - 2019-02-28
+### Fixed
+- Deprecations about core bundle forms
+- Exception on `StringFilter` with null values
+- `more` in `get-autocomplete-item` always returns `true`
```
greg0ire: I tried to create a changelog for #909 (fixes #908) but I cannot say I understood what this is about. @ejkun @Selion05 please advise

ejkun: Took me a while, but I figured it out. Ceil function returns float, so lastPage would always be float. But the current page is int. So the function isLastPage ([here](https://github.com/sonata-project/SonataAdminBundle/blob/3.x/src/Datagrid/Pager.php#L418)) would always return false. Which would make `more` in [retrieveAutocompleteItemsAction](https://github.com/sonata-project/SonataAdminBundle/blob/3.x/src/Action/RetrieveAutocompleteItemsAction.php#L171) to always be true.

This happens because `42.0 === 42` returns true.

Can you put this comment on his PR? To help others understand in the future.

greg0ire:
```
>>> 42.0 === 42
=> false
```

ejkun: Yes, sorry.

PR #912, opened 2019-03-21
Check if constructor is callable when creating new model instance

core23:
## Subject

I am targeting this branch, because this feature is BC.

## Changelog

```markdown
### Added
- Added support for protected (no public constructor) entity creation
```


greg0ire:
Thanks @core23 !

PR #928, opened 2019-07-04
Resolve error in entities with composite ID

### LINKED ISSUE #923: Resolve string value of is with __toString cause trouble with composite key
#### Sonata packages

```
sonata-project/admin-bundle              3.48.0 3.49.0 The missing Symfony Admin Generator
sonata-project/block-bundle              3.14.0 3.15.0 Symfony SonataBlockBundle
sonata-project/cache                     2.0.1  2.0.1  Cache library
sonata-project/core-bundle               3.16.1 3.17.0 Symfony SonataCoreBundle
sonata-project/datagrid-bundle           2.4.0  3.0.0  Symfony SonataDatagridBundle
sonata-project/doctrine-extensions       1.1.5  1.3.0  Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.9.0  3.9.0  Symfony Sonata / Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/easy-extends-bundle       2.5.0  2.5.0  Symfony SonataEasyExtendsBundle
sonata-project/exporter                  1.11.0 2.0.1  Lightweight Exporter library
sonata-project/intl-bundle               2.5.0  2.6.0  Symfony SonataIntlBundle
```

#### Symfony packages

```
symfony/asset                   v4.2.3  v4.3.0  Symfony Asset Component
symfony/browser-kit             v4.2.3  v4.3.0  Symfony BrowserKit Component
symfony/cache                   v4.2.3  v4.3.0  Symfony Cache component with PSR-6, PSR-16, and tags
symfony/config                  v4.2.3  v4.3.0  Symfony Config Component
symfony/console                 v4.2.3  v4.3.0  Symfony Console Component
symfony/contracts               v1.0.2  v1.1.2  A set of abstractions extracted out of the Symfony components
symfony/css-selector            v4.2.3  v4.3.0  Symfony CssSelector Component
symfony/debug                   v4.2.3  v4.3.0  Symfony Debug Component
symfony/debug-bundle            v4.2.3  v4.3.0  Symfony DebugBundle
symfony/dependency-injection    v4.2.3  v4.3.0  Symfony DependencyInjection Component
symfony/doctrine-bridge         v4.2.3  v4.3.0  Symfony Doctrine Bridge
symfony/dom-crawler             v4.2.3  v4.3.0  Symfony DomCrawler Component
symfony/dotenv                  v4.2.3  v4.3.0  Registers environment variables from a .env file
symfony/event-dispatcher        v4.2.3  v4.3.0  Symfony EventDispatcher Component
symfony/expression-language     v4.2.3  v4.3.0  Symfony ExpressionLanguage Component
symfony/filesystem              v4.2.3  v4.3.0  Symfony Filesystem Component
symfony/finder                  v4.2.3  v4.3.0  Symfony Finder Component
symfony/flex                    v1.1.8  v1.2.5  Composer plugin for Symfony
symfony/form                    v4.2.3  v4.3.0  Symfony Form Component
symfony/framework-bundle        v4.2.3  v4.3.0  Symfony FrameworkBundle
symfony/http-foundation         v4.2.3  v4.3.0  Symfony HttpFoundation Component
symfony/http-kernel             v4.2.3  v4.3.0  Symfony HttpKernel Component
symfony/inflector               v4.2.3  v4.3.0  Symfony Inflector Component
symfony/intl                    v4.2.3  v4.3.0  A PHP replacement layer for the C intl extension that includes additional data from the ICU li...
symfony/lock                    v4.2.3  v4.3.0  Symfony Lock Component
symfony/monolog-bridge          v4.2.3  v4.3.0  Symfony Monolog Bridge
symfony/monolog-bundle          v3.3.1  v3.3.1  Symfony MonologBundle
symfony/options-resolver        v4.2.3  v4.3.0  Symfony OptionsResolver Component
symfony/phpunit-bridge          v4.2.3  v4.3.0  Symfony PHPUnit Bridge
symfony/polyfill-ctype          v1.10.0 v1.11.0 Symfony polyfill for ctype functions
symfony/polyfill-intl-icu       v1.10.0 v1.11.0 Symfony polyfill for intl's ICU-related data and classes
symfony/polyfill-mbstring       v1.10.0 v1.11.0 Symfony polyfill for the Mbstring extension
symfony/polyfill-php56          v1.10.0 v1.11.0 Symfony polyfill backporting some PHP 5.6+ features to lower PHP versions
symfony/polyfill-php72          v1.10.0 v1.11.0 Symfony polyfill backporting some PHP 7.2+ features to lower PHP versions
symfony/polyfill-util           v1.10.0 v1.11.0 Symfony utilities for portability of PHP codes
symfony/process                 v4.2.3  v4.3.0  Symfony Process Component
symfony/property-access         v4.2.3  v4.3.0  Symfony PropertyAccess Component
symfony/property-info           v4.2.3  v4.3.0  Symfony Property Info Component
symfony/proxy-manager-bridge    v4.2.3  v4.3.0  Symfony ProxyManager Bridge
symfony/psr-http-message-bridge v1.1.0  v1.2.0  PSR HTTP message bridge
symfony/routing                 v4.2.3  v4.3.0  Symfony Routing Component
symfony/security                v4.2.3  v4.3.0  Symfony Security Component
symfony/security-acl            v3.0.1  v3.0.2  Symfony Security Component - ACL (Access Control List)
symfony/security-bundle         v4.2.3  v4.3.0  Symfony SecurityBundle
symfony/serializer              v4.2.3  v4.3.0  Symfony Serializer Component
symfony/stopwatch               v4.2.3  v4.3.0  Symfony Stopwatch Component
symfony/swiftmailer-bundle      v3.2.5  v3.2.7  Symfony SwiftmailerBundle
symfony/templating              v4.2.3  v4.3.0  Symfony Templating Component
symfony/translation             v4.2.3  v4.3.0  Symfony Translation Component
symfony/twig-bridge             v4.2.3  v4.3.0  Symfony Twig Bridge
symfony/twig-bundle             v4.2.3  v4.3.0  Symfony TwigBundle
symfony/validator               v4.2.3  v4.3.0  Symfony Validator Component
symfony/var-dumper              v4.2.3  v4.3.0  Symfony mechanism for exploring and dumping PHP variables
symfony/var-exporter            v4.2.3  v4.3.0  A blend of var_export() + serialize() to turn any serializable data structure to plain PHP code
symfony/web-link                v4.2.3  v4.3.0  Symfony WebLink Component
symfony/web-profiler-bundle     v4.2.5  v4.3.0  Symfony WebProfilerBundle
symfony/webpack-encore-bundle   v1.0.0  v1.6.0  Integration with your Symfony app & Webpack Encore!
symfony/yaml                    v4.2.3  v4.3.0  Symfony Yaml Component
```

#### PHP version

```
$ php -v
PHP 7.2.16 (cli) (built: Mar  9 2019 01:44:07) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
```

## Subject

In 75ac3a3efb95a4618a51a6033463386de4e7afc6 (bisected), the `__toString` method is used if available.

This is wrong, at least for composite key. I don't see why `__toString` can be trust as an id.

## Steps to reproduce

Create an entity with composite key (two ID field) and ass a `__toString` method returning an another thing than the representative ID.

## Expected results

The composite ID should be generated, without using `__toString`.

## Actual results

The `__toString` method is used, generating bud identifier.

MAGlCUS:
## Subject

Solution for [#923](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/923).

While a more elegant solution is found for this case, this PR would fix the errors when generating URLs in Sonata Admin when using entities with composite and foreign keys. See manual of doctrine: [Link](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/tutorials/composite-primary-keys.html)

Closes #923 

## Changelog

```markdown
### Fixed
- Fix "Resolve string value of is with __toString cause trouble with composite key" (#923)
```

### DIFF
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -16,6 +16,7 @@ namespace Sonata\DoctrineORMAdminBundle\Model;
 use Doctrine\Common\Util\ClassUtils;
 use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\LockMode;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\Mapping\ClassMetadata;
@@ -339,16 +340,10 @@ class ModelManager implements ModelManagerInterface, LockInterface
                 continue;
             }
 
-            if (method_exists($value, '__toString')) {
-                $identifiers[] = (string) $value;
-
-                continue;
-            }
-
             $fieldType = $metadata->getTypeOfField($name);
             $type = $fieldType && Type::hasType($fieldType) ? Type::getType($fieldType) : null;
             if ($type) {
-                $identifiers[] = $type->convertToDatabaseValue($value, $platform);
+                $identifiers[] = $this->getValueFromType($value, $type, $fieldType, $platform);
 
                 continue;
             }
@@ -632,5 +627,17 @@ class ModelManager implements ModelManagerInterface, LockInterface
     protected function camelize($property)
     {
         return str_replace(' ', '', ucwords(str_replace('_', ' ', $property)));
+    }
+
+    /**
+     * @param mixed $value
+     */
+    private function getValueFromType($value, Type $type, string $fieldType, AbstractPlatform $platform): string
+    {
+        if ('binary' === $platform->getDoctrineTypeMapping($fieldType)) {
+            return (string) $type->convertToPHPValue($value, $platform);
+        }
+
+        return $type->convertToDatabaseValue($value, $platform);
     }
 }
--- a/tests/Fixtures/DoctrineType/UuidBinaryType.php
+++ b/tests/Fixtures/DoctrineType/UuidBinaryType.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DoctrineORMAdminBundle\Tests\Fixtures\DoctrineType;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\StringType;
+use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Util\NonIntegerIdentifierTestClass;
+
+/**
+ * Mock for a custom doctrine type used in the ModelManagerTest suite.
+ *
+ * @author Jorge Garces <jgarces@iberdat.com>
+ */
+final class UuidBinaryType extends StringType
+{
+    const NAME = 'uuid_binary';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return self::NAME;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToPHPValue($value, AbstractPlatform $platform)
+    {
+        return !empty($value) ? new NonIntegerIdentifierTestClass($value->toString()) : null;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToDatabaseValue($value, AbstractPlatform $platform)
+    {
+        return $value->toString();
+    }
+}
--- a/tests/Model/ModelManagerTest.php
+++ b/tests/Model/ModelManagerTest.php
@@ -37,6 +37,7 @@ use Sonata\DoctrineORMAdminBundle\Admin\FieldDescription;
 use Sonata\DoctrineORMAdminBundle\Datagrid\OrderByToSelectWalker;
 use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery;
 use Sonata\DoctrineORMAdminBundle\Model\ModelManager;
+use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\DoctrineType\UuidBinaryType;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\DoctrineType\UuidType;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\AbstractEntity;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\AssociatedEntity;
@@ -47,7 +48,6 @@ use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\ProtectedEntity;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\SimpleEntity;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\UuidEntity;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\VersionedEntity;
-use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Util\ClassWithToStringSupport;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Util\NonIntegerIdentifierTestClass;
 use Symfony\Bridge\Doctrine\RegistryInterface;
 
@@ -57,6 +57,9 @@ class ModelManagerTest extends TestCase
     {
         if (!Type::hasType('uuid')) {
             Type::addType('uuid', UuidType::class);
+        }
+        if (!Type::hasType('uuid_binary')) {
+            Type::addType('uuid_binary', UuidBinaryType::class);
         }
     }
 
@@ -368,16 +371,19 @@ class ModelManagerTest extends TestCase
         return $metadata;
     }
 
-    public function testGetIdentifierValuesForIdInObjectTypeWithToStringSupport(): void
+    public function testGetIdentifierValuesForIdInObjectTypeBinaryToStringSupport(): void
     {
-        $entityId = new ClassWithToStringSupport('to-string-id');
+        $uuid = new NonIntegerIdentifierTestClass('efbcfc4b-8c43-4d42-aa4c-d707e55151ac');
 
-        $entity = new \stdClass();
+        $entity = new UuidEntity($uuid);
 
         $meta = $this->createMock(ClassMetadata::class);
         $meta->expects($this->any())
             ->method('getIdentifierValues')
-            ->willReturn([$entityId]);
+            ->willReturn([$entity->getId()]);
+        $meta->expects($this->any())
+            ->method('getTypeOfField')
+            ->willReturn(UuidBinaryType::NAME); //'uuid_binary'
 
         $mf = $this->createMock(ClassMetadataFactory::class);
         $mf->expects($this->any())
@@ -385,6 +391,9 @@ class ModelManagerTest extends TestCase
             ->willReturn($meta);
 
         $platform = $this->createMock(PostgreSqlPlatform::class);
+        $platform->expects($this->any())
+                ->method('getDoctrineTypeMapping')
+                ->willReturn('binary');
 
         $conn = $this->createMock(Connection::class);
         $conn->expects($this->any())
@@ -407,7 +416,7 @@ class ModelManagerTest extends TestCase
         $manager = new ModelManager($registry);
         $result = $manager->getIdentifierValues($entity);
 
-        $this->assertSame('to-string-id', $result[0]);
+        $this->assertSame($entity->getId()->toString(), $result[0]);
     }
 
     public function testNonIntegerIdentifierType(): void


src/Model/ModelManager.php
```diff
@@ -339,12 +339,6 @@ public function getIdentifierValues($entity)
                 continue;
             }
 
-            if (method_exists($value, '__toString')) {
-                $identifiers[] = (string) $value;
-
-                continue;
-            }
-
```
torreytsui: IIRC, optimised UUID failed (which shows in the travis) as it meets the below `if($type)` condition. Meaning, `__toString` likely needs to place before the `$type` check.

I did a little bit digging on the doctrine composite key a few weeks ago and notice that it is resolved by the `$metadata->getIdentifierValues(...)` below.

Maybe, for a quick win to serve all known use cases, we probably just need to check with the $metadata condition.

```php
if (method_exists($value, '__toString') && 0 === \count($metadata->getIdentifierValues(...)) {
    // ...
}
```

torreytsui: (I didn't have time to test it and therefore didn't PR)

core23:
Can you please add a test to reproduce the original problem?

src/Model/ModelManager.php
MAGlCUS: I make test with UuidBinaryOrderedTimeType of Ramsey and with the last solution is working. 
We have to test if the object for the id field is a binary type in DB or not. If it is binary, it has no sense that we use "convertToDatabaseValue" due to we get a binary value. We need use convertToPHPValue instead and cast it to string.
With this we get solution for both case (ID composite and Binary object ID)

torreytsui: Doing binary type check makes sense for the UuidBinaryOrderedTimeType. However this approach seems not so generalised. Object as an ID may not always persist as binary. For example, UuidType persists as string (https://github.com/ramsey/uuid-doctrine/blob/master/src/UuidType.php#L91) and the current proposal may break it.

MAGlCUS: No @torreytsui, there is no that problem. If you use UuidType like Ramsey's bundle, the database type of this object is not binary. In this case, it's used $type->convertToDatabaseValue.
It doesn't make sense to check if the object has the __toString method when the value is already stored in DB in a format that can be "used" as an index in Sonata Admin.
This solution cover all the case of entity's id that you have.

torreytsui: Oh, I see. I overlooked it, it makes more sense to me now. Basically, we assume that all non-binary type db values can be used directly.

src/Model/ModelManager.php
```diff
@@ -339,17 +339,13 @@ public function getIdentifierValues($entity)
...
             $fieldType = $metadata->getTypeOfField($name);
             $type = $fieldType && Type::hasType($fieldType) ? Type::getType($fieldType) : null;
-            if ($type) {
+            if ($type and 'binary' === $platform->getDoctrineTypeMapping($fieldType)) {
```
phansys: Please, try to simplify these checks:

```php
if ($type) {
    $identifiers[] = 'binary' === $platform->getDoctrineTypeMapping($fieldType) ? (string) $type->convertToPHPValue($value, $platform) : $type->convertToDatabaseValue($value, $platform);
 
    continue;
}
```

tests/Fixtures/DoctrineType/UuidBinaryType.php
```diff
@@ -0,0 +1,47 @@
...
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\StringType;
+use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Util\NonIntegerIdentifierTestClass;
+
+/**
+ * Mock for a custom doctrine type used in the ModelManagerTest suite.
+ *
+ * @author Jeroen Thora <jeroen.thora@gmail.com>
+ */
+class UuidBinaryType extends StringType
```
phansys:
```suggestion
final class UuidBinaryType extends StringType
```

tests/Fixtures/DoctrineType/UuidBinaryType.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    const NAME = 'uuid_binary';
+
+    public function getName()
```
phansys: Please, add `{@inheritdoc}` docblock.

src/Model/ModelManager.php
MAGlCUS: Done.

tests/Fixtures/DoctrineType/UuidBinaryType.php
MAGlCUS: done

tests/Fixtures/DoctrineType/UuidBinaryType.php
MAGlCUS: done

src/Model/ModelManager.php
phansys: Please, respect CS by leaving blankspaces around ternary operators (`?` and `:`).

MAGlCUS: Done

src/Model/ModelManager.php
```diff
@@ -339,17 +339,10 @@ public function getIdentifierValues($entity)
...
-                continue;
-            }
-
             $fieldType = $metadata->getTypeOfField($name);
             $type = $fieldType && Type::hasType($fieldType) ? Type::getType($fieldType) : null;
             if ($type) {
-                $identifiers[] = $type->convertToDatabaseValue($value, $platform);
-
+                $identifiers[] = 'binary' === $platform->getDoctrineTypeMapping($fieldType) ? (string) $type->convertToPHPValue($value, $platform) : $type->convertToDatabaseValue($value, $platform);
                 continue;
```
phansys: Please, add a blank line before `continue` statement in order to respect **blank_line_before_statement** rule (I don't know why `flintci/pr` isn't failing with this violation).

MAGlCUS: Done

src/Model/ModelManager.php
```diff
@@ -339,16 +339,10 @@ public function getIdentifierValues($entity)
...
-                $identifiers[] = (string) $value;
-
-                continue;
-            }
-
             $fieldType = $metadata->getTypeOfField($name);
             $type = $fieldType && Type::hasType($fieldType) ? Type::getType($fieldType) : null;
             if ($type) {
-                $identifiers[] = $type->convertToDatabaseValue($value, $platform);
+                $identifiers[] = 'binary' === $platform->getDoctrineTypeMapping($fieldType) ? (string) $type->convertToPHPValue($value, $platform) : $type->convertToDatabaseValue($value, $platform);
```
core23: Please remove the ternary operator here and use a simple if statement, because this line is too long for a quick understanding

MAGlCUS: I think that in this way it is a little confuse, but @phansys ask for that change. Please, could you two agree?

phansys: I think what @core23 is proposing is different of how it was originally, consider this:
```php
if ($type) {
    if ('binary' === $platform->getDoctrineTypeMapping($fieldType)) {
        $identifiers[] = ? (string) $type->convertToPHPValue($value, $platform);
    } else {
        $identifiers[] = $type->convertToDatabaseValue($value, $platform);
    }
}
```

MAGlCUS: With that approach there are too many if blocks nested. But if you think that it is better, ok ;-)

phansys: Regardless the nested `if`s, personally the original approach looks more confusing to me since both `if` and `elseif` conditions were checking against `$type` evaluating to `true`.
Whatever, I'll approve the PR since the CS issues were fixed :+1:

MAGlCUS: We will wait for the @core23 opinion to make the fix.

core23: We should not write lines that are longer than 80/120 chars (@greg0ire  😅 ).

Maybe the best way is a separate method that will return the correct identifiert if any other way is too confusing.

MAGlCUS: Ok, I was thinking about that and I propose: 
```suggestion
                $identifiers[] = $this->getValueFromType($value, $type, $fieldType, $platform);
```
With this function:
```suggestion
    /**
     * 
     * @param mix $value
     * @param Doctrine\DBAL\Types\Type $type
     * @param string $fieldType
     * @param Doctrine\DBAL\Platforms\AbstractPlatform $platform
     */
    private function getValueFromType ($value, $type, $fieldType, $platform) 
    {
        if ('binary' === $platform->getDoctrineTypeMapping($fieldType)){
            return (string) $type->convertToPHPValue($value, $platform);
        }else {
            return $type->convertToDatabaseValue($value, $platform);
        } 
    }
```
What do you think?

phansys: I'm ok with this if you use real type declarations where possible and if the superfluous `else` is removed:
```php
    /**
     * @param mixed $value
     * 
     * @return mixed
     */
    private function getValueFromType($value, Type $type, string $fieldType, AbstractPlatform $platform) 
    {
        if ('binary' === $platform->getDoctrineTypeMapping($fieldType)){
            return (string) $type->convertToPHPValue($value, $platform);
        }

        return $type->convertToDatabaseValue($value, $platform); 
    }

Comment (MAGlCUS) [reply src/Model/ModelManager.php]: Done

File: src/Model/ModelManager.php [hunk 7]
```diff
@@ -633,4 +628,18 @@ protected function camelize($property)
     {
         return str_replace(' ', '', ucwords(str_replace('_', ' ', $property)));
     }
+
+    /**
+     * @param mix $value
```
phansys:
```suggestion
     * @param mixed $value
```

phansys:
Great work @MAGlCUS!
Could you please squash your commits?

src/Model/ModelManager.php
```diff
@@ -633,4 +628,18 @@ protected function camelize($property)
     {
         return str_replace(' ', '', ucwords(str_replace('_', ' ', $property)));
     }
+
+    /**
+     * @param mixed $value
+     *
+     * @return string
+     */
+    private function getValueFromType($value, Type $type, string $fieldType, AbstractPlatform $platform)
```
core23: You can remove the `@return` annotation and use proper type hints here

```suggestion
    private function getValueFromType($value, Type $type, string $fieldType, AbstractPlatform $platform): string
```

MAGlCUS: Other functions in the file use "@return" and not ": type" in the function declaration. I think it's better if we use the same rules in all the file.

core23: That's right for old parts. We can't add the return type on existing methods, because this would be a BC break.
But if you write a new method, you should use it where possible.

After this change, your PR is ready to merge :+1:

core23:
Thanks @MAGlCUS

franmomu:
There is a problem with this, if you use https://github.com/ramsey/uuid-doctrine and UuidType as in the documentation, it fails, `$platform->getDoctrineTypeMapping($fieldType)` throws an exception (`An exception has been thrown during the rendering of a template ("Unknown database type uuid requested, Doctrine\DBAL\Platforms\MySQL80Platform may not support it.").` in my case).

I had to add
```
doctrine:
    dbal:
        types:
            uuid:  Ramsey\Uuid\Doctrine\UuidType
+        mapping_types:
+            uuid: string

```
to make it work

core23:
Can you provide a PR with a better solution @franmomu

peter-gribanov:
I have a similar problem after update 3.9.0 => 3.10.0
```
Unknown database type shopshowcase2productid requested, Doctrine\DBAL\Platforms\MySQL57Platform may not support it.
```
Although this type exists
```yml
doctrine:
    dbal:
        types:
            ShopShowcase2ProductId: Shop\Showcase2\Persistence\DBAL\Type\ProductIdType
```
And it is available
```
# ./bin/console debug:config doctrine

Current configuration for extension with alias "doctrine"
=========================================================

doctrine:
    dbal:
        types:
            ShopShowcase2ProductId:
                class: Shop\Showcase2\Persistence\DBAL\Type\ProductIdType
                commented: null

```

peter-gribanov:
@core23 i create fix for this error #935

PR #934, opened 2019-08-04
Expose const for Operator choices in filter classes

### LINKED ISSUE #920: NumberFilter (and others) getOperator should be public
## Feature Request

I saw this issue https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/515 and I kinda disagree with the fact `getOperator` is private. When I'm doing query by myself, I'd like to have access to `getOperator`.

Or perhaps I'm doing wrong in my use case.

This is my use case:
Admin
```php
$datagridMapper
            ->add(
                'totalSavings',
                'doctrine_orm_callback',
                [
                    'callback'         => [$this->contractAdminRepository, 'getByTotalSavingsFilter'],
                    'field_type'       => 'integer',
                    'operator_type'    => 'choice',
                    'operator_options' => [
                        'choices' => NumberFilter::$operators,
                    ],
                ],
                'integer'
            )
```

Repository
```php
// Injecting numberFilter in __construct

public function getByRelaunchesNumberFilter($queryBuilder, $alias, $field, $value)
    {
        if (!$value['value'] && 0 !== $value['value']) {
            return;
        }

        $operator = $this->numberFilter->getOperator();
        $operator = $operator ? $operator : '=';

        ... // Big query

        return true;
    }
```

I'd like to avoid replicating the `numberFilter` or the `getOperator` function

VincentLanglet:
## Subject

I am targeting this branch, because I expose a new constant

Closes #920

## Changelog

```markdown
### Changed
Create const for operator choices in Filter classes
```

### DIFF
--- a/src/Filter/AbstractDateFilter.php
+++ b/src/Filter/AbstractDateFilter.php
@@ -21,6 +21,16 @@ use Sonata\AdminBundle\Form\Type\Filter\DateType;
 
 abstract class AbstractDateFilter extends Filter
 {
+    public const CHOICES = [
+        DateType::TYPE_EQUAL => '=',
+        DateType::TYPE_GREATER_EQUAL => '>=',
+        DateType::TYPE_GREATER_THAN => '>',
+        DateType::TYPE_LESS_EQUAL => '<=',
+        DateType::TYPE_LESS_THAN => '<',
+        DateType::TYPE_NULL => 'NULL',
+        DateType::TYPE_NOT_NULL => 'NOT NULL',
+    ];
+
     /**
      * Flag indicating that filter will have range.
      *
@@ -178,16 +188,6 @@ abstract class AbstractDateFilter extends Filter
     {
         $type = (int) $type;
 
-        $choices = [
-            DateType::TYPE_EQUAL => '=',
-            DateType::TYPE_GREATER_EQUAL => '>=',
-            DateType::TYPE_GREATER_THAN => '>',
-            DateType::TYPE_LESS_EQUAL => '<=',
-            DateType::TYPE_LESS_THAN => '<',
-            DateType::TYPE_NULL => 'NULL',
-            DateType::TYPE_NOT_NULL => 'NOT NULL',
-        ];
-
-        return $choices[$type] ?? '=';
+        return self::CHOICES[$type] ?? '=';
     }
 }
--- a/src/Filter/ClassFilter.php
+++ b/src/Filter/ClassFilter.php
@@ -21,6 +21,11 @@ use Symfony\Component\Form\FormTypeInterface;
 
 class ClassFilter extends Filter
 {
+    public const CHOICES = [
+        EqualType::TYPE_IS_EQUAL => 'INSTANCE OF',
+        EqualType::TYPE_IS_NOT_EQUAL => 'NOT INSTANCE OF',
+    ];
+
     public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
     {
         if (!$data || !\is_array($data) || !\array_key_exists('value', $data)) {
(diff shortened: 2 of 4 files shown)


core23:
Why do you need this?

VincentLanglet:
Indeed, I don't need this.

What I need is to export the array
```
    [
        NumberType::TYPE_EQUAL => '=',
        NumberType::TYPE_GREATER_EQUAL => '>=',
        NumberType::TYPE_GREATER_THAN => '>',
        NumberType::TYPE_LESS_EQUAL => '<=',
        NumberType::TYPE_LESS_THAN => '<',
    ]
```

My use case would be
```
->add('nbOfSomething', 'doctrine_orm_callback', [
                'callback'         => [$this, 'getNbOfSomething'],
                'field_type' => SymfonyNumberType::class,
                'operator_type' => SymfonyChoiceType::class,
                'operator_options' => SonataNumberType::CHOICE_OPTIONS,
])
```
With
```
public function getNbOfSomething($queryBuilder, $alias, $field, $value)
    {
        if (!$value['value']) {
            return;
        }

        $queryBuilder->andWhere("$alias.$field".NumberFilter::CHOICES[$value].' :'.$field);
        $queryBuilder->setParameter($field, $value['value']);

        return true;
    }
```

I'd like to expose also a `CHOICE_OPTIONS` const in `SonataProject:AdminBundle` in `Form/Type/Filter`, for example for `NumberType`.
```
public const CHOICE_OPTIONS = [
    'choices' => [
            'label_type_equal' => self::TYPE_EQUAL,
            'label_type_greater_equal' => self::TYPE_GREATER_EQUAL,
            'label_type_greater_than' => self::TYPE_GREATER_THAN,
            'label_type_less_equal' => self::TYPE_LESS_EQUAL,
            'label_type_less_than' => self::TYPE_LESS_THAN,
        ],
    'choice_translation_domain' => 'SonataAdminBundle',
    'required' => false,
]
```

Actually I have duplicated both Filter::CHOICES and Type:CHOICE_OPTIONS in my project in order to use it.

core23:
Given your example. I'm not sure if we need to expose the whole choice list. You only need the operator of a given choice value. Maybe it's better to change the visibilty of `getOperator`

VincentLanglet:
> Given your example. I'm not sure if we need to expose the whole choice list. You only need the operator of a given choice value. Maybe it's better to change the visibilty of `getOperator`

Both works.
I found the export of the whole choice list more flexible and avoids a potential breaking change.
The `getOperator` function is just a `$choice[$value] ?? false`.

core23:
>  I found the export of the whole choice list more flexible and avoids a potential breaking change.

IMHO this would end in more BC breaks. If we want to add a new options to the list later, we can't do this on a stable branch, because this could end in a BC break for someone that is only using the existing choices.

VincentLanglet:
Is this really a BC break ?
For me it's the same than "If we want to add a feature and someone does not want it, it's a BC break for him".

Plus If you add an option in the choice list, it will be the same result if you export the choice list or you export the function `getOperator` since the function is directly calling the choice list. So I don't see why exporting the choice list will introduce more BC-break than exporting the function.

And to conclude, I don't believe there will be new operator since it's directly SQL-related.

VincentLanglet:
Hi @core23, does this PR have chance to be merged or should I close it ?

VincentLanglet:
Hi, can someone merge this PR ? Thanks :)

greg0ire:
Thanks @VincentLanglet

greg0ire:
And I guess you'd like a release :P

VincentLanglet:
It's not urgent, but yeah it'll help ;)

greg0ire:
Here you go: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/releases/tag/3.11.0

PR #935, opened 2019-08-13
Сheck platform type is exists before get it

peter-gribanov:
## Subject

Fix a break BC error
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/928#issuecomment-520774529

DoctrineTypeMapping contain only internal types. Trying to get a custom type in [this](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Model/ModelManager.php#L637) place will always result in error.

```php
array (
  'tinyint' => 'boolean',
  'smallint' => 'smallint',
  'mediumint' => 'integer',
  'int' => 'integer',
  'integer' => 'integer',
  'bigint' => 'bigint',
  'tinytext' => 'text',
  'mediumtext' => 'text',
  'longtext' => 'text',
  'text' => 'text',
  'varchar' => 'string',
  'string' => 'string',
  'char' => 'string',
  'date' => 'date',
  'datetime' => 'datetime',
  'timestamp' => 'datetime',
  'time' => 'time',
  'float' => 'float',
  'double' => 'float',
  'real' => 'float',
  'decimal' => 'decimal',
  'numeric' => 'decimal',
  'year' => 'date',
  'longblob' => 'blob',
  'blob' => 'blob',
  'mediumblob' => 'blob',
  'tinyblob' => 'blob',
  'binary' => 'binary',
  'varbinary' => 'binary',
  'set' => 'simple_array',
  'json' => 'json',
  'enum' => 'string',
)
```

## Changelog

```markdown
### Fixed
- Fix a break BC error
```

### DIFF
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -634,10 +634,12 @@ class ModelManager implements ModelManagerInterface, LockInterface
      */
     private function getValueFromType($value, Type $type, string $fieldType, AbstractPlatform $platform): string
     {
-        if ('binary' === $platform->getDoctrineTypeMapping($fieldType)) {
+        if ($platform->hasDoctrineTypeMappingFor($fieldType) &&
+            'binary' === $platform->getDoctrineTypeMapping($fieldType)
+        ) {
             return (string) $type->convertToPHPValue($value, $platform);
         }
 
-        return $type->convertToDatabaseValue($value, $platform);
+        return (string) $type->convertToDatabaseValue($value, $platform);
     }
 }
--- a/tests/Fixtures/DoctrineType/ProductIdType.php
+++ b/tests/Fixtures/DoctrineType/ProductIdType.php
@@ -27,7 +27,7 @@ final class ProductIdType extends Type
      *
      * @return string
      */
-    public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform)
+    public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform): string
     {
         return $platform->getIntegerTypeDeclarationSQL($fieldDeclaration);
     }
@@ -81,7 +81,7 @@ final class ProductIdType extends Type
     /**
      * @return string
      */
-    public function getName()
+    public function getName(): string
     {
         return self::NAME;
     }
--- a/tests/Fixtures/DoctrineType/ProductIdType.php
+++ b/tests/Fixtures/DoctrineType/ProductIdType.php
@@ -49,10 +49,8 @@ final class ProductIdType extends Type
 
     /**
      * @param mixed $value
-     *
-     * @return int|null
      */
-    public function convertToDatabaseValue($value, AbstractPlatform $platform)
+    public function convertToDatabaseValue($value, AbstractPlatform $platform): ?int
     {
         $value = $this->convertToPHPValue($value, $platform);
 
--- a/tests/Fixtures/DoctrineType/ProductIdType.php
+++ b/tests/Fixtures/DoctrineType/ProductIdType.php
@@ -21,34 +21,18 @@ final class ProductIdType extends Type
 {
     const NAME = 'ProductId';
 
-    /**
-     * @param array            $fieldDeclaration
-     * @param AbstractPlatform $platform
-     *
-     * @return string
-     */
     public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform): string
     {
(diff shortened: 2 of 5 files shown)


src/Model/ModelManager.php
```diff
@@ -634,10 +634,12 @@ protected function camelize($property)
...
     {
-        if ('binary' === $platform->getDoctrineTypeMapping($fieldType)) {
+        if ($platform->hasDoctrineTypeMappingFor($fieldType) &&
+            'binary' === $platform->getDoctrineTypeMapping($fieldType)
+        ) {
             return (string) $type->convertToPHPValue($value, $platform);
         }
 
-        return $type->convertToDatabaseValue($value, $platform);
+        return (string) $type->convertToDatabaseValue($value, $platform);
```
phansys: Why is the casting required?
Could you please add a test covering this fix?

peter-gribanov: The return type of [`Doctrine\DBAL\Types\Type::convertToDatabaseValue()`](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Types/Type.php#L144) may be a mixed and this violates the declared return type of `Sonata\DoctrineORMAdminBundle\Model\ModelManager::getValueFromType()`. So, we must either cast the value to a string or change the return type declaration.

peter-gribanov: @phansys i update tests

tests/Fixtures/DoctrineType/ProductIdType.php
```diff
@@ -0,0 +1,88 @@
...
+     * @param mixed            $value
+     * @param AbstractPlatform $platform
+     *
+     * @return mixed
+     */
+    public function convertToDatabaseValue($value, AbstractPlatform $platform)
+    {
+        $value = $this->convertToPHPValue($value, $platform);
+
+        return $value !== null ? $value->getId() : null;
```
peter-gribanov: @phansys here you can see that the value after converting to database value can be `null` or `integer`. You can see similar in [`Doctrine\DBAL\Types\IntegerType`](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Types/IntegerType.php#L34).

tests/Fixtures/DoctrineType/ProductIdType.php
```diff
@@ -0,0 +1,88 @@
...
+use Doctrine\DBAL\Types\ConversionException;
+use Doctrine\DBAL\Types\Type;
+use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\ProductId;
+
+final class ProductIdType extends Type
+{
+    const NAME = 'ProductId';
+
+    /**
+     * @param array            $fieldDeclaration
```
phansys: Please, remove docblocks if they don't add any value.

tests/Fixtures/Entity/Product.php
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity;
+
+class Product
```
phansys: Could we make this class final?

tests/Fixtures/DoctrineType/ProductIdType.php
```diff
@@ -0,0 +1,88 @@
...
+     */
+    public function convertToPHPValue($value, AbstractPlatform $platform): ?ProductId
+    {
+        if ($value === null || $value instanceof ProductId) {
+            return $value;
+        }
+
+        try {
+            return new ProductId((int) $value);
+        } catch (\Exception $e) {
```
phansys: IMO, we should catch `\Throwable` here.

peter-gribanov: I change this

tests/Fixtures/Entity/Product.php
peter-gribanov: Doctrine does not allow use a final entities, but in tests we can make it as final.

tests/Fixtures/DoctrineType/ProductIdType.php
peter-gribanov: I prefer to use the extended docblocks. But for you, I deleted it. Now all is normal?

peter-gribanov:
@phansys sorry. I changed the tests a bit.

PR #938, opened 2019-10-14
updated addIdentifiersToQuery to work with custom types.

oleg-andreyev:
## Subject

Batch actions do not work with custom types.

I am targeting this branch, because bug.

## Changelog

```markdown
### Fixed
- `batchAction` now works with custom-types.
```

- [x] Update the tests

### DIFF
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -409,17 +409,16 @@ class ModelManager implements ModelManagerInterface, LockInterface
             }
 
             $fieldType = $metadata->getTypeOfField($name);
-            $type = $fieldType && Type::hasType($fieldType) ? Type::getType($fieldType) : null;
-            if ($type) {
+            if ($this->hasFieldCustomType($fieldType)) {
+                $type = $this->getFieldCustomType($fieldType);
                 $identifiers[] = $this->getValueFromType($value, $type, $fieldType, $platform);
-
                 continue;
             }
 
             $identifierMetadata = $this->getMetadata(ClassUtils::getClass($value));
 
-            foreach ($identifierMetadata->getIdentifierValues($value) as $value) {
-                $identifiers[] = $value;
+            foreach ($identifierMetadata->getIdentifierValues($value) as $identifierValue) {
+                $identifiers[] = $identifierValue;
             }
         }
 
@@ -500,17 +499,16 @@ class ModelManager implements ModelManagerInterface, LockInterface
 
         $fieldNames = $this->getIdentifierFieldNames($class);
         $qb = $query->getQueryBuilder();
+        $classMetadata = $this->getMetadata($class);
 
-        $prefix = uniqid();
+        $prefix = $this->generateIdentifierPrefix();
         $sqls = [];
         foreach ($idx as $pos => $id) {
             $ids = explode(self::ID_SEPARATOR, (string) $id);
 
             $ands = [];
-            foreach ($fieldNames as $posName => $name) {
-                $parameterName = sprintf('field_%s_%s_%d', $prefix, $name, $pos);
-                $ands[] = sprintf('%s.%s = :%s', current($qb->getRootAliases()), $name, $parameterName);
-                $qb->setParameter($parameterName, $ids[$posName]);
+            foreach ($fieldNames as $posName => $field) {
+                $ands[] = $this->buildInnerIdentifier($prefix, $field, $ids[$posName], $pos, $qb, $classMetadata);
             }
 
             $sqls[] = implode(' AND ', $ands);
@@ -766,6 +764,14 @@ class ModelManager implements ModelManagerInterface, LockInterface
         return str_replace(' ', '', ucwords(str_replace('_', ' ', $property)));
     }
 
+    /**
+     * Generating unique prefix for query.
+     */
+    protected function generateIdentifierPrefix(): string
+    {
+        return uniqid('', true);
+    }
+
     private function getFieldName(ClassMetadata $metadata, string $name): string
     {
         if (\array_key_exists($name, $metadata->fieldMappings)) {
@@ -813,5 +819,29 @@ class ModelManager implements ModelManagerInterface, LockInterface
         }
 
         return (string) $type->convertToDatabaseValue($value, $platform);
+    }
+
+    private function getFieldCustomType(string $fieldType): Type
+    {
+        return Type::getType($fieldType);
+    }
+
+    private function buildInnerIdentifier(string $prefix, string $field, $value, int $pos, QueryBuilder $qb, ClassMetadata $classMetadata): string
+    {
+        $parameterName = sprintf('field_%s_%s_%d', $prefix, $field, $pos);
+
+        $fieldMapping = $classMetadata->getFieldMapping($field);
+        if ($this->hasFieldCustomType($fieldMapping['type'])) {
+            $type = $this->getFieldCustomType($fieldMapping['type'])->getName();
+        }
+
+        $qb->setParameter($parameterName, $value, $type ?? null);
+
+        return sprintf('%s.%s = :%s', current($qb->getRootAliases()), $field, $parameterName);
+    }
+
+    private function hasFieldCustomType(?string $fieldType): bool
+    {
+        return $fieldType && Type::hasType($fieldType);
     }
 }
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -508,7 +508,8 @@ class ModelManager implements ModelManagerInterface, LockInterface
 
             $ands = [];
             foreach ($fieldNames as $posName => $field) {
-                $ands[] = $this->buildInnerIdentifier($prefix, $field, $ids[$posName], $pos, $qb, $classMetadata);
+                $parameterName = sprintf('field_%s_%s_%d', $prefix, $field, $pos);
+                $ands[] = $this->buildInnerIdentifier($field, $ids[$posName], $parameterName, $qb, $classMetadata);
             }
 
             $sqls[] = implode(' AND ', $ands);
@@ -826,10 +827,8 @@ class ModelManager implements ModelManagerInterface, LockInterface
         return Type::getType($fieldType);
     }
 
-    private function buildInnerIdentifier(string $prefix, string $field, $value, int $pos, QueryBuilder $qb, ClassMetadata $classMetadata): string
+    private function buildInnerIdentifier(string $field, $value, string $parameterName, QueryBuilder $qb, ClassMetadata $classMetadata): string
     {
-        $parameterName = sprintf('field_%s_%s_%d', $prefix, $field, $pos);
-
         $fieldMapping = $classMetadata->getFieldMapping($field);
         if ($this->hasFieldCustomType($fieldMapping['type'])) {
             $type = $this->getFieldCustomType($fieldMapping['type'])->getName();
--- a/tests/Model/ModelManagerTest.php
+++ b/tests/Model/ModelManagerTest.php
@@ -982,22 +982,21 @@ final class ModelManagerTest extends TestCase
      */
     public function testAddIdentifiersToQuery(array $expectedParameters, array $identifierFieldNames, array $ids): void
     {
-        $modelManager = $this->getMockBuilder(ModelManager::class)
-            ->disableOriginalConstructor()
-            ->setMethods(['getIdentifierFieldNames'])
-            ->getMock();
+        $managerRegistry = $this->createStub(ManagerRegistry::class);
+        $classMetadataFactory = $this->createStub(ClassMetadataFactory::class);
+        $classMetadata = $this->createStub(ClassMetadata::class);
 
-        $modelManager
-            ->expects($this->once())
-            ->method('getIdentifierFieldNames')
-            ->willReturn($identifierFieldNames);
+        $modelManager = new ModelManager(
+            $managerRegistry,
+            PropertyAccess::createPropertyAccessor()
+        );
 
-        $em = $this->getMockBuilder(EntityManager::class)
+        $entityManager = $this->getMockBuilder(EntityManager::class)
             ->disableOriginalConstructor()
             ->getMock();
 
         $queryBuilder = $this->getMockBuilder(QueryBuilder::class)
-            ->setConstructorArgs([$em])
+            ->setConstructorArgs([$entityManager])
             ->setMethodsExcept(['getParameters', 'setParameter'])
             ->getMock();
 
@@ -1013,8 +1012,34 @@ final class ModelManagerTest extends TestCase
 
         $proxyQuery = $this->getMockBuilder(ProxyQuery::class)
             ->setConstructorArgs([$queryBuilder])
-            ->setMethods(['getSortBy'])
+            ->setMethods(['getSortBy', 'getQueryBuilder'])
             ->getMock();
+
+        $proxyQuery
+            ->expects($this->once())
+            ->method('getQueryBuilder')
+            ->willReturn($queryBuilder);
+
+        $managerRegistry->method('getManagerForClass')
+            ->with(Product::class)
+            ->willReturn($entityManager);
+
+        $entityManager->method('getMetadataFactory')
+            ->willReturn($classMetadataFactory);
+
+        $classMetadataFactory->method('getMetadataFor')
+            ->with(Product::class)
+            ->willReturn($classMetadata);
+
+        $classMetadata
+            ->method('getIdentifierFieldNames')
+            ->willReturn($identifierFieldNames);
+
+        $classMetadata
+            ->method('getFieldMapping')
+            ->willReturn([
+                'type' => 'string',
+            ]);
 
         $modelManager->addIdentifiersToQuery(Product::class, $proxyQuery, $ids);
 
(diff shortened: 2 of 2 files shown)


core23:
Can you please add some tests @oleg-andreyev ?

oleg-andreyev:
@core23 planning to find some time on weekends.

core23:
Can you please add a test

oleg-andreyev:
@core23 I’ve got message for stale bot and planning to do it on weekdays. Thanks!

src/Model/ModelManager.php
```diff
@@ -629,6 +637,11 @@ protected function camelize($property)
         return str_replace(' ', '', ucwords(str_replace('_', ' ', $property)));
     }
 
+    protected function generatePrefix(): string
```
core23: Can you make this private

oleg-andreyev: I can't. I've extracted it into a separate method for testability, see https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/938/files#diff-c50f6d3b29e91b97ea76304d137c9929R1006

src/Model/ModelManager.php
```diff
@@ -414,7 +415,14 @@ public function addIdentifiersToQuery($class, ProxyQueryInterface $queryProxy, a
             foreach ($fieldNames as $posName => $name) {
                 $parameterName = sprintf('field_%s_%s_%d', $prefix, $name, $pos);
                 $ands[] = sprintf('%s.%s = :%s', current($qb->getRootAliases()), $name, $parameterName);
-                $qb->setParameter($parameterName, $ids[$posName]);
+
```
core23: Can you please try to extract some small methods. This will improve readability.

VincentLanglet:
Hi @oleg-andreyev ! Do you have time to finish this PR ? This seems like almost done ! :)

src/Model/ModelManager.php
oleg-andreyev: done

oleg-andreyev:
Failing tests aren't related to current PR, can someone please trigger travis manually?

VincentLanglet:
@oleg-andreyev Done

oleg-andreyev:
@core23 please review PR

src/Model/ModelManager.php
```diff
@@ -666,4 +671,20 @@ private function getValueFromType($value, Type $type, string $fieldType, Abstrac
 
         return (string) $type->convertToDatabaseValue($value, $platform);
     }
+
+    private function getFieldCustomType(?string $fieldType): ?Type
+    {
+        return Type::hasType($fieldType) ? Type::getType($fieldType) : null;
+    }
+
+    private function buildInnerIdentifier(string $prefix, string $field, $value, int $pos, QueryBuilder $qb, ClassMetadata $classMetadata)
```
phansys, line 680:
```suggestion
    private function buildInnerIdentifier(string $prefix, string $field, $value, int $pos, QueryBuilder $qb, ClassMetadata $classMetadata): string
```

VincentLanglet:
@core23 @phansys Can you take a new look at the PR ? :)

src/Model/ModelManager.php
```diff
@@ -404,23 +404,23 @@ public function addIdentifiersToQuery($class, ProxyQueryInterface $queryProxy, a
     {
         $fieldNames = $this->getIdentifierFieldNames($class);
         $qb = $queryProxy->getQueryBuilder();
+        /** @var ClassMetadata $classMetadata */
```
phansys, line 407: `getMetadata()` has already declared a return type in its `@return` type hint. Why we need this inline docblock?
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/babf4865cc6c4ba8af8ef8b21df8118401566605/src/Model/ModelManager.php#L59-L64
```suggestion
```

src/Model/ModelManager.php
```diff
@@ -631,6 +631,11 @@ protected function camelize($property)
         return str_replace(' ', '', ucwords(str_replace('_', ' ', $property)));
     }
 
+    protected function generatePrefix(): string
```
phansys, line 634: This name seems very generic in the context of a model manager. What is the purpose for the generated prefix?
IMO, we should improve the name and add a description if it is not clear enough.

src/Model/ModelManager.php
```diff
@@ -666,4 +671,20 @@ private function getValueFromType($value, Type $type, string $fieldType, Abstrac
 
         return (string) $type->convertToDatabaseValue($value, $platform);
     }
+
+    private function getFieldCustomType(?string $fieldType): ?Type
+    {
+        return Type::hasType($fieldType) ? Type::getType($fieldType) : null;
```
phansys, line 677: Since the argument 1 for `Type::hasType()` is declared as `string`, we should not pass `null` values.

src/Model/ModelManager.php
```diff
@@ -666,4 +671,20 @@ private function getValueFromType($value, Type $type, string $fieldType, Abstrac
 
         return (string) $type->convertToDatabaseValue($value, $platform);
     }
+
+    private function getFieldCustomType(?string $fieldType): ?Type
```
phansys, line 675: Even if this method is not part of the API, I think we should be consistent with the same principles we're enforcing in other packages about avoid returning `null` on getters when there is no value associated to the provided argument (see sonata-project/SonataAdminBundle#6050).

src/Model/ModelManager.php
```diff
@@ -666,4 +671,20 @@ private function getValueFromType($value, Type $type, string $fieldType, Abstrac
...
+        return Type::hasType($fieldType) ? Type::getType($fieldType) : null;
+    }
+
+    private function buildInnerIdentifier(string $prefix, string $field, $value, int $pos, QueryBuilder $qb, ClassMetadata $classMetadata): string
+    {
+        $parameterName = sprintf('field_%s_%s_%d', $prefix, $field, $pos);
+
+        $fieldMapping = $classMetadata->getFieldMapping($field);
+        $type = $this->getFieldCustomType($fieldMapping['type']);
+        $qb->setParameter($parameterName, $value, $type ? $type->getName() : null);
```
phansys, line 686:
```suggestion
        $qb->setParameter($parameterName, $value, null !== $type ? $type->getName() : null);
```

tests/Model/ModelManagerTest.php
```diff
@@ -966,6 +966,57 @@ public function testGetUrlsafeIdentifierNull(): void
         $this->assertNull($model->getNormalizedIdentifier(null));
     }
 
+    public function testAddIdentifiersToQueryWithCustomIdMapping(): void
+    {
+        $managerRegistry = $this->createMock(ManagerRegistry::class);
```
phansys, line 971:
```suggestion
        $managerRegistry = $this->createStub(ManagerRegistry::class);
```

tests/Model/ModelManagerTest.php
```diff
@@ -966,6 +966,57 @@ public function testGetUrlsafeIdentifierNull(): void
...
+        $entityManager = $this->createMock(EntityManagerInterface::class);
```
phansys, line 972:
```suggestion
        $entityManager = $this->createStub(EntityManagerInterface::class);
```

tests/Model/ModelManagerTest.php
```diff
@@ -966,6 +966,57 @@ public function testGetUrlsafeIdentifierNull(): void
...
+        $classMetadataFactory = $this->createMock(ClassMetadataFactory::class);
```
phansys, line 973:
```suggestion
        $classMetadataFactory = $this->createStub(ClassMetadataFactory::class);
```

tests/Model/ModelManagerTest.php
```diff
@@ -966,6 +966,57 @@ public function testGetUrlsafeIdentifierNull(): void
...
+        $proxyQuery = $this->createMock(ProxyQuery::class);
```
phansys, line 974:
```suggestion
        $proxyQuery = $this->createStub(ProxyQuery::class);
```

tests/Model/ModelManagerTest.php
```diff
@@ -966,6 +966,57 @@ public function testGetUrlsafeIdentifierNull(): void
...
+        $queryBuilder = new QueryBuilder($entityManager);
+        $classMetadata = $this->createMock(ClassMetadata::class);
```
phansys, line 976:
```suggestion
        $classMetadata = $this->createStub(ClassMetadata::class);
```

src/Model/ModelManager.php
oleg-andreyev: removed

src/Model/ModelManager.php
oleg-andreyev: before it was hardcoded as `uniqid`, so I've extracted it into the separate method for testability

src/Model/ModelManager.php
oleg-andreyev: it's generated based on the return type of `\Doctrine\ORM\Mapping\ClassMetadataInfo::getTypeOfField` as of today it's `@return string|null`

src/Model/ModelManager.php
oleg-andreyev: done

tests/Model/ModelManagerTest.php
oleg-andreyev: actually `createStub` is just a proxy for `createMock`.

phansys: Yes, that's the current situation. But in order to make these chages forward compatible, we should use `createStub()` if there isn't any expectation about the class methods.
Please, see https://github.com/sebastianbergmann/phpunit/issues/3120#issue-322494156.

oleg-andreyev: Make sense, so I've updated it in last commit

src/Model/ModelManager.php
phansys: I mean, we should respect the signature exposed by `Type::hasType()`.
This way is how it looks like at [`doctrine/dbal:2.10.2`](https://github.com/doctrine/dbal/blob/2.10.2/lib/Doctrine/DBAL/Types/Type.php#L249-L259):
```php
    /**
     * Checks if exists support for a type.
     *
     * @param string $name The name of the type.
     *
     * @return bool TRUE if type is supported; FALSE otherwise.
     */
    public static function hasType($name)
    {
        return self::getTypeRegistry()->has($name);
    }
```

phansys (dismissed):

src/Model/ModelManager.php
core23: 👎 For adding an extension point just for tests. You could use a better assert to verify this.

jordisala1991:
What is the status of this Pr?

VincentLanglet:
Are you still interested by this PR @oleg-andreyev ?

oleg-andreyev:
Can I'll rebase this branch, but it's taking too long to merge, so I completely forgot about this PR.

VincentLanglet:
> I'll rebase this branch, but it's taking too long to merge, so I completely forgot about this PR.

There is few contirbutors, but lot of work/issues/PR.
Don't hesitate to ping us to friendly remember us to take a review.

core23 (approved):

tests/Model/ModelManagerTest.php
```diff
@@ -982,22 +982,21 @@ public function testGetUrlsafeIdentifierNull(): void
...
-            ->getMock();
+        $managerRegistry = $this->createStub(ManagerRegistry::class);
+        $classMetadataFactory = $this->createStub(ClassMetadataFactory::class);
+        $classMetadata = $this->createStub(ClassMetadata::class);
 
-        $modelManager
-            ->expects($this->once())
-            ->method('getIdentifierFieldNames')
-            ->willReturn($identifierFieldNames);
+        $modelManager = new ModelManager(
```
oleg-andreyev, line 989: mocking target is a bad practice.

tests/Model/ModelManagerTest.php
```diff
@@ -982,22 +982,21 @@ public function testGetUrlsafeIdentifierNull(): void
      */
     public function testAddIdentifiersToQuery(array $expectedParameters, array $identifierFieldNames, array $ids): void
     {
-        $modelManager = $this->getMockBuilder(ModelManager::class)
```
oleg-andreyev, line 985: mocking target is a bad practice.

oleg-andreyev:
@OskarStark @core23 @phansys I've rebased branch on latest 3.x, please review.

src/Model/ModelManager.php
```diff
@@ -814,4 +820,30 @@ private function getValueFromType($value, Type $type, string $fieldType, Abstrac
 
         return (string) $type->convertToDatabaseValue($value, $platform);
     }
+
+    private function getFieldCustomType(string $fieldType): Type
+    {
+        return Type::getType($fieldType);
+    }
+
+    private function buildInnerIdentifier(string $prefix, string $field, $value, int $pos, QueryBuilder $qb, ClassMetadata $classMetadata): string
```
VincentLanglet, line 829: There is a lot of parameters in this private method.

Is there a way to reduce the number ?
Maybe the parameterName can be an argument instead of passing the prefix and the pos.

src/Model/ModelManager.php
```diff
@@ -409,17 +409,16 @@ public function getIdentifierValues($entity)
             }
 
             $fieldType = $metadata->getTypeOfField($name);
-            $type = $fieldType && Type::hasType($fieldType) ? Type::getType($fieldType) : null;
-            if ($type) {
+            if ($this->hasFieldCustomType($fieldType)) {
+                $type = $this->getFieldCustomType($fieldType);
```
VincentLanglet, line 413: When I look at the logic inside these function, I'm not sure we need to add private method for this.

src/Model/ModelManager.php
```diff
@@ -814,4 +820,30 @@ private function getValueFromType($value, Type $type, string $fieldType, Abstrac
...
+        $parameterName = sprintf('field_%s_%s_%d', $prefix, $field, $pos);
+
+        $fieldMapping = $classMetadata->getFieldMapping($field);
+        $type = null;
+
+        if ($this->hasFieldCustomType($fieldMapping['type'])) {
+            $type = $this->getFieldCustomType($fieldMapping['type']);
+        }
+
+        $qb->setParameter($parameterName, $value, null !== $type ? $type->getName() : null);
```
VincentLanglet, line 840:
```suggestion
        if ($this->hasFieldCustomType($fieldMapping['type'])) {
            $type = $this->getFieldCustomType($fieldMapping['type'])->getName();
        }

        $qb->setParameter($parameterName, $value, $type ?? null);
```

src/Model/ModelManager.php
```diff
@@ -500,17 +499,16 @@ public function addIdentifiersToQuery($class, ProxyQueryInterface $query, array
 
         $fieldNames = $this->getIdentifierFieldNames($class);
         $qb = $query->getQueryBuilder();
+        $classMetadata = $this->getMetadata($class);
 
-        $prefix = uniqid();
+        $prefix = $this->generateIdentifierPrefix();
```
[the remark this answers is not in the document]
oleg-andreyev: to have an ability to override it in tests, otherwise, it's impossible to predict results with hardcoded `unique`

src/Model/ModelManager.php
oleg-andreyev: It's used in two places. I usually try to remove duplicates. This reduces the burden on the developers when they need to update "Type" class usage.

src/Model/ModelManager.php
oleg-andreyev: LGTM

src/Model/ModelManager.php
oleg-andreyev: Initially wanted to make `addIdentifiersToQuery` a bit cleaner and remove chunk of code from foreach-loop but other hand I can revert this private method

src/Model/ModelManager.php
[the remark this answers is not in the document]
VincentLanglet: We shouldn't add method for this.
You could check the name by using a regex instead.

src/Model/ModelManager.php
VincentLanglet: Yes, I saw the duplicate.

But it's basically like a method 
```
private function foo($a) {
    return $this->bar($a);
}
```

oleg-andreyev: Right. Similar to findBy and findOneBy https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/560e9f039d63cf2732e4f6c428549d9ac67826c5/src/Model/ModelManager.php#L285-L293

src/Model/ModelManager.php
VincentLanglet: The method `addIdentifiersToQuery` is complex, so this private method has a lot of sens. 

```
$parameterName = sprintf('field_%s_%s_%d', $prefix, $name, $pos);
$this-> buildInnerIdentifier($field, $value, $paraterName, $qb, $classMetadata)
```
Was just a suggestion, not a blocking point.

src/Model/ModelManager.php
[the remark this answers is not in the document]
oleg-andreyev: One good practice from DDD - avoid having dependencies on I/O outside of the Infrastructure layer, so I tend to remove such dependencies into a separate method/class so that it could be replaced later without affecting main functionality and tests. What do you think about it?

src/Model/ModelManager.php
oleg-andreyev: It will reduce a few args. Good idea. Will do it.

src/Model/ModelManager.php
VincentLanglet: But these one are public, and required by the interface.

src/Model/ModelManager.php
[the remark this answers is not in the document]
OskarStark: I like the idea 💡

oleg-andreyev:
@core23 @OskarStark @phansys can you please take a look?

src/Model/ModelManager.php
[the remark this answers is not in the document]
VincentLanglet: @oleg-andreyev In Sonata 4.0, the ModelManager will be final.
- So a protected method is useless, it's basically a private one.
- And creating a private method for calling `uniqid()` seems useless to me.

VincentLanglet (changes requested):

PR #950, opened 2019-11-19
allow DoctrineBundle 2.0

### LINKED ISSUE #947: Support doctrine/doctrine-bundle 2.0?
## Feature Request

Since `doctrine/doctrine-bundle` 2.0 will be released soon: Are there any plans supporting it?

https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/composer.json#L27
https://github.com/doctrine/DoctrineBundle/releases/tag/2.0.0-Beta1

dmaicher:
## Subject

I am targeting this branch, because this is not be a BC break?

Closes #947 

## Changelog

```markdown
### Added
- Added support for Doctrine-Bundle 2.0
```

### DIFF
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\DoctrineORMAdminBundle\Model;
 
+use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\Common\Util\ClassUtils;
 use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\LockMode;
@@ -35,14 +36,13 @@ use Sonata\DoctrineORMAdminBundle\Admin\FieldDescription;
 use Sonata\DoctrineORMAdminBundle\Datagrid\OrderByToSelectWalker;
 use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery;
 use Sonata\Exporter\Source\DoctrineORMQuerySourceIterator;
-use Symfony\Bridge\Doctrine\RegistryInterface;
 use Symfony\Component\Form\Exception\PropertyAccessDeniedException;
 
 class ModelManager implements ModelManagerInterface, LockInterface
 {
     public const ID_SEPARATOR = '~';
     /**
-     * @var RegistryInterface
+     * @var ManagerRegistry
      */
     protected $registry;
 
@@ -51,7 +51,7 @@ class ModelManager implements ModelManagerInterface, LockInterface
      */
     protected $cache = [];
 
-    public function __construct(RegistryInterface $registry)
+    public function __construct(ManagerRegistry $registry)
     {
         $this->registry = $registry;
     }
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     ],
     "require": {
         "php": "^7.1",
-        "doctrine/doctrine-bundle": "^1.8",
+        "doctrine/doctrine-bundle": "^1.8 || ^2.0",
         "doctrine/orm": "^2.5",
         "sonata-project/admin-bundle": "^3.29",
         "sonata-project/core-bundle": "^3.14",
--- a/tests/Model/ModelManagerTest.php
+++ b/tests/Model/ModelManagerTest.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Sonata\DoctrineORMAdminBundle\Tests\Model;
 
 use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\Common\Persistence\ObjectManager;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DBALException;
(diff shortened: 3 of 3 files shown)


src/Model/ModelManager.php
```diff
@@ -51,7 +51,7 @@ class ModelManager implements ModelManagerInterface, LockInterface
      */
     protected $cache = [];
 
-    public function __construct(RegistryInterface $registry)
+    public function __construct(ManagerRegistry $registry)
```
dmaicher: this was the only issue I noticed:

https://github.com/doctrine/DoctrineBundle/blob/master/UPGRADE-2.0.md#registry

core23: This will work for both major versions?

dmaicher: Yes. I tested it with 1.10 and 1.11 as well.

greg0ire: This is a BC-break. The type declaration should be dropped, and a deprecation triggered to tell people they have to pass an instance of `ManagerRegistry` if they didn't. Also, a `NEXT_MAJOR` note should be added to remind us to add a type declaration on the master branch.

dmaicher: Are you sure this is a BC break?

I think all instances of `RegistryInterface` are actually instance of `ManagerRegistry`?

dmaicher: The change is this one:

https://github.com/doctrine/DoctrineBundle/commit/43a1929fb193ce166affabfb1d7ec2763594b2ac

Which extends this one:

https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Doctrine/ManagerRegistry.php#L23

greg0ire: @dmaicher the user could write a decorator for the manager registry, so technically it's a BC break, yeah.

dmaicher: Ok indeed :+1: this case would break.

dmaicher: But wait. Would it really break?

If I have something like this:

```php
class MyManager extends ModelManager
{
    public function __construct(RegistryInterface $registry)
    {
        parent::__construct($registry);
    }
}
```

It will still work fine since we made it more loose on the parent class.

`RegistryInterface` actually extends `ManagerRegistry`.

Or am I missing something?

eugenekurasov: You can do decorator for `RegistryInterface`, or you can implement your `RegistryInterface`.

dmaicher: Which still means I have an instance of `ManagerRegistry` then?

https://github.com/symfony/symfony/blob/4.4/src/Symfony/Bridge/Doctrine/RegistryInterface.php#L24

eugenekurasov: You mean to change for this -https://github.com/doctrine/persistence/blob/master/lib/Doctrine/Persistence/ManagerRegistry.php ?

dmaicher: I fail to see how the change is a BC break. Maybe you have an example?

eugenekurasov: https://symfony.com/doc/current/service_container/service_decoration.html
You can decorate your interface.
But interesting why is not any tests is failed?

eugenekurasov: Your change maybe don't throw exception, because `RegistryInterface` has been extend from `ManagerRegistry`, but if someone decorate `RegistryInterface` they can be have unexpected behavior.

dmaicher:
@core23 @greg0ire I adjusted the tests so we cover both cases

eugenekurasov:
Why is coverage decreased?

dmaicher:
> Why is coverage decreased?

No idea. Seems wrong to me this change.

core23:
> > Why is coverage decreased?
> 
> No idea. Seems wrong to me this change.

This is a known problem, because each PR shares the same coverage database

src/Model/ModelManager.php
greg0ire: Service decoration is exactly what I had in mind. The decorator would not extend the concrete class. In fact, it wouldn't even know about it, but only about the interface.

eugenekurasov: Remove type declaration or change to parent class, will be have 'little' break BC. I don't have idea how it can do without break BC.

eugenekurasov: Maybe to add deprecation triggered on `RegistryInterface`? I think is can help to detected `unexpected behavior `? @greg0ire What you think about it?

greg0ire: Removing the type declaration is absolutely not a BC break since this is a constructor. I think I have been clear on what needs to be done in my first message

Metabor:
I am really looking forward for this change, to use Sonata Admin in a default Symfony 4.4 setup (which uses DoctrineBundle 2).

src/Model/ModelManager.php
dmaicher: @greg0ire I still fail to see how this can be a BC break.

The now deprecated `RegistryInterface` extends Doctrine's `ManagerRegistry` interface:

https://github.com/symfony/symfony/blob/4.4/src/Symfony/Bridge/Doctrine/RegistryInterface.php#L24

So this will work even for any custom implementation of `RegistryInterface` as the tests show.

greg0ire: Oh right, I didn't have that piece of the puzzle (I'm at the SfCon and that kind of inheritance is hard to check with just a phone). Everything should be fine indeed :)

OskarStark:
Thanks David

I will request a release internally

PR #958, opened 2019-12-03
Disable validation groups on datagrid filter form

patrick-benoit:
## Subject

Datagrid filters form data is passed as array, so it is validated with the Valid constraint (see https://symfony.com/doc/3.4/form/without_class.html#adding-validation).
Some errors related to linked objets, but unrelated to the form, can then appear in the filters block.

### Example
On a child admin datagrid, the parent id is stored in a hidden field. The parent object is validated while building the form and if it has a Callback validator, errors might be displayed in the filters block.

image

I am targeting this branch, because it's a patch without backward compatibility changes.

## Changelog

```markdown
### Fixed
Disabled validation group in Builder/DatagridBuilder::getBaseDatagrid
```


core23:
Thanks @patrick-benoit

PR #960, opened 2019-12-06
Fix sonata-project/SonataAdminBundle#5781

zyberspace:
## Fix sonata-project/SonataAdminBundle#5781

Hi, this fixes sonata-project/SonataAdminBundle#5781 by making FilterTypeGuesser return the FQCN of the filters instead of their aliases. The code that added the aliases was [removed](https://github.com/sonata-project/SonataAdminBundle/commit/51201320bbb53431acc4b303b1735ee1474ea3c5#diff-e7680a7f9961b136663cf182b119e0c5L38) in sonata admin v3.56.0.

*Note:* There is also a PR for the sonata admin bundle to re-add the removed code: sonata-project/SonataAdminBundle#5782

I am targeting this branch, because this is a backwards compatible bugfix.

## Changelog

```markdown
### Fixed
- Fixed sonata-project/SonataAdminBundle#5781 by making FilterTypeGuesser return the FQCN of the filters instead of their aliases because it isn't supported by sonata admin anymore.
```

- [x] Update the tests *(Should i also add a new test for this?)*
- [ ] Update the documentation *(`docs/reference/filter_field_definition.rst` still contains the old aliases, maybe this should be upgraded all together?)*
- [ ] Add an upgrade note

### DIFF
--- a/src/Guesser/FilterTypeGuesser.php
+++ b/src/Guesser/FilterTypeGuesser.php
@@ -15,6 +15,13 @@ namespace Sonata\DoctrineORMAdminBundle\Guesser;
 
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Sonata\AdminBundle\Model\ModelManagerInterface;
+use Sonata\DoctrineORMAdminBundle\Filter\BooleanFilter;
+use Sonata\DoctrineORMAdminBundle\Filter\DateFilter;
+use Sonata\DoctrineORMAdminBundle\Filter\DateTimeFilter;
+use Sonata\DoctrineORMAdminBundle\Filter\ModelFilter;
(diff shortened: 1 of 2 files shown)


greg0ire:
> `docs/reference/filter_field_definition.rst` still contains the old aliases, maybe this should be upgraded all together?)

Please remove them indeed, the less documentation we have on "wrong" things the better. You should do this in a separate commit, or even in a separate PR, that's up to you.

zyberspace:
Looks like this was now fixed in 0d67e6f00ec3 by @OskarStark. This means this PR can be closed, right?

greg0ire:
You're correct, I'm afraid! Sorry for not spotting that before, and thanks a lot for trying to fix this!

PR #965, opened 2020-01-09
Streamline deprecation messages

OskarStark:


src/Datagrid/Pager.php
```diff
@@ -26,7 +26,7 @@ class Pager extends BasePager
     /**
      * NEXT_MAJOR: remove this property.
      *
-     * @deprecated This property is deprecated since version 2.4 and will be removed in 3.0
+     * @deprecated since sonata-project/doctrine-orm-admin-bundle 2.4 and will be removed in 3.0
```
OskarStark: Btw if we keep this old deprecation, in 3.x it gets removed in 4.0, not 3.0, right?

core23: We will remove this with 4.0. You can change this

OskarStark:
```suggestion
     * @deprecated since sonata-project/doctrine-orm-admin-bundle 2.4 and will be removed in 4.0
```

PR #978, opened 2020-02-07
Added configuration for case_sensitive option in StringFilter

### LINKED ISSUE #334: proposal: case insensitive search on postgres
for now, we do this:
https://gist.github.com/dbu/9524776#file-caseinsensitivestringfilter-php-L40

doctrine_orm_filter_types.xml add an option to configure the default
behaviour for sonata.admin.orm.filter.type.string
additionally, we could add an option to the filter to set it per case
(who knows, maybe i have one field where i really want case sensitive
filtering).

btw, one can achieve the opposite, make mysql case sensitive for likes
with  "column COLLATE utf8_bin LIKE :value" (resp = or NOT LIKE)

not sure how this is best modeled. if we don't care about BC, i would
say postgres should by default start to compare case insensitive. but
this would be very surprising for existing applications that rely on the
behaviour.

having 2 more ChoiceType options sounds like a bad idea too, it will
overkill for the admin user most of the time.

interested in a PR to add an option to the DI configuration to explicitly set case sensitivity (with 3 states: default, case-sensitive, case-insensitive), plus a per-filter option.

i don't know about other db than postgres + mysql, so i would throw an exception that this is not supported if an explicit option is set, so that people at least know something is not working.

Nadrshin:
## Subject

I am targeting this branch, because I need an opportunity to set a case-insensitive search for the StringFilter through configuration.

Closes #334

## Changelog

```markdown
### Added
- Added configuration for `case_sensitive` option in `StringFilter`
```

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\DoctrineORMAdminBundle\DependencyInjection;
 
+use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
 
@@ -41,13 +42,6 @@ class Configuration implements ConfigurationInterface
         } else {
             $rootNode = $treeBuilder->getRootNode();
         }
-
-        $caseSensitiveInfo = <<<'CASESENSITIVE'
-Whether the StringFilter should behave case sensitive or not.
-Using case-insensitivity might lead to performance issues.
-See https://use-the-index-luke.com/sql/where-clause/functions/case-insensitive-search
-for more information.
-CASESENSITIVE;
 
         $rootNode
             ->children()
@@ -83,23 +77,40 @@ CASESENSITIVE;
                         ->end()
                     ->end()
                 ->end()
-                ->arrayNode('filters')
+            ->end()
+            ->append($this->getFiltersNode())
+        ;
+
+        return $treeBuilder;
+    }
+
+    private function getFiltersNode(): ArrayNodeDefinition
+    {
+        $treeBuilder = new TreeBuilder();
+        $node = $treeBuilder->root('filters');
+
+        $caseSensitiveInfo = <<<'CASESENSITIVE'
+Whether the StringFilter should behave case sensitive or not.
+Using case-insensitivity might lead to performance issues.
+See https://use-the-index-luke.com/sql/where-clause/functions/case-insensitive-search
+for more information.
+CASESENSITIVE;
+
+        $node
+            ->addDefaultsIfNotSet()
+            ->children()
+                ->arrayNode('string')
                     ->addDefaultsIfNotSet()
                     ->children()
-                        ->arrayNode('string')
-                            ->addDefaultsIfNotSet()
-                            ->children()
-                                ->booleanNode('case_sensitive')
-                                    ->defaultTrue()
-                                    ->info($caseSensitiveInfo)
-                                ->end()
-                            ->end()
+                        ->booleanNode('case_sensitive')
+                            ->defaultTrue()
+                            ->info($caseSensitiveInfo)
                         ->end()
                     ->end()
                 ->end()
             ->end()
         ;
 
-        return $treeBuilder;
+        return $node;
     }
 }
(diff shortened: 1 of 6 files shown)


src/DependencyInjection/Configuration.php
```diff
@@ -76,6 +83,20 @@ public function getConfigTreeBuilder()
                         ->end()
                     ->end()
                 ->end()
+                ->arrayNode('filters')
```
core23: Can you please move this to a new method

src/Filter/StringFilter.php
```diff
@@ -24,6 +24,19 @@ class StringFilter extends Filter
...
 
+    /**
+     * @var bool
+     */
+    private $caseSensitive;
+
+    /**
+     * @param bool $caseSensitive
+     */
+    public function __construct($caseSensitive)
```
core23:
```suggestion
    public function __construct(bool $caseSensitive = false)
```

src/DependencyInjection/Configuration.php
Nadrshin: I added method `getFiltersNode`.

src/Filter/StringFilter.php
Nadrshin: Done. But I used `true` as default value.

Nadrshin:
@core23, I provided unit tests.

It is my first PR in github. Therefore, there may be problems.

PS: Maybe I should rename the variable `StringFilter::$caseSensitive` to `StringFilter::$defaultCaseSensitive`?

core23:
> @core23, I provided unit tests.
> 
> It is my first PR in github. 

Welcome to the family :)

> 
> PS: Maybe I should rename the variable `StringFilter::$caseSensitive` to `StringFilter::$defaultCaseSensitive`?

Good catch! This would be a better name.

src/Filter/StringFilter.php
```diff
@@ -24,6 +24,16 @@ class StringFilter extends Filter
         ChoiceType::TYPE_EQUAL => '=',
     ];
 
+    /**
+     * @var bool
+     */
+    private $defaultCaseSensitive;
+
+    public function __construct(bool $defaultCaseSensitive = true)
```
OskarStark: If someone sets it it’s not default anymore.
I would go without the default prefix

core23: It's just to keep BC for users who extend this class

VincentLanglet:
The tests fail weirdly. You should try to commit amend and force push to re-trigger the run.

src/DependencyInjection/Configuration.php
```diff
@@ -77,8 +78,39 @@ public function getConfigTreeBuilder()
...
+            ->append($this->getFiltersNode())
         ;
 
         return $treeBuilder;
     }
+
+    private function getFiltersNode(): ArrayNodeDefinition
+    {
+        $treeBuilder = new TreeBuilder();
+        $node = $treeBuilder->root('filters');
```
franmomu:
```suggestion
        $treeBuilder = new TreeBuilder('filters');
        $node = $treeBuilder->getRootNode();
```

greg0ire:
@VincentLanglet it depends on the RDBMS. On Postgres, case sensitivity is the default IIRC.

VincentLanglet:
> @VincentLanglet it depends on the RDBMS. On Postgres, case sensitivity is the default IIRC.

IMO, It should not rely on this. It is weird having a `case sensitive` option if it's overriden by the database.

Shouldn't we add something like
```
if ($this->getOption('case_sensitive')) {
    $or->add(sprintf('BINARY %s.%s %s :%s', $alias, $field, $operator, $parameterName));
} else {
    $or->add(sprintf('BINARY LOWER(%s.%s) %s :%s', $alias, $field, $operator, $parameterName));
}
```

greg0ire:
I was going to argue that functions make indices unusable and should be avoided, but using `LIKE` with '%' at the beginning of a pattern do that too… is it what's happening in your case? What happens when you search for `EST`? Does `Testa` come up? If yes, we're probably not using indices already.

VincentLanglet:
> I was going to argue that functions make indices unusable and should be avoided, but using `LIKE` with '%' at the beginning of a pattern do that too… is it what's happening in your case? What happens when you search for `EST`? Does `Testa` come up? If yes, we're probably not using indices already.

image

greg0ire:
Ok well we might as well use functions everywhere then, since performance is already ruined :sweat_smile:

VincentLanglet:
The fix would be better with this feature (in this PR or in another), because I consider this like a bug, but it can be considered like a breaking chance for sql database users.
Giving an option to override easily the behaviour could be great.

@Nadrshin Do you have time to finish your PR ? :)

Nadrshin:
Sorry, I don't have time.

PR #979, opened 2020-02-07
Add new filter for empty fields

core23:
## Subject

This filter allows filtering for (non-)nullable fields.

## Changelog

```markdown
### Added
-  Add new filter for empty fields
```

### DIFF
--- a/docs/reference/filter_field_definition.rst
+++ b/docs/reference/filter_field_definition.rst
@@ -44,6 +44,7 @@ For now, only `Doctrine ORM` filters are available:
 * ``Sonata\DoctrineORMAdminBundle\Filter\DateTimeFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\DateTimeType`` Form Type, renders a datetime field,
 * ``Sonata\DoctrineORMAdminBundle\Filter\DateTimeRangeFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\DateTimeRangeType`` Form Type, renders a 2 datetime fields,
 * ``Sonata\DoctrineORMAdminBundle\Filter\ClassFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\DefaultType`` Form type, renders a choice list field.
+* ``Sonata\DoctrineORMAdminBundle\Filter\EmptyFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\DefaultType`` Form type, renders a choice list field.
 
 Example
 -------
@@ -159,6 +160,26 @@ ClassFilter
         {
             $datagridMapper
                 ->add('type', ClassFilter::class, ['sub_classes' => $this->getSubClasses()]);
+        }
+    }
+
+Empty
+-----
+
+``Sonata\DoctrineORMAdminBundle\Filter\EmptyFilter`` supports filtering for empty (null) entity fields::
+
+    namespace Sonata\NewsBundle\Admin;
+
+    use Sonata\AdminBundle\Admin\AbstractAdmin;
+    use Sonata\AdminBundle\Datagrid\DatagridMapper;
+    use Sonata\AdminBundle\Filter\EmptyFilter;
+
+    final class PostAdmin extends AbstractAdmin
+    {
+        protected function configureDatagridFilters(DatagridMapper $datagridMapper)
+        {
+            $datagridMapper
+                ->add('deleted', EmptyFilter::class, ['field_name' => 'deletedAt']);
         }
     }
 
--- a/src/Filter/EmptyFilter.php
+++ b/src/Filter/EmptyFilter.php
@@ -1,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DoctrineORMAdminBundle\Filter;
+
+use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
+use Sonata\AdminBundle\Form\Type\Filter\DefaultType;
+use Sonata\Form\Type\BooleanType;
+use Symfony\Component\Form\Extension\Core\Type\HiddenType;
+
+final class EmptyFilter extends Filter
+{
+    /**
+     * @param string       $alias
+     * @param string       $field
+     * @param mixed[]|null $data
+     */
+    public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data): void
+    {
+        if (null === $data || !\is_array($data) || !\array_key_exists('value', $data)) {
+            return;
+        }
+
+        if (BooleanType::TYPE_YES === (int) $data['value']) {
+            $this->applyWhere(
+                $queryBuilder,
+                $queryBuilder
+                ->expr()
+                ->isNull(sprintf('%s.%s', $alias, $field))
+            );
+        } elseif (BooleanType::TYPE_NO === (int) $data['value']) {
+            $this->applyWhere(
+                $queryBuilder,
+                $queryBuilder
+                ->expr()
+                ->isNotNull(sprintf('%s.%s', $alias, $field))
+            );
+        }
+    }
+
+    public function getDefaultOptions()
+    {
+        return [
+            'field_type' => BooleanType::class,
+            'operator_type' => HiddenType::class,
+            'operator_options' => [],
+        ];
+    }
+
+    public function getRenderSettings()
+    {
+        return [DefaultType::class, [
+            'field_type' => $this->getFieldType(),
+            'field_options' => $this->getFieldOptions(),
+            'operator_type' => $this->getOption('operator_type'),
+            'operator_options' => $this->getOption('operator_options'),
+            'label' => $this->getLabel(),
+        ]];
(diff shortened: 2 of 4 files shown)


src/Filter/EmptyFieldFilter.php
```diff
@@ -0,0 +1,105 @@
...
+use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
+use Sonata\AdminBundle\Form\Type\Filter\DefaultType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+
+final class EmptyFieldFilter extends Filter
+{
+    /**
+     * Field is set.
+     */
+    public const TYPE_YES = 1;
```
OskarStark: Couldn’t we use our own constants instead of yes and now with a separate meaning (phpdoc)?

docs/reference/filter_field_definition.rst
```diff
@@ -44,6 +44,7 @@ For now, only `Doctrine ORM` filters are available:
 * ``Sonata\DoctrineORMAdminBundle\Filter\DateTimeFilter``: depends on the ``sonata_type_filter_datetime`` Form Type, renders a datetime field,
 * ``Sonata\DoctrineORMAdminBundle\Filter\DateTimeRangeFilter``: depends on the ``sonata_type_filter_datetime_range`` Form Type, renders a 2 datetime fields,
 * ``Sonata\DoctrineORMAdminBundle\Filter\ClassFilter``: depends on the ``sonata_type_filter_default`` Form type, renders a choice list field.
+* ``Sonata\DoctrineORMAdminBundle\Filter\EmptyFieldFilter``: depends on the ``sonata_type_filter_default`` Form type, renders a choice list field.
```
franmomu:
```suggestion
* ``Sonata\DoctrineORMAdminBundle\Filter\EmptyFieldFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\DefaultType`` Form type, renders a choice list field.
```

src/Filter/EmptyFieldFilter.php
```diff
@@ -0,0 +1,104 @@
...
+            'choices' => [
+                'label_type_empty' => static::TYPE_EMPTY,
+                'label_type_not_empty' => static::TYPE_NOT_EMPTY,
+            ],
+        ]);
+    }
+
+    public function getRenderSettings()
+    {
+        return [DefaultType::class, [
```
VincentLanglet: I think you should create a SonataAdminBundle/Form/Type.
For example the `NumberFilter` use a `NumberType`.

Since there is multiple ORM Repository, this is a way to use same labels, constants, configuration, ...

src/Filter/EmptyFieldFilter.php
```diff
@@ -0,0 +1,104 @@
...
+     * @param array|string|null $data
+     */
+    public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data): void
+    {
+        if (null === $data || !\is_array($data) || !\array_key_exists('value', $data)) {
+            return;
+        }
+
+        if ((int) $data['value'] === static::TYPE_NOT_EMPTY) {
+            $queryBuilder->andWhere(sprintf('%s.%s IS null', $alias, $field));
```
VincentLanglet: If I filter by TYPE_NOT_EMPTY, I expect the not null

VincentLanglet: And would it possible to handle OTM, OTO, MTO, ... relations ?

If I have A human entity OTM with dog entity.
If I look for every human without dog (i.e. human.dogs.id null), or every dog with human without a name (i.e dog.human.name), could this filter be used ?

src/Filter/EmptyFieldFilter.php
VincentLanglet: An "EmptyOperatorType" like this https://github.com/sonata-project/SonataAdminBundle/blob/3.x/src/Form/Type/Operator/ContainsOperatorType.php in the SonataAdminBundle would be better.

This allow to be used by others PersistenceBundle.

tests/Filter/EmptyFieldFilterTest.php
```diff
@@ -0,0 +1,36 @@
...
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+
+final class EmptyFieldFilterTest extends TestCase
+{
+    public function testRenderSettings(): void
+    {
+        $filter = new EmptyFieldFilter();
+        $filter->initialize('field_name', [
+            'field_options' => ['class' => 'FooBar'],
+            ]);
```
phansys, line 27:
```suggestion
        ]);
```

src/Filter/EmptyFieldFilter.php
```diff
@@ -0,0 +1,98 @@
...
+
+        if ((int) $data['value'] === static::TYPE_EMPTY) {
+            $queryBuilder->andWhere(sprintf('%s.%s IS NULL', $alias, $field));
+        } elseif ((int) $data['value'] === static::TYPE_NOT_EMPTY) {
+            $queryBuilder->andWhere(sprintf('%s.%s IS NOT NULL', $alias, $field));
+        } else {
+            throw new \UnexpectedValueException('Invalid filter value given');
+        }
+
+        $this->active = null !== $data['value'];
```
VincentLanglet, line 46: Others filter of this bundle doesn't touch the `$this->active` property.

It's handle https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Filter/Filter.php#L62 for what I understand.

src/Filter/EmptyFieldFilter.php
```diff
@@ -0,0 +1,98 @@
...
+    public const TYPE_NOT_EMPTY = 2;
+
+    public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data): void
+    {
+        if (null === $data || !\is_array($data) || !\array_key_exists('value', $data)) {
+            return;
+        }
+
+        if ((int) $data['value'] === static::TYPE_EMPTY) {
+            $queryBuilder->andWhere(sprintf('%s.%s IS NULL', $alias, $field));
```
VincentLanglet, line 39: You have to use `applyWhere` to handle both OR filter and AND filter.

src/Filter/EmptyFieldFilter.php
```diff
@@ -0,0 +1,98 @@
...
+        } elseif ((int) $data['value'] === static::TYPE_NOT_EMPTY) {
+            $queryBuilder->andWhere(sprintf('%s.%s IS NOT NULL', $alias, $field));
```
VincentLanglet, line 41: You have to use `applyWhere` to handle both OR filter and AND filter.

src/Filter/EmptyFieldFilter.php
```diff
@@ -0,0 +1,98 @@
...
+            'choices' => [
+                'label_type_empty' => static::TYPE_EMPTY,
+                'label_type_not_empty' => static::TYPE_NOT_EMPTY,
+            ],
+        ]);
+    }
+
+    public function getRenderSettings()
+    {
+        return [DefaultType::class, [
```
VincentLanglet, line 71: WDYT about using the `yes/no` BooleanType ?

`Firstname: Yes` returns every not null values.
`Firstname: No` returns every null values.

core23:
Thanks for the Feedback @VincentLanglet . This was just some very old code from my private project I want to share.

src/Filter/EmptyFieldFilter.php
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public function getParentAssociationMappings()
+    {
+        $mappings = $this->getOption('parent_association_mappings', []);
+
+        $fields = explode('.', $this->getFieldName());
+        for ($i = 0; $i < \count($fields) - 1; ++$i) {
+            $mappings[] = ['fieldName' => $fields[$i]];
+        }
```
phansys, line 78:
```suggestion
        $fields = explode('.', $this->getFieldName(), -1);
        foreach ($fields as $field) {
            $mappings[] = ['fieldName' => $field];
        }
```

src/Filter/EmptyFieldFilter.php
```diff
@@ -0,0 +1,97 @@
...
+     * @return string[]
+     */
+    protected function association(ProxyQueryInterface $queryBuilder, $data): array
+    {
+        $alias = $queryBuilder->entityJoin($this->getParentAssociationMappings());
+        $part = strrchr('.'.$this->getFieldName(), '.');
+        $fieldName = substr(false === $part ? $this->getFieldType() : $part, 1);
+
+        return [$alias, $fieldName];
+    }
```
VincentLanglet, line 96: I'm curious. Why do you need this ?

There is nothing similar in `StringFilter`, `NumberFilter`, `ChoiceFilter` or `BooleanFilter`. 
What is different here ?

tests/Filter/EmptyFieldFilterTest.php
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Sonata\DoctrineORMAdminBundle\Tests\Filter;
+
+use PHPUnit\Framework\TestCase;
+use Sonata\DoctrineORMAdminBundle\Filter\EmptyFieldFilter;
+use Sonata\Form\Type\BooleanType;
+
+final class EmptyFieldFilterTest extends TestCase
+{
+    public function testRenderSettings(): void
```
VincentLanglet, line 22: I think more tests could be great, at least two more tests, testing the query for the null case and the not null case.

core23: Class is now 100% covered

phansys:
Thank you @core23!

PR #989, opened 2020-02-20
Breaks if entity has many identifiers and one of the not last identifiers is an entity

peter-gribanov:
Not overwrite the entity metadata on get the identifier metadata in `ModelManager::getIdentifierValues()`

## Subject

Critical error here:
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Model/ModelManager.php#L351

You overwrite the entity metadata variable on get the identifier metadata and in the next itarations we tring get metadata from previous identifier, not a entity metadata.

```php
public function getIdentifierValues($entity)
{
    $class = ClassUtils::getClass($entity);
    // Entity metadata!!!
    $metadata = $this->getMetadata($class);
    $platform = $this->getEntityManager($class)->getConnection()->getDatabasePlatform();

    $identifiers = [];

    foreach ($metadata->getIdentifierValues($entity) as $name => $value) {
        if (!\is_object($value)) {
            $identifiers[] = $value;

            continue;
        }

        $type = $fieldType && Type::hasType($fieldType) ? Type::getType($fieldType) : null;
        if ($type) {
            $identifiers[] = $this->getValueFromType($value, $type, $fieldType, $platform);

            continue;
        }

        // Entity metadata overwritten to identifier metadata!!!
        $metadata = $this->getMetadata(ClassUtils::getClass($value));

        foreach ($metadata->getIdentifierValues($value) as $value) {
            $identifiers[] = $value;
        }
    }

    return $identifiers;
}
```

Entity for reproduce bug

```php
/**
 * @ORM\Entity(readOnly=true)
 * @ORM\Table(name="billing__transaction")
 */
class TransactionAdminView
{
    /**
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="NONE")
     * @ORM\OneToOne(targetEntity="AccountAdminView", fetch="EAGER")
     * @ORM\JoinColumn(name="account_holder", referencedColumnName="id", nullable=false)
     *
     * @var AccountAdminView
     */
    public $account;

    /**
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="NONE")
     * @ORM\Column(name="transaction", type="BillingTransactionType", nullable=false)
     *
     * @var TransactionType
     */
    public $transaction;
}
```

## Changelog

```markdown
### Fixed

 * Breaks if entity has many identifiers and one of the not last identifiers is an entity.
```

### DIFF
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -348,9 +348,9 @@ class ModelManager implements ModelManagerInterface, LockInterface
                 continue;
             }
 
-            $metadata = $this->getMetadata(ClassUtils::getClass($value));
+            $identifierMetadata = $this->getMetadata(ClassUtils::getClass($value));
 
-            foreach ($metadata->getIdentifierValues($value) as $value) {
+            foreach ($identifierMetadata->getIdentifierValues($value) as $value) {
                 $identifiers[] = $value;
             }
         }


src/Model/ModelManager.php
```diff
@@ -348,9 +348,9 @@ public function getIdentifierValues($entity)
                 continue;
             }
 
-            $metadata = $this->getMetadata(ClassUtils::getClass($value));
+            $identifierMetadata = $this->getMetadata(ClassUtils::getClass($value));
```
core23: This is just a variable rename?

peter-gribanov: Yes

greg0ire: The goal is to avoid reusing the same variable to hold 2 different objects 👍

greg0ire:
The bug was introduced in ec1949a98c1dddf79533db0b272f9573ad6e63ac

greg0ire:
Thanks @peter-gribanov

PR #990, opened 2020-02-20
Address PHPUnit's warnings

greg0ire:
## Subject

I am targeting this branch, because it is a BC.

I tried using Rector, by using
```
composer require rector/rector-prefixed phpunit/phpunit --dev
vendor/bin/rector process tests --set phpunit90 --dry-run
```

You can see the output here: [rector_output.txt](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/files/4232688/rector_output.txt)

Then I had a look and saw why it does not help: https://github.com/rectorphp/rector/blob/master/config/set/phpunit/phpunit90.yaml

So I ended up doing the changes manually.


greg0ire:
I also check that it actually works with PHPUnit 9 (no fixing the warnings is not a guarantee of that).

franmomu:
Thanks @greg0ire !

PR #997, opened 2020-03-04
use SortBy before custom orderBy

VincentLanglet:
## Subject

I am targeting this branch, because I don't consider this like a BC-break but like a bug fix.

When adding
```
$query->addOrderBy()
```
In the `createQuery` function of an Admin class, the `orderBy` clauses was applied before the `sort_by` parameter of the datagrid value. Which break the behaviour of the list orders.
But in this situation I expect the opposite behaviour
- The list is first ordered by the selected column
- In case of row with the same values, the `orderBy` in the createQuery is used.

## Changelog
```markdown
### Fixed
The `_sort_by_ ` datagrid value is properly applied before any custom `orderBy`.
```

### DIFF
--- a/src/Datagrid/ProxyQuery.php
+++ b/src/Datagrid/ProxyQuery.php
@@ -124,13 +124,18 @@ class ProxyQuery implements ProxyQueryInterface
 
         // todo : check how doctrine behave, potential SQL injection here ...
         if ($this->getSortBy()) {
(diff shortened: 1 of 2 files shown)


VincentLanglet:
@core23 Test are added :)

greg0ire:
Thanks @VincentLanglet !

PR #998, opened 2020-03-04
Handle case without sort_by value

VincentLanglet:
## Subject

Following https://github.com/sonata-project/SonataAdminBundle/issues/5929, I want to introduce the ability of using `sortable => true` without a default column to sort the list.
This means that `$value['_sort_by']` can be 'not set'. 

## Changelog

```markdown
### Added
Allow `_sort_by` filter to not be initially defined. 
```

### DIFF
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -500,10 +500,7 @@ class ModelManager implements ModelManagerInterface, LockInterface
     {
         $values = $datagrid->getValues();
 
-        if (isset($values['_sort_by'])
-            && ($values['_sort_by']->getName() === $fieldDescription->getName()
-                || $values['_sort_by']->getName() === $fieldDescription->getOption('sortable'))
-        ) {
+        if ($this->isFieldAlreadySorted($fieldDescription, $datagrid)) {
             if ('ASC' === $values['_sort_order']) {
                 $values['_sort_order'] = 'DESC';
             } else {
@@ -632,6 +629,18 @@ class ModelManager implements ModelManagerInterface, LockInterface
     protected function camelize($property)
     {
         return str_replace(' ', '', ucwords(str_replace('_', ' ', $property)));
+    }
+
+    private function isFieldAlreadySorted(FieldDescriptionInterface $fieldDescription, DatagridInterface $datagrid): bool
+    {
+        $values = $datagrid->getValues();
+
+        if (!isset($values['_sort_by'])) {
+            return false;
+        }
+
+        return $values['_sort_by']->getName() === $fieldDescription->getName()
+            || $values['_sort_by']->getName() === $fieldDescription->getOption('sortable');
     }
 
     /**
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -500,7 +500,10 @@ class ModelManager implements ModelManagerInterface, LockInterface
     {
         $values = $datagrid->getValues();
 
-        if ($fieldDescription->getName() === $values['_sort_by']->getName() || $values['_sort_by']->getName() === $fieldDescription->getOption('sortable')) {
+        if (isset($values['_sort_by'])
+            && ($values['_sort_by']->getName() === $fieldDescription->getName()
+                || $values['_sort_by']->getName() === $fieldDescription->getOption('sortable'))
+        ) {
             if ('ASC' === $values['_sort_order']) {
                 $values['_sort_order'] = 'DESC';
             } else {
@@ -519,7 +522,9 @@ class ModelManager implements ModelManagerInterface, LockInterface
     {
         $values = $datagrid->getValues();
 
-        $values['_sort_by'] = $values['_sort_by']->getName();
+        if (isset($values['_sort_by'])) {
+            $values['_sort_by'] = $values['_sort_by']->getName();
+        }
         $values['_page'] = $page;
 
         return ['filter' => $values];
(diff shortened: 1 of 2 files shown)


src/Model/ModelManager.php
```diff
@@ -500,7 +500,10 @@ public function getSortParameters(FieldDescriptionInterface $fieldDescription, D
     {
         $values = $datagrid->getValues();
 
-        if ($fieldDescription->getName() === $values['_sort_by']->getName() || $values['_sort_by']->getName() === $fieldDescription->getOption('sortable')) {
+        if (isset($values['_sort_by'])
```
core23: This looks very complicated. You  can extra a meaningful method?

VincentLanglet: I agree it's complicated, but I don't know the meaning of this check.
I just wanted to add a check for the definition of `_sort_by`.

So I can't really give meaning to this...

VincentLanglet: @core23 Do you have a name for the check ? Because I don't.

I prefer to let this like it was and just adding an isset to not create a wrong name

franmomu: how about `isTheFieldAlreadySorted`? or similar, looks like it what it does.

VincentLanglet: @core23 @franmomu Done.

greg0ire:
If the list is not sorted, [the SGBDR may return rows in whatever order it sees fit, and that may change over time](https://stackoverflow.com/a/20050403/353612). Not an issue when there is less than one page of results, but if there are more results, this will mean bugs. Not sure we want to deal with those.

VincentLanglet:
> If the list is not sorted, [the SGBDR may return rows in whatever order it sees fit, and that may change over time](https://stackoverflow.com/a/20050403/353612).

Everytime I made query, my SQL database ordered by id by default... So I'm surprised.

But anyway, there is already a protection in the function `execute` of the ProxyQuery. @greg0ire 
```
/* By default, always add a sort on the identifier fields of the first
         * used entity in the query, because RDBMS do not guarantee a
         * particular order when no ORDER BY clause is specified, or when
         * the field used for sorting is not unique.
         */

        $identifierFields = $queryBuilder
            ->getEntityManager()
            ->getMetadataFactory()
            ->getMetadataFor(current($queryBuilder->getRootEntities()))
            ->getIdentifierFieldNames();

        $existingOrders = [];
        /** @var Query\Expr\OrderBy $order */
        foreach ($queryBuilder->getDQLPart('orderBy') as $order) {
            foreach ($order->getParts() as $part) {
                $existingOrders[] = trim(str_replace([Criteria::DESC, Criteria::ASC], '', $part));
            }
        }

        foreach ($identifierFields as $identifierField) {
            $order = $rootAlias.'.'.$identifierField;
            if (!\in_array($order, $existingOrders, true)) {
                $queryBuilder->addOrderBy(
                    $order,
                    $this->getSortOrder() // reusing the sort order is the most natural way to go
                );
            }
        }
```

Instead using a default `sortBy` value in the AdminBundle (which mean that I use some `addOrderBy`, the default `sortBy` is applied before my custom `addOrderBy` and ruining everything), I think it's better to not having a default `sortBy` in the Admin and using one in the `execute` function of the ProxyQuery (and this is already implemented).

This way: 
- I can use `sortBy`
- I can use `addOrderBy` without `sortBy`
- I can use both `addOrderBy` and `sortBy`

greg0ire:
Never witnessed myself either but if it can happen, I'd rather not take the risk. Anyway, since there is this last resort protection, LGTM

VincentLanglet:
@greg0ire Can you take a look at https://github.com/sonata-project/SonataAdminBundle/issues/5929#issuecomment-594474199 ?

Since there is always a `orderBy('id')` in the query by default, I would like to remove the default `_sort_by => 'id'` set in the `modelManager->getDefaultSortValues`.

Plus I found out that the `_sort_order`, `_sort_by`, `_page`, ... could be overridden by 3 differents way. Globally with `modelManager->getDefaultSortValues` and locally in the admin with the attribute `datagridValues` or the method `getDefaultFilterValues`. Should we deprecate some ?

src/Model/ModelManager.php
```diff
@@ -519,7 +519,9 @@ public function getPaginationParameters(DatagridInterface $datagrid, $page)
     {
         $values = $datagrid->getValues();
 
-        $values['_sort_by'] = $values['_sort_by']->getName();
+        if (isset($values['_sort_by'])) {
```
core23: No `instanceof` check?

src/Model/ModelManager.php
```diff
@@ -629,6 +631,18 @@ protected function camelize($property)
         return str_replace(' ', '', ucwords(str_replace('_', ' ', $property)));
     }
 
+    private function isFieldAlreadySorted(FieldDescriptionInterface $fieldDescription, DatagridInterface $datagrid): bool
+    {
+        $values = $datagrid->getValues();
+
+        if (!isset($values['_sort_by'])) {
```
core23: Same here

src/Model/ModelManager.php
VincentLanglet: Shouldn't the check be somewhere else ?

if set, $values['_sort_by'] should be a FieldDescriptionInterface.
So I don't expect a check here, I expect an exception when I tried to set something else.

core23: It's a public method that uses an untyped array value of the object, so you should check it.

VincentLanglet: @core23 Done

VincentLanglet:
@greg0ire You didn't approved the PR, is it ok for you if I merge this ? :)

PR #999, opened 2020-03-06
allow null selection in ChoiceFilter

### LINKED ISSUE #991: ChoiceFilter should handle NULL value (and have a better support of ALL value).
## Feature Request

This is the `ChoiceFilter` code
```php
if (\is_array($data['value'])) {
    if (0 === \count($data['value'])) {
        return;
    }

    if (\in_array('all', $data['value'], true)) {
        return;
    }

    // Have to pass IN array value as parameter. See: http://www.doctrine-project.org/jira/browse/DDC-3759
    $completeField = sprintf('%s.%s', $alias, $field);
    if (ChoiceType::TYPE_NOT_CONTAINS === $data['type']) {
        $this->applyWhere($queryBuilder, $queryBuilder->expr()->notIn($completeField, ':'.$parameterName));
    } else {
        $this->applyWhere($queryBuilder, $queryBuilder->expr()->in($completeField, ':'.$parameterName));
    }
} else {
    if ('' === $data['value'] || null === $data['value'] || false === $data['value'] || 'all' === $data['value']) {
        return;
    }

    if (ChoiceType::TYPE_NOT_CONTAINS === $data['type']) {
        $this->applyWhere($queryBuilder, sprintf('%s.%s <> :%s', $alias, $field, $parameterName));
    } else {
        $this->applyWhere($queryBuilder, sprintf('%s.%s = :%s', $alias, $field, $parameterName));
    }

}
```

You can see that:
- There is a specific behaviour for the `all` value. AFIK, this is not documented. And if you chose the `all` value, the filter is not applied, even if you chose the `NOT CONTAINS` operator. But, for me it should return no results instead of every result. IMHO, the `all` behaviour should be deprecated or at least fixed.
- I have multiple filter where sometimes I want to get all entities with the value A, sometimes all entities with the value B, and sometimes all entities with no value in database since my field is nullable. Actually the choice filter does not handle null values. 

So if `$data['value'] = null`, I expect the query to do `IS NULL` or `IS NOT NULL`.
And if `$data[`value`]` is an array with a null value, I expect to add a `OR IS NULL` check.

I can make the PR if you agree with the changes.

rgrassian:
## Subject

Closes #991

## Changelog

```markdown
### Added
- allow null to be selected in ChoiceFilter
```

### DIFF
--- a/src/Filter/ChoiceFilter.php
+++ b/src/Filter/ChoiceFilter.php
@@ -87,7 +87,7 @@ class ChoiceFilter extends Filter
         }
     }
 
-    private function filterWithSingleValue(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
+    private function filterWithSingleValue(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $data = []): void
     {
         if ('' === $data['value'] || false === $data['value'] || 'all' === $data['value']) {
             return;
--- a/src/Filter/ChoiceFilter.php
+++ b/src/Filter/ChoiceFilter.php
@@ -52,7 +52,7 @@ class ChoiceFilter extends Filter
         ]];
     }
 
-    private function filterWithMultipleValues(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
+    private function filterWithMultipleValues(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $data = []): void
     {
         if (0 === \count($data['value'])) {
             return;
--- a/src/Filter/ChoiceFilter.php
+++ b/src/Filter/ChoiceFilter.php
@@ -27,37 +27,9 @@ class ChoiceFilter extends Filter
         }
 
         if (\is_array($data['value'])) {
-            if (0 === \count($data['value'])) {
-                return;
-            }
-
-            if (\in_array('all', $data['value'], true)) {
-                return;
-            }
-
-            // Have to pass IN array value as parameter. See: http://www.doctrine-project.org/jira/browse/DDC-3759
-            $completeField = sprintf('%s.%s', $alias, $field);
-            $parameterName = $this->getNewParameterName($queryBuilder);
-            if (ChoiceType::TYPE_NOT_CONTAINS === $data['type']) {
-                $this->applyWhere($queryBuilder, $queryBuilder->expr()->notIn($completeField, ':'.$parameterName));
-            } else {
-                $this->applyWhere($queryBuilder, $queryBuilder->expr()->in($completeField, ':'.$parameterName));
-            }
-            $queryBuilder->setParameter($parameterName, $data['value']);
+            $this->filterWithMultipleValues($queryBuilder, $alias, $field, $data);
         } else {
-            if ('' === $data['value'] || null === $data['value'] || false === $data['value'] || 'all' === $data['value']) {
-                return;
-            }
-
-            $parameterName = $this->getNewParameterName($queryBuilder);
-
-            if (ChoiceType::TYPE_NOT_CONTAINS === $data['type']) {
-                $this->applyWhere($queryBuilder, sprintf('%s.%s <> :%s', $alias, $field, $parameterName));
-            } else {
-                $this->applyWhere($queryBuilder, sprintf('%s.%s = :%s', $alias, $field, $parameterName));
-            }
-
-            $queryBuilder->setParameter($parameterName, $data['value']);
+            $this->filterWithSingleValue($queryBuilder, $alias, $field, $data);
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/Filter/ChoiceFilter.php
```diff
@@ -35,29 +35,57 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
                 return;
             }
 
+            $isNullSelected = \in_array(null, $data['value'], true);
+            $data['value'] = array_filter($data['value'], static function ($value) {
+                return null !== $value;
+            });
```
peter-gribanov: I think that here you can get empty array.

src/Filter/ChoiceFilter.php
```diff
@@ -35,29 +35,57 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
...
+                    ));
+                    $queryBuilder->setParameter($parameterName, $data['value']);
+                }
             } else {
-                $this->applyWhere($queryBuilder, $queryBuilder->expr()->in($completeField, ':'.$parameterName));
+                if ($isNullSelected) {
+                    $this->applyWhere($queryBuilder, $queryBuilder->expr()->orX(
+                        $queryBuilder->expr()->in($completeField, ':'.$parameterName),
+                        $queryBuilder->expr()->isNull($completeField)
+                    ));
```
peter-gribanov: Empty data array will result in an incorrect query here.
```sql
SELECT * FROM ... WHERE field IN () OR field IS NULL
```

VincentLanglet: Empty data array will no result in an incorrect query.
Empty data are always replaced by `[null]`.

And we get the following query
```
SELECT * FROM ... WHERE field IN (NULL) OR field IS NULL
```

src/Filter/ChoiceFilter.php
VincentLanglet: Yes, but this is not a problem.

peter-gribanov:
IMHO The code turned out to be rather cumbersome and difficult to understand. Maybe you should apply the strategy pattern or separate the logic by methods?

src/Filter/ChoiceFilter.php
peter-gribanov: Yes. You're right. I'm confusing something.

VincentLanglet:
> IMHO The code turned out to be rather cumbersome and difficult to understand. Maybe you should apply the strategy pattern or separate the logic by methods?

We improved the way the query is build. I don't really see any difficulty to understand what is done so I don't think we should create privateMethod just to create privateMethod.

VincentLanglet:
@greg0ire @OskarStark @phansys I'll be happy if you find some time to review this PR :)

VincentLanglet:
> Looks good to me, but I didn't test it in a real project 👍

We started by writing the code directly in our vendor in order to check this works.

VincentLanglet:
@greg0ire @core23 @phansys If you have time to take look, it would be great.
We would like to use this new feature in our project :)

src/Filter/ChoiceFilter.php
```diff
@@ -35,29 +35,51 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
...
+            $data['value'] = array_filter($data['value'], static function ($value) {
+                return null !== $value;
+            });
+
             // Have to pass IN array value as parameter. See: http://www.doctrine-project.org/jira/browse/DDC-3759
             $completeField = sprintf('%s.%s', $alias, $field);
             $parameterName = $this->getNewParameterName($queryBuilder);
             if (ChoiceType::TYPE_NOT_CONTAINS === $data['type']) {
-                $this->applyWhere($queryBuilder, $queryBuilder->expr()->notIn($completeField, ':'.$parameterName));
+                $andConditions = [$queryBuilder->expr()->isNotNull($completeField)];
```
core23: This method is very long. Can you please try to extract some small methods. This improves reliability.

VincentLanglet: Does 
```
if (\is_array($data['value'])) {
    $this->filterMultipleValues(...)
else {
    $this->filterSimpleValue()
}
```

would be enough ? @core23 @peter-gribanov

src/Filter/ChoiceFilter.php
```diff
@@ -35,29 +35,51 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
...
             }
 
             $parameterName = $this->getNewParameterName($queryBuilder);
 
             if (ChoiceType::TYPE_NOT_CONTAINS === $data['type']) {
-                $this->applyWhere($queryBuilder, sprintf('%s.%s <> :%s', $alias, $field, $parameterName));
+                if (null === $data['value']) {
+                    $this->applyWhere($queryBuilder, $queryBuilder->expr()->isNotNull(sprintf('%s.%s', $alias, $field)));
+                } else {
+                    $this->applyWhere($queryBuilder, sprintf('%s.%s <> :%s', $alias, $field, $parameterName));
```
phansys: I'd use `!=` or `Expr::neq()` instead.

tests/Filter/ChoiceFilterTest.php
```diff
@@ -55,11 +55,35 @@ public function testFilterArray(): void
 
         $filter->filter($builder, 'alias', 'field', ['type' => ChoiceType::TYPE_CONTAINS, 'value' => ['1', '2']]);
 
-        $this->assertSame(['in_alias.field', 'in_alias.field IN :field_name_0'], $builder->query);
+        $this->assertSame('in_alias.field', $builder->query[0]);
+        $this->assertSame('in_alias.field IN :field_name_0', $builder->query[1]->__toString());
```
phansys:
```suggestion
        $this->assertSame('in_alias.field IN :field_name_0', (string) $builder->query[1]);
```

tests/Filter/ChoiceFilterTest.php
```diff
@@ -55,11 +55,35 @@ public function testFilterArray(): void
...
+    {
+        $filter = new ChoiceFilter();
+        $filter->initialize('field_name', ['field_options' => ['class' => 'FooBar']]);
+
+        $builder = new ProxyQuery(new QueryBuilder());
+
+        $filter->filter($builder, 'alias', 'field', ['type' => ChoiceType::TYPE_CONTAINS, 'value' => ['1', null]]);
+
+        $this->assertSame('in_alias.field', $builder->query[0]);
+        $this->assertSame('in_alias.field IN :field_name_0 OR alias.field IS NULL', $builder->query[1]->__toString());
```
phansys:
```suggestion
        $this->assertSame('in_alias.field IN :field_name_0 OR alias.field IS NULL', (string) $builder->query[1]);
```

tests/Filter/ChoiceFilterTest.php
```diff
@@ -55,11 +55,35 @@ public function testFilterArray(): void
...
+        $this->assertSame(['field_name_0' => ['1']], $builder->parameters);
+        $this->assertTrue($filter->isActive());
+
+        $builder = new ProxyQuery(new QueryBuilder());
+
+        $filter->filter($builder, 'alias', 'field', ['type' => ChoiceType::TYPE_NOT_CONTAINS, 'value' => ['1', null]]);
+
+        $this->assertSame('alias.field IS NOT NULL AND alias.field NOT IN :field_name_0', $builder->query[0]->__toString());
```
phansys:
```suggestion
        $this->assertSame('alias.field IS NOT NULL AND alias.field NOT IN :field_name_0', (string) $builder->query[0]);
```

tests/Filter/QueryBuilder.php
```diff
@@ -93,6 +94,14 @@ public function orX($x = null)
         return new Orx(\func_get_args());
     }
 
+    /**
+     * @return Andx
+     */
+    public function andX($x = null)
```
phansys:
```suggestion
    public function andX($x = null): Andx
```

tests/Filter/QueryBuilder.php
```diff
@@ -114,6 +123,16 @@ public function isNull($queryPart)
         return $queryPart.' IS NULL';
     }
 
+    /**
+     * @param string $queryPart
+     *
+     * @return string
+     */
+    public function isNotNull($queryPart)
```
phansys:
```suggestion
    public function isNotNull(string $queryPart): string
```

src/Filter/ChoiceFilter.php
```diff
@@ -79,4 +51,64 @@ public function getRenderSettings()
             'label' => $this->getLabel(),
         ]];
     }
+
+    private function filterWithMultipleValues(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
```
phansys:
```suggestion
    private function filterWithMultipleValues(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $data = []): void
```

src/Filter/ChoiceFilter.php
```diff
@@ -79,4 +51,64 @@ public function getRenderSettings()
...
+            $orConditions = [$queryBuilder->expr()->in($completeField, ':'.$parameterName)];
+            if ($isNullSelected) {
+                $orConditions[] = $queryBuilder->expr()->isNull($completeField);
+            }
+            $this->applyWhere($queryBuilder, $queryBuilder->expr()->orX()->addMultiple($orConditions));
+            $queryBuilder->setParameter($parameterName, $data['value']);
+        }
+    }
+
+    private function filterWithSingleValue(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
```
phansys:
```suggestion
    private function filterWithSingleValue(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $data = []): void
```

tests/Filter/QueryBuilder.php
```diff
@@ -114,6 +120,14 @@ public function isNull($queryPart)
         return $queryPart.' IS NULL';
     }
 
+    /**
+     * @param string $queryPart
+     */
+    public function isNotNull($queryPart): string
```
phansys:
```suggestion
    public function isNotNull(string $queryPart): string
```

rgrassian:
@core23 and @phansys it's done

VincentLanglet:
@core23 Can you take a look ? I wait for you approbation to merge this :)

VincentLanglet:
Thanks !

VincentLanglet:
@greg0ire Can you make a new release ? :)

PR #1012, opened 2020-03-24
Update composer.json

wbloszyk:
## Subject

Allow using SonataAdminBundle 4.x

I am targeting this branch, because SonataAdminBundlle 4.x is in dev.

Or should I add it in 3.x branch? SonataCoreBundle should avoid install this bundle 4.x (next major) with SonataAdminBundle 3.x.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,8 +27,7 @@
         "doctrine/doctrine-bundle": "^1.8 || ^2.0",
         "doctrine/orm": "^2.5",
         "doctrine/persistence": "^1.3.4",
-        "sonata-project/admin-bundle": "^3.61",
-        "sonata-project/core-bundle": "^3.14",
+        "sonata-project/admin-bundle": "^3.63 || ^4.0",
         "sonata-project/exporter": "^1.11.0 || ^2.0",
         "symfony/config": "^4.4",
         "symfony/console": "^4.4",
@@ -50,7 +49,7 @@
     },
     "require-dev": {
         "simplethings/entity-audit-bundle": "^0.9 || ^1.0",
-        "sonata-project/block-bundle": "^3.17",
+        "sonata-project/block-bundle": "^3.18 || ^4.0",
         "symfony/phpunit-bridge": "^5.0"
     },
     "suggest": {


composer.json
```diff
@@ -27,8 +27,7 @@
         "doctrine/doctrine-bundle": "^1.8 || ^2.0",
         "doctrine/orm": "^2.5",
         "doctrine/persistence": "^1.3.4",
-        "sonata-project/admin-bundle": "^3.61",
-        "sonata-project/core-bundle": "^3.14",
+        "sonata-project/admin-bundle": "^3.63 || ^4.0",
```
greg0ire: Right now the tests pass, but what is actually installed is the admin bundle v3. Besides, there is already a job dedicated to testing with admin v4:

https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/8a7805f49ec9895517af44b8d6df5874d676f88e/.travis.yml#L54

wbloszyk: I know about it. This PR is for allow using(install) DoctrineORMAdminBundle(master) with AdminBundle 4.x(master). Before major release dev-kit should be update, travis here should be update. Composer should be update.  

New DoctrineORMAdminBundle should not support CoreBundle:
> https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/8a7805f49ec9895517af44b8d6df5874d676f88e/.travis.yml#L40

greg0ire: > This PR is for allow using(install) DoctrineORMAdminBundle(master) with AdminBundle 4.x(master).

Why do you want to do that?

VincentLanglet: Indeed, there no 4.0 version for the admin bundle so I do not get it...

wbloszyk: I'm working on AdminLTE 3 integration and webpack. I have to do it in master versions.

wbloszyk: AdminLTE 3 should be add to AdminBundle And all storage-AdminBundle

greg0ire: You should be able to use the same sort of inline alias trick, `dev-master as 3.0.0` for admin bundle should work, but maybe that's somehow inconvenient?

PR #1014, opened 2020-03-24
Remove dependency for "sonata-project/core-bundle"

phansys:
## Subject
Remove dependency for "sonata-project/core-bundle".

Occurrences of "Sonata\\CoreBundle\\" and "Sonata\\Form\\" in this package are used just to get class names, so I've replaced them with hardcoded strings.
The dependency is yet present because "sonata-project/admin-bundle" and "sonata-project/block-bundle".
```
composer why sonata-project/core-bundle
sonata-project/admin-bundle  3.45.1  requires  sonata-project/core-bundle (^3.14.0)  
sonata-project/block-bundle  3.14.0  requires  sonata-project/core-bundle (^3.13.7)  
```

## Changelog

```markdown
### Removed
- Removed dependency against "sonata-project/core-bundle".
```
## To do
    
- [x] Check if there is a service or something else provided by "sonata-project/core-bundle".

### DIFF
--- a/src/Builder/FormContractor.php
+++ b/src/Builder/FormContractor.php
@@ -23,8 +23,6 @@ use Sonata\AdminBundle\Form\Type\ModelHiddenType;
 use Sonata\AdminBundle\Form\Type\ModelListType;
 use Sonata\AdminBundle\Form\Type\ModelType;
 use Sonata\AdminBundle\Form\Type\ModelTypeList;
-use Sonata\CoreBundle\Form\Type\CollectionType as DeprecatedCollectionType;
-use Sonata\Form\Type\CollectionType;
 use Symfony\Component\Form\Extension\Core\Type\FormType;
 use Symfony\Component\Form\FormFactoryInterface;
 
@@ -149,12 +147,13 @@ class FormContractor implements FormContractorInterface
                 ClassMetadata::ONE_TO_ONE,
                 ClassMetadata::MANY_TO_ONE,
             ], true)) {
+                // NEXT_MAJOR: Import the class from "sonata-project/form-extensions" and use `CollectionType::class` instead.
                 throw new \RuntimeException(sprintf(
                     'You are trying to add `%s` field `%s` which is not One-To-One or  Many-To-One.'
                     .' Maybe you want `%s` instead?',
                     AdminType::class,
                     $fieldDescription->getName(),
-                    CollectionType::class
+                    'Sonata\Form\Type\CollectionType'
                 ));
             }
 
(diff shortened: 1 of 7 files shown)


phansys:
All services declared at "sonata-project/core-bundle" start with "sonata.core.", and in this package there are no consumed services with that name. So, IMO this PR is able to be merged.

src/Builder/FormContractor.php
```diff
@@ -154,7 +152,7 @@ public function getDefaultOptions($type, FieldDescriptionInterface $fieldDescrip
                     .' Maybe you want `%s` instead?',
                     AdminType::class,
                     $fieldDescription->getName(),
-                    CollectionType::class
+                    'Sonata\Form\Type\CollectionType'
```
VincentLanglet: Should we add a Todo here (and for every others changes), to use `::class` again when possible ?

phansys: I've omitted these comments because I thought it will be an obvious decision, but I can add them if there is consensus about the requirement.

VincentLanglet: IMO, adding a comment
- Will help to not forget,
- Find all changes to do in one grep
- Will help to know the new value 

For example, what will be the new value when I see `'Sonata\Form\Type\CollectionType'` and `'Sonata\CoreBundle\Form\Type\CollectionType'`, I can't be sure what to do, since there is only one CollectionType in the form-extension.

phansys: Do you think you could be in doubt about the decision between a class that belongs to `Sonata\Form\` namespace and other that belongs to `Sonata\CoreBundle\`? :thinking: 
The user already knows that SonataCoreBundle is deprecated. Even, there will be no choice that allows "sonata-project/core-bundle" to be installed in the same project.
Anyway, I've added the comments.

src/Filter/DateRangeFilter.php
```diff
@@ -33,6 +31,7 @@ class DateRangeFilter extends AbstractDateFilter
 
     public function getFieldType()
     {
-        return $this->getOption('field_type', DateRangeType::class);
+        // NEXT_MAJOR: Import the class from "sonata-project/form-extensions" and use `DateRangeType::class` instead.
+        return $this->getOption('field_type', 'Sonata\Form\Type\DateRangeType');
```
core23: Wouldn't it break, if you remove the core bundle depdency?

phansys: I think you're (sadly) right, since "Sonata\Form\Type\DateRangeType" is used as default value if the option "field_type" doesn't exist.
This means I should add one more item to the list of things I must be aware of:
* Usage of classes (new instances, static method calls, argument and return declarations, inheritance, `::class`, etc)
* Services
* Twig filters
* **Form types**

And finally, this also means this PR can't be merged :sob: 

Thank you!

core23: > And finally, this also means this PR can't be merged sob

That's sadly right :( We will have a hard upgrade path when releasing the new major versions

phansys:
I'm closing this since isn't possible to merge as is. See https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1014#discussion_r398471184.

PR #1015, opened 2020-03-26
Drop SonataCoreBundle

### LINKED ISSUE #988: SonataCoreBundle is deprecated
As stated on [its README](https://github.com/sonata-project/SonataCoreBundle/blob/3.x/README.md), SonataCoreBundle is deprecated.

This bundle should not rely on SonataCoreBundle anymore

wbloszyk:
## Subject

I am targeting this branch, because it its BC-break.

Closes #988 

## Changelog

```markdown
### Removed
- Removed `SonataCoreBundle` 
```


src/Builder/FormContractor.php
```diff
@@ -168,7 +167,7 @@ public function getDefaultOptions($type, FieldDescriptionInterface $fieldDescrip
             };
             $fieldDescription->setOption('edit', $fieldDescription->getOption('edit', 'admin'));
         // NEXT_MAJOR: Check only against FQCNs when dropping support for Symfony 2.8
```
VincentLanglet: 2.8 is dropped so you can check only FQCN

tests/Builder/FormContractorTest.php
```diff
@@ -95,7 +94,6 @@ public function testDefaultOptionsForSonataFormTypes(): void
         ];
         $collectionTypes = [
             'sonata_type_collection',
```
VincentLanglet: I think all these string can be removed too.

composer.json
```diff
@@ -27,9 +27,9 @@
         "doctrine/doctrine-bundle": "^1.8 || ^2.0",
         "doctrine/orm": "^2.5",
         "doctrine/persistence": "^1.3.4",
-        "sonata-project/admin-bundle": "^3.61",
-        "sonata-project/core-bundle": "^3.14",
+        "sonata-project/admin-bundle": "dev-master as 3.61",
```
VincentLanglet: What is the goal of this ?
Doesn't this say we should work on the AdminBundle 4.0 release before the DoctrineORMBundle release ?

wbloszyk: I upgrade sandbox to:
- symfony 4.4
- behat 3 (with runnable tests)
- webpack in frontend (working but need more work)

For now I want:
- drop CoreBundle becouse it make so many conflict
- update sandbox: update block or disabled where it is possible  

Then i can release sandbox demo which I need to my work.

wbloszyk: BTW I already install Admin + ORM : [http://demo.wbloszyk.pl](http://demo.wbloszyk.pl)

wbloszyk:
@VincentLanglet 
Goal is prepare bundles to upgrade. You have right, We should drop CoreBundle and support for BlockBundle 3.x in every bundles (master) only. Then install any bundles@dev will be easy.

wbloszyk:
Closing this in favor of #1048

PR #1016, opened 2020-03-26
Remove some NEXT_MAJOR in stable branch !

VincentLanglet:
## Subject

I am targeting this branch, because since Sf2.8 is dropped, it's BC.

## Changelog

```markdown
### Removed
Code used for 2.8 symfony support
```

### DIFF
--- a/src/Builder/FormContractor.php
+++ b/src/Builder/FormContractor.php
@@ -99,13 +99,7 @@ class FormContractor implements FormContractorInterface
         $options = [];
         $options['sonata_field_description'] = $fieldDescription;
 
-        // NEXT_MAJOR: Check only against FQCNs when dropping support for Symfony 2.8
-        if (\in_array($type, [
-            'sonata_type_model',
-            'sonata_type_model_list',
-            'sonata_type_model_hidden',
-            'sonata_type_model_autocomplete',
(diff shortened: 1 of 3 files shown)


greg0ire:
Thanks @VincentLanglet !

core23:
No special docs update for this BC break on the stable branch?

greg0ire:
There is no BC break IMO, 2.8 has already been dropped.

src/Builder/FormContractor.php
```diff
@@ -134,8 +127,7 @@ public function getDefaultOptions($type, FieldDescriptionInterface $fieldDescrip
                     ));
                 }
             }
-            // NEXT_MAJOR: Check only against FQCNs when dropping support for Symfony 2.8
-        } elseif ('sonata_type_admin' === $type || $this->checkFormClass($type, [AdminType::class])) {
```
core23: This could cause problems. We dropped a named form type. You can still define a named form type in symfony 5 @greg0ire

core23: Only the symfony (internal) form types removed this feature. This is not related to any other 3rd party form

greg0ire: Then maybe we should revert and change the comments

PR #1018, opened 2020-03-27
Add method annotation for ProxyQuery

VincentLanglet:
## Subject

What do you think about adding `@method` for IDE ?

## Changelog

```markdown
### Added
- `@method` Annotation for ProxyQuery
```

### DIFF
--- a/src/Datagrid/ProxyQuery.php
+++ b/src/Datagrid/ProxyQuery.php
@@ -15,12 +15,63 @@ namespace Sonata\DoctrineORMAdminBundle\Datagrid;
 
 use Doctrine\Common\Collections\Criteria;
 use Doctrine\DBAL\Types\Type;
+use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
 
 /**
  * This class try to unify the query usage with Doctrine.
+ *
+ * @method Query\Expr    expr()
+ * @method QueryBuilder  setCacheable($cacheable)
+ * @method bool          isCacheable()
+ * @method QueryBuilder  setCacheRegion($cacheRegion)
+ * @method string|null   getCacheRegion()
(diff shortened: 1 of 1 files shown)


src/Datagrid/ProxyQuery.php
```diff
@@ -15,12 +15,63 @@
...
+use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
 
 /**
  * This class try to unify the query usage with Doctrine.
+ *
+ * @method Query\Expr    expr()
+ * @method QueryBuilder  setCacheable($cacheable)
```
phansys: We should use the argument types exposed at `QueryBuilder` ([ref](https://github.com/doctrine/orm/blob/6138afdca9b78a75e162fecaf8be20fff99e15f8/lib/Doctrine/ORM/QueryBuilder.php#L184)).
```suggestion
 * @method QueryBuilder  setCacheable(bool $cacheable)
```

VincentLanglet: Since it's casting in boolean, it theorically accept anything

image

phansys: I know. I'm asking to enforce the type declaration from our side based on the `@param` docblock.

VincentLanglet:
```
@method QueryBuilder  setCacheable(bool $cacheable)
```
Wouldn't this let you believe if you want to override the function that you have to type the param: 
```
public function setCacheable(bool $cacheable) { ... }
```
And wouldn't this let you believe that$cacheable is always a bool when it can in fact be everything ?

I understand your point, but I was thinking as better we should use the same type declaration that the QueryBuilder use, no ?

phansys: It's fine. Let's save these updates for later :+1:

greg0ire:
Thanks @VincentLanglet !

PR #1022, opened 2020-03-29
Drop support of php 7.1

### LINKED ISSUE #1021: Composer contraints can't be satisfied on PHP 7.1
PHP 7.1

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - sonata-project/core-bundle 3.18.0 requires php ^7.2 -> your PHP version (7.1.28) does not satisfy that requirement.
    - sonata-project/admin-bundle 3.63.0 requires sonata-project/core-bundle ^3.18.0 -> satisfiable by sonata-project/core-bundle[3.18.0].
    - Installation request for sonata-project/admin-bundle ^3.63 -> satisfiable by sonata-project/admin-bundle[3.63.0].

```

triggered by #1017

VincentLanglet:
## Subject

Closes #1021

## Changelog

```markdown
### Removed
- Drop support of php 7.1
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         }
     ],
     "require": {
-        "php": "^7.1",
+        "php": "^7.2",
         "doctrine/doctrine-bundle": "^1.8 || ^2.0",
         "doctrine/orm": "^2.5",
         "doctrine/persistence": "^1.3.4",


VincentLanglet:
@phansys What do you expect in an upgrade note ?

I never saw an upgrade note when we dropped some symfony version, neither when we drop php 7.1 in the core-bundle https://github.com/sonata-project/SonataCoreBundle/pull/729

Composer will just resolve the version for the user.

phansys:
I'm in favor to treat PHP and any of its extensions like any other dependency. But I remember myself trying a update like this before in a package (I don't remember which one), and receiving a comment which now I consider valid. The PHP version is not easy to upgrade like a PHP package handled by composer, since in a conventional managed infrastructure, it generally requires the intervention of an IT team or similar.
That's the only concern I have around this, and that's why I think an upgrade note could be a good idea.

VincentLanglet:
> I'm in favor to treat PHP and any of its extensions like any other dependency. But I remember myself trying a update like this before in a package (I don't remember which one), and receiving a comment which now I consider valid. The PHP version is not easy to upgrade like a PHP package handled by composer, since in a conventional managed infrastructure, it generally requires the intervention of an IT team or similar.
> That's the only concern I have around this, and that's why I think an upgrade note could be a good idea.

I still don't understand, sorry 😅

First of all, it's already not possible to use the last version with php 7.1, as explained here: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1021
I'm just updating the composer.json.

Secondly, if it was still possible to use this package with php 7.1, when installing, the composer resolution will take the right version ; a previous one. 

Finally, I don't know what would be more necessary to say than `Drop of php 7.1 version support`, which is already in the Changelog. If the user need to know how to upgrade php version, he'll look at the documentation on internet, not on Sonata Upgrade note. This seems to be used for code migrations where explanation/examples are needed (https://github.com/sonata-project/SonataAdminBundle/blob/3.x/UPGRADE-3.x.md).

phansys:
If the package is not installable as is, forget about it and remove the label.
IMO, we should try to bump the PHP version only in major releases (since generally it involves syntax changes, deprecations and new features), that's why I associate a PHP version bump with the `UPGRADE` file.

greg0ire:
> we should try to bump the PHP version only in major releases 

For that we would have to release majors a bit more often :sweat_smile: 
I think we don't have to do that for php versions that are in red in the diagram. https://www.php.net/supported-versions.php

greg0ire:
Thanks @VincentLanglet !

PR #1024, opened 2020-03-31
Add explicit `null` return values at `ModelManager::getNormalizedIdentifier()`

phansys:
## Subject
Add explicit `null` return values at `ModelManager::getNormalizedIdentifier()`.

## Changelog

```markdown
### Fixed
- Fixed returning `void` in `ModelManager::getNormalizedIdentifier()`, which is intended to return a value or `null`.
```


greg0ire:
Thanks @phansys !

PR #1032, opened 2020-04-12
Fix camel case for `ModelManager::getUrlSafeIdentifier()` method

phansys:
## Subject
Fix camel case for `ModelManager::getUrlSafeIdentifier()` method.


franmomu:
Thanks @phansys !

PR #1042, opened 2020-05-06
Add Array filter to search value in `@ORM\Column(type="array")`

### LINKED ISSUE #1029: ArrayFilter : Contains, Not contains, Equal, Empty, ...
## Feature Request

It would be nice to have a filter for fields using `@ORM\Column(type="array")`.

FYI, the data is save as 
```
 a:0:{} 
 a:1:{i:0;s:47:"lead.form.specific_documentations.older_than_85";} 
 a:2:{i:0;s:47:"lead.form.specific_documentations.older_than_85";i:1;s:49:"lead.form.specific_documentations.protected_adult";} 
```

The ability to filter on 
- Contains : "lead.form.specific_documentations.older_than_85".
- Contains : "lead.form.specific_documentations.older_than_85" and "lead.form.specific_documentations.protected_adult"
- Contains : "lead.form.specific_documentations.older_than_85" and "lead.form.specific_documentations.protected_adult"
- Not Contains : "lead.form.specific_documentations.protected_adult"
- Empty
...

Would be great !

VincentLanglet:
## Subject

This filter allow to search for specific value when you use `@ORM\Column(type="array")`.

Let's say you have in database
1 => ['a']
2 => ['b']
3 => ['c']
4 => ['a', 'b']
5 => ['a', 'c']
6 => ['a', 'b', 'c']

If you search
-  CONTAINS `['a']` ; you'll get 1,4,5,6.
-  NOT CONTAINS `['a']` ; you'll get 2,3.
-  EQUAL `['a']` ; you'll get 1.
-  CONTAINS `['a', 'b']` ; you'll get 4,6.
-  NOT CONTAINS `['a', 'b']` ; you'll get 3.
-  EQUAL `['a', 'b']` ; you'll get 4.

Closes #1029

## Changelog

```markdown
### Added
- `ArrayFilter` which supports `@ORM\Column(type="array")`
```

### DIFF
--- a/docs/reference/filter_field_definition.rst
+++ b/docs/reference/filter_field_definition.rst
@@ -38,6 +38,7 @@ For now, only `Doctrine ORM` filters are available:
 * ``Sonata\DoctrineORMAdminBundle\Filter\NumberFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\NumberType`` Form Type,
 * ``Sonata\DoctrineORMAdminBundle\Filter\ModelAutocompleteFilter``: uses ``Sonata\AdminBundle\Form\Type\Filter\ModelAutocompleteType`` form type, can be used as replacement of ``Sonata\DoctrineORMAdminBundle\Filter\ModelFilter`` to handle too many items that cannot be loaded into memory.
 * ``Sonata\DoctrineORMAdminBundle\Filter\StringFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\ChoiceType`` Form Type,
+* ``Sonata\DoctrineORMAdminBundle\Filter\StringListFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\ChoiceType`` Form Type,
 * ``Sonata\DoctrineORMAdminBundle\Filter\DateFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\DateType`` Form Type, renders a date field,
 * ``Sonata\DoctrineORMAdminBundle\Filter\DateRangeFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\DateRangeType`` Form Type, renders a 2 date fields,
 * ``Sonata\DoctrineORMAdminBundle\Filter\DateTimeFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\DateTimeType`` Form Type, renders a datetime field,
@@ -64,6 +65,34 @@ Example
                 ->add('tags', null, [], null, ['expanded' => true, 'multiple' => true]);
         }
     }
+
+StringListFilter
+----------------
+
+This filter is made for filtering on values saved in databases as serialized arrays of strings with the
+``@ORM\Column(type="array")`` annotation. It is recommended to use another table and ``OneToMany`` relations
+if you want to make complex ``SQL`` queries or if your table is too big and you get performance issues but
+this filter can provide some basic queries::
+
+    protected function configureDatagridFilters(DatagridMapper $datagridMapper): void
+    {
+        $datagridMapper
+            ->add('labels', StringListFilter::class, [], ChoiceType::class, [
+                'choices' => [
+                    'patch' => 'patch',
+                    'minor' => 'minor',
+                    'major' => 'major',
+                    'approved' => 'approved',
+                    // ...
+                ],
+                'multiple' => true,
+            ]);
+    }
+
+.. note::
+
+    The filter can give bad results with associative arrays since it is not easy to distinguish between keys
+    and values for a serialized associative array.
 
 ModelAutocompleteFilter
 -----------------------
--- a/docs/reference/filter_field_definition.rst
+++ b/docs/reference/filter_field_definition.rst
@@ -32,18 +32,17 @@ Available filter types
 
 For now, only `Doctrine ORM` filters are available:
 
-* ``Sonata\DoctrineORMAdminBundle\Filter\BooleanFilter``: depends on the ``sonata_type_filter_default`` Form Type, renders yes or no field,
-* ``Sonata\DoctrineORMAdminBundle\Filter\CallbackFilter``: depends on the ``sonata_type_filter_default`` Form Type, types can be configured as needed,
-* ``Sonata\DoctrineORMAdminBundle\Filter\ChoiceFilter``: depends on the ``sonata_type_filter_choice`` Form Type, renders yes or no field,
-* ``Sonata\DoctrineORMAdminBundle\Filter\NumberFilter``: depends on the ``sonata_type_filter_number`` Form Type,
-* ``Sonata\DoctrineORMAdminBundle\Filter\ModelAutocompleteFilter``: uses ``sonata_type_model_autocomplete`` form type, can be used as replacement of ``Sonata\DoctrineORMAdminBundle\Filter\ModelFilter`` to handle too many items that cannot be loaded into memory.
-* ``Sonata\DoctrineORMAdminBundle\Filter\StringFilter``: depends on the ``sonata_type_filter_choice``,
-* ``Sonata\DoctrineORMAdminBundle\Filter\NumberFilter``: depends on the ``sonata_type_filter_choice`` Form Type, renders yes or no field,
-* ``Sonata\DoctrineORMAdminBundle\Filter\DateFilter``: depends on the ``sonata_type_filter_date`` Form Type, renders a date field,
-* ``Sonata\DoctrineORMAdminBundle\Filter\DateRangeFilter``: depends on the ``sonata_type_filter_date_range`` Form Type, renders a 2 date fields,
-* ``Sonata\DoctrineORMAdminBundle\Filter\DateTimeFilter``: depends on the ``sonata_type_filter_datetime`` Form Type, renders a datetime field,
-* ``Sonata\DoctrineORMAdminBundle\Filter\DateTimeRangeFilter``: depends on the ``sonata_type_filter_datetime_range`` Form Type, renders a 2 datetime fields,
-* ``Sonata\DoctrineORMAdminBundle\Filter\ClassFilter``: depends on the ``sonata_type_filter_default`` Form type, renders a choice list field.
+* ``Sonata\DoctrineORMAdminBundle\Filter\BooleanFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\DefaultType`` Form Type, renders yes or no field,
+* ``Sonata\DoctrineORMAdminBundle\Filter\CallbackFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\DefaultType`` Form Type, types can be configured as needed,
+* ``Sonata\DoctrineORMAdminBundle\Filter\ChoiceFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\ChoiceType`` Form Type,
+* ``Sonata\DoctrineORMAdminBundle\Filter\NumberFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\NumberType`` Form Type,
+* ``Sonata\DoctrineORMAdminBundle\Filter\ModelAutocompleteFilter``: uses ``Sonata\AdminBundle\Form\Type\Filter\ModelAutocompleteType`` form type, can be used as replacement of ``Sonata\DoctrineORMAdminBundle\Filter\ModelFilter`` to handle too many items that cannot be loaded into memory.
+* ``Sonata\DoctrineORMAdminBundle\Filter\StringFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\ChoiceType`` Form Type,
+* ``Sonata\DoctrineORMAdminBundle\Filter\DateFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\DateType`` Form Type, renders a date field,
+* ``Sonata\DoctrineORMAdminBundle\Filter\DateRangeFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\DateRangeType`` Form Type, renders a 2 date fields,
+* ``Sonata\DoctrineORMAdminBundle\Filter\DateTimeFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\DateTimeType`` Form Type, renders a datetime field,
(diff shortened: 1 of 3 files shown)


src/Filter/ArrayFilter.php
```diff
@@ -0,0 +1,62 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DoctrineORMAdminBundle\Filter;
+
+use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
+use Sonata\AdminBundle\Form\Type\Filter\ChoiceType;
+use Sonata\AdminBundle\Form\Type\Operator\ContainsOperatorType;
+
+class ArrayFilter extends Filter
```
core23, line 20:
```suggestion
final class ArrayFilter extends Filter
```

VincentLanglet: Done @core23 :)

core23:
If I got you wrong, you have to input the string `['a']` to get any result. This is very technical for a normal user.

franmomu:
IMHO we should not support filter for a `ORM\Column(type="array")` field, this kind of field (`serialize`) is not made for queries in SQL, it is quite complex and users can get confused. 

IIUC the implementation, it only works with plain arrays of strings, but this field could contain basically anything.

If someone wants to query a field like that, IMHO should consider using another type field or mecanism.

VincentLanglet:
> If I got you wrong, you have to input the string `['a']` to get any result. This is very technical for a normal user.

No it works like the ChoiceFilter, you can provide an array or a single element, since there is the following code:
```
if (!\is_array($data['value'])) {
    $data['value'] = [$data['value']];
}
```

You can use it
```
->add('foo', ArrayFilter)
```
Or
```
->add('foo', ArrayFilter, [], ChoiceType, [ ... ])
```

> IIUC the implementation, it only works with plain arrays of strings, but this field could contain basically anything.

This dont works with strings only. For example if you look for the value `true`, it will look for `b:1;`.

> IMHO we should not support filter for a `ORM\Column(type="array")` field, this kind of field (`serialize`) is not made for queries in SQL
> If someone wants to query a field like that, IMHO should consider using another type field or mecanism.

The `ORM\Column(type="array")` is may not the best to make query but if someone use it, he has his reasons. Developers doesn't always have the hand over the Database.
IMHO we're not here to decide what should be done or not, we should support every `ORM\Column type` if possible.

This PR close an opened issue, which was approved by others, and we have a working solution to provide.

> it is quite complex and users can get confused.

This only means that we need to document it here https://sonata-project.org/bundles/doctrine-orm-admin/master/doc/reference/filter_field_definition.html.

I'll add documentation.

greg0ire:
Please do atomic commits, right now you have one commit for the new feature, one commit for its documentation, and that second commit contains unrelated things, and then there is one more commit about improving documentation that does not say why there is an improvement or even what that improvement is. `git reset origin/3.x && git add --patch` should help with this.

VincentLanglet:
Should be better now @greg0ire

greg0ire:
> Should be better now @greg0ire

It's a bit better indeed, for instance the 3rd commit is ok, apart from one thing; it should be first, and all the other commits should be squashed together, because it makes no sense to separate them:
- reverting the code without reverting the tests makes no sense, and vice versa, and same with docs
- every piece of code is _supposed_ to come with tests and docs, it's only a separate _type_ of things, but it is part of your logical change.

VincentLanglet:
> Gave this a technical review, but maybe concerns raised by @franmomu are justified?
> 
> > Developers doesn't always have the hand over the Database.
> 
> Surely if you develop such an application you can request a column to be added or things like that, can't you?
> 
> > The ORM\Column(type="array") is may not the best to make query but if someone use it, he has his reasons.
> 
> I think that happened to me in the distant past, and I think the reason was I did not have the filtering requirement from the past, so I ended up either using `LIKE` as you did or, when I had the time and motivation, move to some a separate enum like entity.

It's indeed easily possible to add new column, but not always easy to migrate some existing columns when the `ORM\Column(type="array")` was already used. Not everybody can find time and motivation to do the migration. It's harder if you're not the one managing the database.

> It does work but I didn't contribute it because it has a few implications, like what if the user saves an associative array? What if the user uses values or keys that are literally `s:`? Maybe there should be some restrictions, to avoid generating support.

I tried a general approach without too much context in the filter and be as far as possible usable and re-usable. But I'm ok with the fact to add restrictions. How would you add them ?

I can also change the name of the filter, if it helps. Since the most useful cases are when you use it with an array of string, I change to something like `ArrayStringFilter`

> There are also performance implications, this will not perform well on big tables because of the `LIKE %` makes indices unusable.
> 
> What I'm trying to say is that giving power to developers, "treating them like adults" like some people like to say, can lead to weird bugs/performance issues that can generate a lot of support. 

I was more like "This is already useful for me, this will certainly be useful for others, let's share it. Plus thanks to open-source and contribution, this will be permanently improved other time.". But I understand your point.

> I'm a bit undecided on this feature, but one thing I am sure of is that this shouldn't be merged without at least big fat warnings in the documentation, and requires maybe even some restrictions in the code.

I'm open to any "big fat warnings" or "restrictions" to add. @greg0ire ;)

greg0ire:
> How would you add them ?

I would maybe disallow searching for characters used for serialization, like `:` or `"`

> `ArrayStringFilter`

`StringListFilter` filter? See https://twitter.com/Ocramius/status/1252560762238894080

> I'm open to any "big fat warnings" or "restrictions" to add. @greg0ire ;)

Then please do. Some can be forbidden in the code directly, like the above, some others should be warned about: "You should use this if you can't have that list in a separate table, and the list is not used with a table that is too big, and your list consist only in simple strings that don't contain metacharacters like `"` or `:`. Even then be wary that this will give confusing results if your lists consists in elements like `s`, `b`,…"

VincentLanglet:
> > How would you add them ?
> 
> I would maybe disallow searching for characters used for serialization, like `:` or `"`

`:` is search in the array as `s:1:":";`, so there should be no problems, same for `"`.
I'm not sure there is issue with these values.
 
> > `ArrayStringFilter`
> 
> `StringListFilter` filter? See https://twitter.com/Ocramius/status/1252560762238894080

I do like `StringListFilter`.

> Even then be wary that this will give confusing results if your lists consists in elements like `s`, `b`,…". 

Since I search for the serialize value `s:1:"b";`, this should work correctly.

> Also, you should maybe warn about associative array (I don't know if it's possible to save `["foo" => "bar"]` with this ORM type, but from the docs alone it seems possible).

Yes I need to add a warning indeed.

src/Filter/ArrayFilter.php
```diff
@@ -0,0 +1,62 @@
...
+        $andConditions = $queryBuilder->expr()->andX();
+        foreach ($data['value'] as $value) {
+            $parameterName = $this->getNewParameterName($queryBuilder);
+            $andConditions->add(sprintf('%s.%s %s :%s', $alias, $field, $operator, $parameterName));
+
+            $queryBuilder->setParameter($parameterName, '%'.serialize($value).'%');
+        }
+
+        if (ContainsOperatorType::TYPE_EQUAL === $data['type']) {
+            $andConditions->add(sprintf('%s.%s LIKE \'a:%s:%%\'', $alias, $field, \count($data['value'])));
```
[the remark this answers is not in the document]
VincentLanglet: Hum, maybe, don't know, how and what ? 😅

Currently the `StringFilter` doesn't seems to escape something.

franmomu:
You can save any serializable thing:

```php
<?php

$object = new \StdClass();
$object->a = [
    's' => 's:1:"s"'
];
$object->s = 's:1:"s"';

$data = [
    $object,
    ['s' => 's:1:"s"'],
    ['s' => [
        's' => 's'
    ]]
];
```
resulting in:

```
a:3:{i:0;O:8:"stdClass":2:{s:1:"a";a:1:{s:1:"s";s:7:"s:1:"s"";}s:1:"s";s:7:"s:1:"s"";}i:1;a:1:{s:1:"s";s:7:"s:1:"s"";}i:2;a:1:{s:1:"s";a:1:{s:1:"s";s:1:"s";}}}
```

That's why I said it could be complex. What I meant is what @greg0ire said, doing this could lead to more issues (as it is not a full solution). In the case of a list of string I guess it makes more sense to use `simple_array` type.

VincentLanglet:
> In the case of a list of string I guess it makes more sense to use `simple_array` type.

`array` is still more robust than `simple_array` for list of string, since you cannot use `,` in your values saved in the array to avoid conflict with the `implode/explode`.

docs/reference/filter_field_definition.rst
```diff
@@ -66,8 +66,27 @@ Example
         }
     }
 
-doctrine_orm_model_autocomplete
--------------------------------
+ArrayFilter
+-----------
+
+This filter is made to filtering on values saved in databases as a serialized arrays of string with the
```
greg0ire, line 72:
```suggestion
This filter is made for filtering on values saved in databases as serialized arrays of strings with the
```

docs/reference/filter_field_definition.rst
```diff
@@ -66,8 +66,27 @@ Example
...
+``@ORM\Column(type="array")`` annotation. It is recommended to use another table and ``OneToMany`` relations
+if you want to make complex ``SQL`` queries or if your table is too big and you get performance issue but
```
greg0ire, line 74:
```suggestion
if you want to make complex ``SQL`` queries or if your table is too big and you get performance issues but
```

docs/reference/filter_field_definition.rst
```diff
@@ -66,8 +66,27 @@ Example
...
+    protected function configureDatagridFilters(DatagridMapper $datagridMapper): void
+    {
+        $datagridMapper
+            ->add('keywords', ArrayFilter::class);
+    }
+
+.. note::
+
+    The filter can give bad results with associative arrays since it is not easy to distinguish between keys
+    and values for a serialized associative array.
```
greg0ire, line 86: Add also a warning about values that are `s` or `b` or whatever can confuse this? BTW, maybe this would be less risky with json…

VincentLanglet: I don't understand why `s` or `b` can confuse this @greg0ire. 
I serialize the value you're looking for. So if you're looking for `s`, it'll search
`"s:1:"s";"` this can't be confused with a `s` in a bigger word or the `s` used by the serialize array to say `string`.

The only possible confusion is when you store string which are already contain serialize value like storing the array `['s:4:"toto";', 'toto']`. But this seems to be a really weird case. Should I mention it ?

It does not seems to me easy to migrate a `@ORM\Column(type="array")` to a `@ORM\Column(type="json")`, isn't it ? So I would prefer to offer a filter for `array` and a filter for `json` (or the same filter with a `type` option).

VincentLanglet:
@greg0ire Is this filter now technically OK for you ?
Can we discuss about the pro/con about adding this filter ?

docs/reference/filter_field_definition.rst
greg0ire: Oh right, `s` and `b` look safe to search for indeed, my bad.

tests/Filter/StringListFilterTest.php
```diff
@@ -0,0 +1,151 @@
...
+        $filter->filter($builder, 'alias', 'field', ['value' => 'asd', 'type' => ContainsOperatorType::TYPE_EQUAL]);
+        $this->assertSame(['alias.field LIKE :field_name_0 AND alias.field LIKE \'a:1:%\''], $builder->query);
+        $this->assertSame(['field_name_0' => '%s:3:"asd";%'], $builder->parameters);
+        $this->assertTrue($filter->isActive());
+    }
+
+    /**
+     * @dataProvider multipleValuesDataProvider
+     */
+    public function testMultipleValues(array $value, ?int $type, array $query, array $parameters): void
```
greg0ire, line 105: Please describe array elements with phpdoc

docs/reference/filter_field_definition.rst
```diff
@@ -32,18 +32,18 @@ Available filter types
...
-* ``Sonata\DoctrineORMAdminBundle\Filter\StringFilter``: depends on the ``sonata_type_filter_choice``,
-* ``Sonata\DoctrineORMAdminBundle\Filter\NumberFilter``: depends on the ``sonata_type_filter_choice`` Form Type, renders yes or no field,
-* ``Sonata\DoctrineORMAdminBundle\Filter\DateFilter``: depends on the ``sonata_type_filter_date`` Form Type, renders a date field,
-* ``Sonata\DoctrineORMAdminBundle\Filter\DateRangeFilter``: depends on the ``sonata_type_filter_date_range`` Form Type, renders a 2 date fields,
-* ``Sonata\DoctrineORMAdminBundle\Filter\DateTimeFilter``: depends on the ``sonata_type_filter_datetime`` Form Type, renders a datetime field,
-* ``Sonata\DoctrineORMAdminBundle\Filter\DateTimeRangeFilter``: depends on the ``sonata_type_filter_datetime_range`` Form Type, renders a 2 datetime fields,
-* ``Sonata\DoctrineORMAdminBundle\Filter\ClassFilter``: depends on the ``sonata_type_filter_default`` Form type, renders a choice list field.
+* ``Sonata\DoctrineORMAdminBundle\Filter\ArrayFilter``: depends on the ``Sonata\AdminBundle\Form\Type\Filter\ChoiceType`` Form Type,
```
core23, line 35: This is right? Your commit introduces a StringListFilter

docs/reference/filter_field_definition.rst
```diff
@@ -66,8 +66,27 @@ Example
...
+This filter is made for filtering on values saved in databases as serialized arrays of strings with the
+``@ORM\Column(type="array")`` annotation. It is recommended to use another table and ``OneToMany`` relations
+if you want to make complex ``SQL`` queries or if your table is too big and you get performance issues but
+this filter can provide some basic queries::
+
+    protected function configureDatagridFilters(DatagridMapper $datagridMapper): void
+    {
+        $datagridMapper
+            ->add('keywords', ArrayFilter::class);
+    }
```
core23, line 81: Please give some examples how to fill the filter. What are correct user inputs

docs/reference/filter_field_definition.rst
VincentLanglet: Indeed, it's fixed.

docs/reference/filter_field_definition.rst
VincentLanglet: I updated the example, is it ok for you now ?

PR #1046, opened 2020-05-18
Default to lastPage when pager is looking for a non-existent page

VincentLanglet:
## Subject

Closes https://github.com/sonata-project/SonataAdminBundle/issues/6090

## Changelog

```markdown
### Changed
- When looking for a non-existent page in the listAdmin, the datagrid is now displaying the last page instead of displaying a "No result found".
```

### DIFF
--- a/src/Datagrid/Pager.php
+++ b/src/Datagrid/Pager.php
@@ -72,19 +72,20 @@ class Pager extends BasePager
 
         $this->setNbResults($this->computeNbResult());
 
-        $this->getQuery()->setFirstResult(null);
-        $this->getQuery()->setMaxResults(null);
-
         if (\count($this->getParameters()) > 0) {
             $this->getQuery()->setParameters($this->getParameters());
         }
 
         if (0 === $this->getPage() || 0 === $this->getMaxPerPage() || 0 === $this->getNbResults()) {
             $this->setLastPage(0);
+            $this->getQuery()->setFirstResult(null);
+            $this->getQuery()->setMaxResults(null);
         } else {
-            $offset = ($this->getPage() - 1) * $this->getMaxPerPage();
-
             $this->setLastPage((int) ceil($this->getNbResults() / $this->getMaxPerPage()));
+
+            // Offset should be computed after we set the last page
+            // because the setLastPage method can modify the getPage return value
+            $offset = ($this->getPage() - 1) * $this->getMaxPerPage();
 
             $this->getQuery()->setFirstResult($offset);
             $this->getQuery()->setMaxResults($this->getMaxPerPage());
--- a/tests/Datagrid/PagerTest.php
+++ b/tests/Datagrid/PagerTest.php
@@ -100,4 +100,34 @@ class PagerTest extends TestCase
         $pager->setQuery($proxyQuery);
         $pager->computeNbResult();
     }
+
+    /**
+     * @dataProvider initDataProvider
+     */
+    public function testInit(int $computedNbResult, int $page, int $expectedFirstResult): void
+    {
+        $query = $this->createMock(ProxyQuery::class);
+        $pager = $this->createPartialMock(Pager::class, ['computeNbResult']);
+
+        $pager->method('computeNbResult')->willReturn($computedNbResult);
+
+        $pager->setMaxPerPage(10);
+        $pager->setPage($page);
+        $pager->setQuery($query);
+
+        $query->expects($this->once())->method('setFirstResult')->with($expectedFirstResult);
+        $query->expects($this->once())->method('setMaxResults')->with(10);
+
+        $pager->init();
+    }
+
+    public function initDataProvider(): iterable
+    {
+        return [
+            [50, 3, 20],
+            [50, 10, 40],
+            [15, 3, 10],
+            [15, 1, 0],
+        ];
+    }
 }


greg0ire:
I would expect this to be solved with a https://httpstatuses.com/303 , the description matches this use case perfectly: 

> A 303 response to a GET request indicates that the origin server does not have a representation of the target resource that can be transferred by the server over HTTP. However, the Location field value refers to a resource that is descriptive of the target resource, such that making a retrieval request on that other resource might result in a representation that is useful to recipients without implying that it represents the original target resource.

Also, it would be best to have the same behavior across bundles, so IMO there should be a common exception that persistence bundles could throw for this (`OutOfRangeException`?)

We wouldn't have these UX issues if we were using [cursor based pagination](https://use-the-index-luke.com/no-offset), but that's another story…

VincentLanglet:
> I would expect this to be solved with a https://httpstatuses.com/303 , the description matches this use case perfectly:

My issue with HttpException is they doesn't help the user to get away from the trouble.

If you're on this 303 page, how can you fix the selected page ? They are "normal" way to access a non-existent page. For example refreshing the last page after deleting some rows. With a 303 they'll stuck for ever.
I work with persisted filters. Every time my users will come on the list they'll get the 303 exception because they're out of range, and every time they won't be able to change the page value in a user friendly way.

Plus the exception is a BC-break.

> Also, it would be best to have the same behavior across bundles, so IMO there should be a common exception that persistence bundles could throw for this (`OutOfRangeException`?)

The same PR can be done for others bundles in order to get the same behaviour. I didn't find a way to make the PR directly in the AdminBundle.
IMHO the adminBundle already took a decision about this issue since the `setPage` method is changing the page value. And the ORMBundle was just not implementing correctly the Pager.

greg0ire:
Why do you keep talking about "303 exception"? 3xx status codes are for redirections. I meant to have an exception caught somewhere else (admin bundle?) then do that redirection.

greg0ire:
> Merging this can be a way to wait.

It's arguably a BC-break too though, isn't it?

VincentLanglet:
> > Merging this can be a way to wait.
> 
> It's arguably a BC-break too though, isn't it?

Personally, I see this as a bugfix.
It's always arguably. Every bugfix is a BC-break since the behaviour is changing.

Let's say there are 20 totals results, 10 results per page, and you're looking for page 5.

When you write 
```
$offset = ($this->getPage() - 1) * $this->getMaxPerPage();	
$this->setLastPage((int) ceil($this->getNbResults() / $this->getMaxPerPage()));
```

The offset is 40, computed from page 5, and the lastPage is 2.
Now everytime I try to access to the page of the pager, it returns 2. I have no way to compute the 40 offset, if I want to display it somewhere.

The datagrid will display a message "No result" without displaying the list of available pages and without giving me the ability to go to an existent page. Like there was 0 result.

This behaviour is a bug IMHO. A php bug and a UX bug.

When you write 
```
$this->setLastPage((int) ceil($this->getNbResults() / $this->getMaxPerPage()));
$offset = ($this->getPage() - 1) * $this->getMaxPerPage();
```

Now the offset is 10, computed from page 2, and the lastPage is 2.
If I want to compute again the offset, I can get the same value.

The datagrid will act as if I ask for page 2. Which is consistent with the fact that the pager has the page 2 as `page` value.

greg0ire:
> Personally, I see this as a bugfix.

Well, I see the redirection as a bugfix too. The exception would be a BC-break though. But do we really need it to implement the required logic? I don't think we actually do:

```
if not first page and no results
then redirect to $storage->getLastPage()
```

VincentLanglet:
> But do we really need it to implement the required logic? I don't think we actually do:
> 
> ```
> if not first page and no results
> then redirect to $storage->getLastPage()
> ```

Sorry, I'm not sure to understand what you mean here.
Are you saying that we can implement the redirect in the SonataAdminBundle without introducing the exception ? And then introduce the exception in 4.0.

Yes, this could be doable.

But I'm not sure there is need for a debate since I think both should be done. Fixing the Datagrid/Controller and fixing the implementation of the Pager. Because we can use one of them without the other. 

The way the Pager was written is not correct with the way the AbstractPager was designed.
Either we should swap the lines like I did in the PR, either the `setLastPage` shouldn't modify the `page` value.
I chose the first solution because it's BC, as a bonus it solved the UX issue, and I had no idea about how big are the impact of the second solution. What do you prefer between these choices ?

VincentLanglet:
The fact that if you request for the page after trying to get the results, you'll get the lastPage value can lead to some tricky situation.

For example, if you have 5 results, with 10 results per page.

If you request the page 3, you gonna make the request with the offset 20. Then the last page will be 1. Then the page will be 1.

If you now you check
```
if not first page and no results
then redirect to $storage->getLastPage()
```
You won't be redirect, since the Pager is thinking you're page 1.

So, just modifying the controller won't be enough.
Plus it could make two query instead of one.

Datagrid getResults is calling 
```
$this->buildPager

$this->pager->getResult()
```

Datagrid buildPager is calling
```
$this->pager->setPage($page);
$this->pager->setQuery($this->query);
$this->pager->init();
```

And pager init method
```
$this->computeNbResult();
...
$this->setLastPage();
// offset, maxResult, ...
```

When I looked again the code, it seems like there is no accessible `lastPage` or `nbResult` property in the Pager. So we'll may need to add some methods in the PagerInterface.

I'm glad if I'm wrong and you can lead me in the SonataAdminBundle PR, but I really think it's
- either this PR in the doctrine bundles
- either a weird fix in the sonata admin 3.x which won't work every time or make extra queries.
- either a proper fix in the sonata admin 4.0 which is not easy and need multiple BC-break like throwing an exception, modifying the interface, etc.

And even if we chose solution 2 or 3, this PR is not needed only if we decide to remove the `page` value update in the `setLastPage` method.

greg0ire:
>Are you saying that we can implement the redirect in the SonataAdminBundle without introducing the exception ? And then introduce the exception in 4.0.

That's exactly what I was saying, yeah.


> The way the Pager was written is not correct with the way the AbstractPager was designed.

Is it? Please elaborate on that.

> The fact that if you request for the page after trying to get the results, you'll get the lastPage value can lead to some tricky situation.

Are you talking about the change in this PR, or the http redirection solution?

> If you request the page 3, you gonna make the request with the offset 20. Then the last page will be 1. Then the page will be 1.

:thinking: the last page should be 2, shouldn't it?

> You won't be redirect, since the Pager is thinking you're page 1.


>  Plus it could make two query instead of one.

This out-of-range scenario is definitely not a performance critical scenario.
:thinking: if you're on page 3, why would it think you're on page 1?

> When I looked again the code, it seems like there is no accessible lastPage or nbResult property in the Pager. 

Really?

https://github.com/sonata-project/SonataAdminBundle/blob/26e64c23b3ea49ed5c168617cac7160722dc78cb/src/Datagrid/Pager.php#L329-L332

https://github.com/sonata-project/SonataAdminBundle/blob/26e64c23b3ea49ed5c168617cac7160722dc78cb/src/Datagrid/Pager.php#L313-L316

VincentLanglet:
> > The way the Pager was written is not correct with the way the AbstractPager was designed.
> 
> Is it? Please elaborate on that.

I'm talking about the fact `setLastPage` modify the `getPage` return value and the fact we use the old `getPage` value, instead of the new one by calling `getPage` before `setLastPage` and not after.

> > The fact that if you request for the page after trying to get the results, you'll get the lastPage value can lead to some tricky situation.
> 
> Are you talking about the change in this PR, or the http redirection solution?

I'm talking about a general way.

If I override the template, and try to access `pager.page`.
Or If I override the Datagrid, or a function in the admin and try to call get Page.

> > If you request the page 3, you gonna make the request with the offset 20. Then the last page will be 1. Then the page will be 1.
> 
> 🤔 the last page should be 2, shouldn't it?

I gave an example with 5 total results and 10 results per page.
So page one will show result 1, 2, 3, 4, 5. There is only one page, so lastPage is 1.

> > Plus it could make two query instead of one.
> 
> This out-of-range scenario is definitely not a performance critical scenario.

I know nothing about performance.
Is a `COUNT(*)` or a too big offset, always fast ?
Even if it's on a table with 1 000 000 rows with complex query ?

> > You won't be redirect, since the Pager is thinking you're page 1.
> 
> 🤔 if you're on page 3, why would it think you're on page 1?

From AbstractPager
```
    /**
     * @param int $page
     */
    protected function setLastPage($page)
    {
        $this->lastPage = $page;

        if ($this->getPage() > $page) {
            $this->setPage($page);
        }
    }
```

So `getPage()` was returning 3.
But as soon as, you call `setLastPage(1)`, `getPage()` will now return 1.

You made the query for the page 3, but now the Pager say you're page 1.

If you swap the line, like in my PR, you say the last page (and the page) before making the query so you make the query for page 1.
The other solution I gave was to remove the `setPage` call in the `setLastPage` method. But it's a BC-break and we'll have to wait the 4.0 version.

Both can be made. Swapping the line for the quick fix (3.x), and removing the `setPage` call for the long term fix (4.0 +).

> > When I looked again the code, it seems like there is no accessible lastPage or nbResult property in the Pager.
> 
> Really?
> 
> https://github.com/sonata-project/SonataAdminBundle/blob/26e64c23b3ea49ed5c168617cac7160722dc78cb/src/Datagrid/Pager.php#L329-L332
> 
> https://github.com/sonata-project/SonataAdminBundle/blob/26e64c23b3ea49ed5c168617cac7160722dc78cb/src/Datagrid/Pager.php#L313-L316

You're quoting `abstract class Pager`.

But `DatagridInterface` returns a `PagerInterface`
https://github.com/sonata-project/SonataAdminBundle/blob/26e64c23b3ea49ed5c168617cac7160722dc78cb/src/Datagrid/DatagridInterface.php#L29-L32

And the `PagerInterface` doesn't have these functions
https://github.com/sonata-project/SonataAdminBundle/blob/26e64c23b3ea49ed5c168617cac7160722dc78cb/src/Datagrid/PagerInterface.php

VincentLanglet:
> You're quoting `abstract class Pager`.
> 
> But `DatagridInterface` returns a `PagerInterface`
> https://github.com/sonata-project/SonataAdminBundle/blob/26e64c23b3ea49ed5c168617cac7160722dc78cb/src/Datagrid/DatagridInterface.php#L29-L32
> 
> And the `PagerInterface` doesn't have these functions
> https://github.com/sonata-project/SonataAdminBundle/blob/26e64c23b3ea49ed5c168617cac7160722dc78cb/src/Datagrid/PagerInterface.php

BTW

SimplePager has `getNbResults` and `getLastPage` too.
https://github.com/sonata-project/SonataAdminBundle/blob/26e64c23b3ea49ed5c168617cac7160722dc78cb/src/Datagrid/SimplePager.php

Updating the PagerInterface for 4.0 doesn't seems a bad idea.

greg0ire:
> I know nothing about performance.
> Is a `COUNT(*)` or a too big offset, always fast ?
> Even if it's on a table with 1 000 000 rows with complex query ?

Ah that must be why `getLastPage()` is not part of the base interface.

VincentLanglet:
> > I know nothing about performance.
> > Is a `COUNT(*)` or a too big offset, always fast ?
> > Even if it's on a table with 1 000 000 rows with complex query ?
> 
> Ah that must be why `getLastPage()` is not part of the base interface.

The simplePager still expose a `getLastPage` with an approximation.

I think the Interface is missing some function without reason.

In the twig we're calling
`pager.getResults()`,
`pager.getNbResults()`,
`pager.count()`
`admin.datagrid.pager.lastPage`, 
`admin.datagrid.pager.haveToPaginate()`,
`admin.datagrid.pager.page`,
`admin.datagrid.pager.previouspage`,
`admin.datagrid.pager.getLinks()`, 
`admin.datagrid.pager.nextpage`

So all these function should be in the interface.

greg0ire:
>  the setLastPage shouldn't modify the page value.

That's what makes all this so hard to understand. IMO the page the user requested should be immutable. And I would like a redirection because I think we should not return the last page when the user asked for a page that does not exist. We should be explicit about the requested page not existing, and the last page being a better choice.

greg0ire:
> I think the Interface is missing some function without reason.

Are you sure it's not about performance?

VincentLanglet:
> > I think the Interface is missing some function without reason.
> 
> Are you sure it's not about performance?

The SimplePager is exposing all these function even if the performance are improved.
And since the twig is using these function, the pager has to expose these functions.

We created an Interface, but in the twig, we're doing like it's an AbstractPager instead of doing it's like a PagerInterface.

greg0ire:
>  Even if it's on a table with 1 000 000 rows with complex query ?

If it's a complex query (with filters), and you don't have proper indices on your DB, it can definitely be slow.

greg0ire:
> We created an Interface, but in the twig, we're doing like it's an AbstractPager instead of doing it's like a PagerInterface.

Oh that's a real bummer…

greg0ire:
That being said, is all this only ever used through Twig? I think there are a few bundles that come with APIs: https://github.com/sonata-project/SonataUserBundle/tree/4.x/src/Controller/Api

greg0ire:
Also, see https://github.com/sonata-project/SonataDatagridBundle/pull/62

VincentLanglet:
> > We created an Interface, but in the twig, we're doing like it's an AbstractPager instead of doing it's like a PagerInterface.
> 
> Oh that's a real bummer…

Yes, I don't know if I should make the PR in 4.0 or this situation can allow a PR in 3.x

> That being said, is all this only ever used through Twig? I think there are a few bundles that come with APIs: https://github.com/sonata-project/SonataUserBundle/tree/4.x/src/Controller/Api

This bundle is using `Sonata\DatagridBundle\Pager\PagerInterface`, which is not `Sonata\AdminBundle\Datagrid\PagerInterface`.

> Also, see [sonata-project/SonataDatagridBundle#62](https://github.com/sonata-project/SonataDatagridBundle/pull/62)

Seems like the `Sonata\DatagridBundle\Pager\PagerInterface` interface updated but not `Sonata\AdminBundle\Datagrid\PagerInterface`

I discovered few weeks ago that both `DatagridBundle` and `AdminBundle` are exposing `Pager` and `Datagrid` really similar but with few differences (because of different PR over time).

That's how I had the idea of a repository with all interfaces. To avoid this situation.

VincentLanglet:
> > > We created an Interface, but in the twig, we're doing like it's an AbstractPager instead of doing it's like a PagerInterface.
> > 
> > 
> > Oh that's a real bummer…
> 
> Yes, I don't know if I should make the PR in 4.0 or this situation can allow a PR in 3.x
> 
> > That being said, is all this only ever used through Twig? I think there are a few bundles that come with APIs: https://github.com/sonata-project/SonataUserBundle/tree/4.x/src/Controller/Api
> 
> This bundle is using `Sonata\DatagridBundle\Pager\PagerInterface`, which is not `Sonata\AdminBundle\Datagrid\PagerInterface`.
> 
> > Also, see [sonata-project/SonataDatagridBundle#62](https://github.com/sonata-project/SonataDatagridBundle/pull/62)
> 
> Seems like the `Sonata\DatagridBundle\Pager\PagerInterface` interface updated but not `Sonata\AdminBundle\Datagrid\PagerInterface`
> 
> I discovered few weeks ago that both `DatagridBundle` and `AdminBundle` are exposing `Pager` and `Datagrid` really similar but with few differences (because of different PR over time).
> 
> That's how I had the idea of a repository with all interfaces. To avoid this situation. But I didn't really know how to start.

greg0ire:
> Yes, I don't know if I should make the PR in 4.0 or this situation can allow a PR in 3.x

Adding to an existing interface is not ok. Creating a new interface and having all implementations also implement it is though.

I don't know why we have 2 interfaces in 2 bundles and what is the plan regarding those. Maybe @core23 does?

VincentLanglet:
> > Yes, I don't know if I should make the PR in 4.0 or this situation can allow a PR in 3.x
> 
> Adding to an existing interface is not ok. Creating a new interface and having all implementations also implement it is though.

I agree. 
But here the fact is that if you just implement the interface, without extra function, the twig are throwing errors. That's why I asked myself the question.

> I don't know why we have 2 interfaces in 2 bundles and what is the plan regarding those. Maybe @core23 does?

DatagridBundle was made to split the AdminBundle, but the project stoped half way and now they took different way which makes things harder.

greg0ire:
> DatagridBundle was made to split the AdminBundle, but the project stoped half way and now they took different way which makes things harder.

Ok :sweat_smile: maybe start by fixing that, since it makes everything more complex?

VincentLanglet:
> > DatagridBundle was made to split the AdminBundle, but the project stoped half way and now they took different way which makes things harder.
> 
> Ok 😅 maybe start by fixing that, since it makes everything more complex?

How ? It could lead to weeks of development with multiple BC-break.

DatagridInterface from AdminBundle has extra methods hasDisplayableFilters.

It will also have getSortParameters, getPaginationParameters in the next major, which use the `FieldDescriptionInterface`. We can't move these function in the DatagridBundle or it will lead to a circular reference (and we can't move the `FieldDescriptionInterface` because it require the AdminInterface etc...)

PagerInterface from AdminBundle has less methods.
Interface from DatagridBundle has typehint.

Etc...

Imho the solution would be
- Wait for the 4.0 release because there are enough changes.
- Creating a repository with all the interface. (which is a lot of work)
- Using these interfaces in the bundles, in 4.x or in master (futur 5.0) and deprecate the old interface. (which is a lot of work)
- Release 5.0 and remove the old interface.

So I can't consider this as the solution for now.

VincentLanglet:
Another and easier solution seems to have an `AdminDatagridInterface` extending the `DatagridInterface` from the DatagridBundle.

For others classes this could be do-able. I'll try to do something.

phansys:
Reading the discussion about the status code and the expected behavior, I think we could register an exception listener for HTTP exceptions (4xx, 5xx). This way, we should be able to expose the real situation in a semantic fashion (from a pure HTTP perspective), and then, decide where to redirect the request (if required) and what to do later (with a flash message or something silmilar).

VincentLanglet:
Closing in favor of https://github.com/sonata-project/SonataAdminBundle/pull/6500

PR #1055, opened 2020-05-31
Improve Template mapping

VincentLanglet:
## Subject

There is no template for types like `number` or `orm_one_to_many`.
So I added a mapping between wrong types and right types to use, only if the developer didn't added a template.

## Changelog

```markdown
### Changed
- `decimal` and `float` type use the `float` template if no `number` template exists

### Fixed
- `one_to_one`, `one_to_many`, `many_to_one` and `many_to_many` type are correctly using the template defined in your config instead of the Sonata one.
```

### DIFF
--- a/src/Guesser/TypeGuesser.php
+++ b/src/Guesser/TypeGuesser.php
@@ -15,11 +15,29 @@ namespace Sonata\DoctrineORMAdminBundle\Guesser;
 
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Sonata\AdminBundle\Model\ModelManagerInterface;
+use Sonata\AdminBundle\Templating\TemplateRegistry;
 use Symfony\Component\Form\Guess\Guess;
 use Symfony\Component\Form\Guess\TypeGuess;
 
 class TypeGuesser extends AbstractTypeGuesser
 {
+    /**
+     * This is a mapping between the old deprecated value we provided in the TypeGuesser
+     * and the value we should use to get the correct SonataAdminBundle template. Making the change
+     * directly in the TypeGuesser would be a BC-break if a user overrides the templates config.
+     *
+     * NEXT_MAJOR: remove this constant.
+     *
+     * @internal
+     */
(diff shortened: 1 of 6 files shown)


VincentLanglet:
If you have time to look at @sonata-project/contributors :)

src/Guesser/TypeGuesser.php
```diff
@@ -15,11 +15,30 @@
...
+    /**
+     * This is a mapping between the old deprecated value we provided in the TypeGuesser
+     * and the value we should use to get the correct SonataAdminBundle template. Making the change
+     * directly in the TypeGuesser would be a BC-break if a user overrides the templates config.
+     *
+     * NEXT_MAJOR: remove this constant.
+     *
+     * @internal
+     */
+    public const DEPRECATED_TYPES = [
```
greg0ire, line 33: There is no deprecation triggered in this PR. Can you link to the PR that adds the deprecation? I feel like I don't have all the context needed to understand the PR.

VincentLanglet: This is related to https://github.com/sonata-project/SonataAdminBundle/pull/6123

Do you think something should be added ?
A deprecation message in the `private function getTemplate($type)` method ?

VincentLanglet: My issue is, if I add a deprecation message, the user will received it every time because the Bundle is currently using the deprecated constant. And changing this could be considered as a BC-break.

greg0ire: All the deprecations that can be solved by the user sound good to me :)

VincentLanglet: Should be ok this way.

VincentLanglet:
@greg0ire I don't think so.

This is more SonataAdmin related and there is already an upgrade note https://github.com/sonata-project/SonataAdminBundle/blob/3.x/UPGRADE-3.x.md#upgrade-from-367-to-368

PR #1058, opened 2020-06-11
Update composer.json

wbloszyk:
## Subject

This change will allow install optionally only CoreBundle v3.20 (which constains all deprecations notes).

#### Remove process will be look like:
- update CoreBundle to 3.20
- remove deprecations
- remove CoreBundle
- upgrade extensions to 1.x

I am targeting this branch, because this change respect BC.


jordisala1991:
Thank you @wbloszyk

PR #1061, opened 2020-06-28
Add deprecation for StringFilter format option

VincentLanglet:
## Subject

I already found weird the `format` option of the `StringFilter`.
If I look for something containing `foo`, there is IMHO no multiple way to check for this. It always `LIKE %foo%`.

Now, this option is weirder since we have the `START WITH` and an `END WITH` checks.
Why would these checks be made with `%foo` and `foo%` but the CONTAINS one could use another format than `%foo%` ?

That's why I think we should deprecate the `format` option and remove it in the next major.

## Changelog

```markdown
### Deprecated
- `format` option of the `StringFilter`.
```

### DIFF
--- a/src/Filter/StringFilter.php
+++ b/src/Filter/StringFilter.php
@@ -79,6 +79,13 @@ class StringFilter extends Filter
                     break;
                 default:
                     $format = $this->getOption('format');
+
+                    if ('%%%s%%' !== $format) {
+                        @trigger_error(
+                            'The "format" option is deprecated since sonata-project/doctrine-orm-admin-bundle 3.x and will be removed in version 4.0.',
+                            E_USER_DEPRECATED
+                        );
+                    }
             }
 
             $queryBuilder->setParameter(
@@ -94,6 +101,7 @@ class StringFilter extends Filter
     public function getDefaultOptions()
     {
         return [
+            // NEXT_MAJOR: Remove the format option.
             'format' => '%%%s%%',
             'case_sensitive' => true,
         ];
--- a/tests/Filter/StringFilterTest.php
+++ b/tests/Filter/StringFilterTest.php
@@ -38,7 +38,7 @@ class StringFilterTest extends TestCase
     public function testNullValue(): void
     {
         $filter = new StringFilter();
-        $filter->initialize('field_name', ['format' => '%s']);
+        $filter->initialize('field_name');
 
         $builder = new ProxyQuery(new QueryBuilder());
         $this->assertSame([], $builder->query);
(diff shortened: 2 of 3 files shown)


phansys:
>Why would these checks be made with %foo and foo% but the CONTAINS one could use another format than %foo% ?

I still see the single character mask as a valid use for this option. By instance `LIKE f_o`. Depending on the engine, I think the `LIKE` syntax is able to support even more masks than `%` or `_`.

VincentLanglet:
> > Why would these checks be made with %foo and foo% but the CONTAINS one could use another format than %foo% ?
> 
> I still see the single character mask as a valid use for this option. By instance `LIKE f_o`. Depending on the engine, I think the `LIKE` syntax is able to support even more masks than `%` or `_`.

I didn't know there is others mask. Maybe it's a use case but you will have 
`Contains` -> `_foo_`
`Start`-> `foo%`
`End` -> `%foo`
This is not consistent. A mask option could be better to change `%` to `_`.
`Contains` -> `_foo_`
`Start`-> `foo_`
`End` -> `_foo`
But do you have a use case ?

There shouldn't be multiple way to contain something or start with something.
If you have a special rule, it would be better to have a special filter.

tests/Filter/StringFilterTest.php
```diff
@@ -195,4 +194,22 @@ public function caseSensitiveDataProvider(): array
 
         ];
     }
+
+    /**
+     * @group legacy
+     *
+     * @expectedDeprecation The format option is deprecated since sonata-project/doctrine-orm-admin-bundle 3.x and will be removed in version 4.0.
```
phansys, line 201:
```suggestion
     * @expectedDeprecation The "format" option is deprecated since sonata-project/doctrine-orm-admin-bundle 3.x and will be removed in version 4.0.
```

src/Filter/StringFilter.php
```diff
@@ -79,6 +79,13 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
                     break;
                 default:
                     $format = $this->getOption('format');
+
+                    if ('%%%s%%' !== $format) {
+                        @trigger_error(
+                            'The format option is deprecated since sonata-project/doctrine-orm-admin-bundle 3.x and will be removed in version 4.0.',
```
phansys, line 85:
```suggestion
                            'The "format" option is deprecated since sonata-project/doctrine-orm-admin-bundle 3.x and will be removed in version 4.0.',
```

phansys:
It's enough for me. Just pointing that MSSQL has support for [other masks](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql?view=sql-server-ver15#arguments).

VincentLanglet:
Done @phansys

tests/Filter/StringFilterTest.php
```diff
@@ -195,4 +194,22 @@ public function caseSensitiveDataProvider(): array
 
         ];
     }
+
+    /**
+     * @group legacy
```
phansys, line 199: What to do in the next major about this test?

VincentLanglet: I added a `NEXT_MAJOR` comment.

PR #1068, opened 2020-07-12
Merge 3.x into Master

VincentLanglet:


src/Block/AuditBlockService.php
```diff
@@ -33,13 +34,60 @@ class AuditBlockService extends AbstractBlockService
     protected $auditReader;
 
     /**
-     * @param string $name
+     * NEXT_MAJOR: Allow only Environment|EngineInterface for argument 1 and AuditReader for argument 2.
```
phansys, line 37: Sorry for the off topic, but reading this comment I remembered the state of the audit library.
I think it is currently in a fragile and unmaintained status. See simplethings/EntityAuditBundle#329.
I'll try to keep track for this concern in a proper way.

VincentLanglet: Do you mean this Service should be deprecated ?

phansys: Personally I'd like to keep the service, since I'm using it exhaustively in my projects; but sadly I don't see many choices:
1. Ping the project's maintainers in order to know if there is a clear decision about the future of the project (https://github.com/simplethings/EntityAuditBundle/pull/338#issuecomment-648768688, https://github.com/simplethings/EntityAuditBundle/issues/343#issue-648751967);
2. Find for a replacement;
3. Deprecate and remove the feature.

VincentLanglet: Ok, keeping the service is not an issue.

If we want, we can still deprecate it in the 4.x and remove it in the 5.0 version. We're not in hurry

PR #1067, opened 2020-07-11
Prepare support for BlockBundle 4

VincentLanglet:
## Subject

I am targeting this branch, because BC.

## Changelog

```markdown
### Deprecated
- Passing a string as argument 1 when instantiating Sonata\DoctrineORMAdminBundle\Block\AuditBlockService 
```

PR #1069, opened 2020-07-12
Improve deprecation messages for AuditBlockService

VincentLanglet:
## Subject

### DIFF
--- a/src/Block/AuditBlockService.php
+++ b/src/Block/AuditBlockService.php
@@ -72,7 +72,9 @@ class AuditBlockService extends AbstractBlockService
                     __METHOD__,
                     EngineInterface::class,
                     Environment::class,
-                    \is_object($templatingOrDeprecatedName) ? \get_class($templatingOrDeprecatedName) : \gettype($templatingOrDeprecatedName)
+                    \is_object($templatingOrDeprecatedName)
+                        ? 'instance of '.\get_class($templatingOrDeprecatedName)
+                        : \gettype($templatingOrDeprecatedName)
                 ));
             }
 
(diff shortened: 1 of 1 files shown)


VincentLanglet:
I fixed https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1067

@phansys Please review.

src/Block/AuditBlockService.php
```diff
@@ -44,7 +44,7 @@ public function __construct($templatingOrDeprecatedName, object $templatingOrAud
         if ($templatingOrAuditReader instanceof EngineInterface) {
             @trigger_error(sprintf(
                 'Passing %s as argument 2 to %s() is deprecated since sonata-project/doctrine-orm-admin-bundle 3.x'
-                .' and will throw a \TypeError in version 4.0. You must pass an instance of %s instead.',
+                .' and will throw a \TypeError exception in version 4.0. You must pass an instance of %s instead.',
```
core23, line 47: A type error is an error, not an exception: https://www.php.net/manual/en/class.typeerror.php

phansys:
Thank you!

src/Block/AuditBlockService.php
phansys: I know `\Error` is sibling of `\Exception`, since both of them implements `\Throwable`, but since both can be thrown and catched, I tend to refer both as exceptions (regardless its class names).
To resume, maybe me reasoning can be expressed like this:

|Verb  |Noun       |
|---      |--             |
|trigger|error       |
|throw |exception|

PR #1070, opened 2020-07-12
Add typehint for private methods and improve tests

VincentLanglet:
## Subject

I wanted to add typehint on the master branch, but some can be added on 3.x.
Plus, `willReturn()` throw an error when the return typehint is `void`. So I modified the tests to prevent this issue.

I am targeting this branch, because pedantic.

### DIFF
--- a/src/Builder/FormContractor.php
+++ b/src/Builder/FormContractor.php
@@ -183,10 +183,7 @@ class FormContractor implements FormContractorInterface
         return $options;
     }
 
-    /**
-     * @return bool
-     */
-    private function hasAssociation(FieldDescriptionInterface $fieldDescription)
+    private function hasAssociation(FieldDescriptionInterface $fieldDescription): bool
     {
         return \in_array($fieldDescription->getMappingType(), [
             ClassMetadata::ONE_TO_MANY,
@@ -196,13 +193,7 @@ class FormContractor implements FormContractorInterface
         ], true);
     }
 
-    /**
-     * @param string $type
-     * @param array  $classes
-     *
-     * @return array
-     */
-    private function checkFormClass($type, $classes)
+    private function checkFormClass(string $type, array $classes): array
     {
         return array_filter($classes, static function ($subclass) use ($type) {
             return is_a($type, $subclass, true);
--- a/src/Builder/FormContractor.php
+++ b/src/Builder/FormContractor.php
@@ -98,7 +98,7 @@ class FormContractor implements FormContractorInterface
         $options = [];
         $options['sonata_field_description'] = $fieldDescription;
 
-        if ($this->checkFormClass($type, [
+        if ($this->isAnyInstanceOf($type, [
             ModelType::class,
             ModelTypeList::class,
             ModelListType::class,
@@ -116,7 +116,7 @@ class FormContractor implements FormContractorInterface
             $options['class'] = $fieldDescription->getTargetModel();
             $options['model_manager'] = $fieldDescription->getAdmin()->getModelManager();
 
-            if ($this->checkFormClass($type, [ModelAutocompleteType::class])) {
+            if ($this->isAnyInstanceOf($type, [ModelAutocompleteType::class])) {
                 if (!$fieldDescription->getAssociationAdmin()) {
                     throw new \RuntimeException(sprintf(
                         'The current field `%s` is not linked to an admin.'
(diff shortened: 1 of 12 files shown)


src/Filter/NumberFilter.php
```diff
@@ -62,11 +62,9 @@ public function getRenderSettings()
...
     /**
-     * @param string $type
-     *
-     * @return bool
+     * @return string|false
      */
-    private function getOperator($type)
+    private function getOperator(int $type)
     {
         return self::CHOICES[$type] ?? false;
```
jordisala1991, line 69: isnt it better if this returns the default operator instead of false?

VincentLanglet: Indeed. Updated.

src/Filter/ClassFilter.php
```diff
@@ -39,10 +39,6 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
 
         $operator = $this->getOperator((int) $data['type']);
```
jordisala1991, line 40: This parameter and the line above looks a bit strange compared to the other filters, can we equalize?

src/Filter/StringFilter.php
```diff
@@ -43,10 +43,6 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
 
         $operator = $this->getOperator((int) $data['type']);
```
jordisala1991, line 44: Same here, on the NumberFilter looks better

VincentLanglet:
This way @jordisala1991 ?

src/Builder/FormContractor.php
```diff
@@ -196,13 +193,7 @@ private function hasAssociation(FieldDescriptionInterface $fieldDescription)
...
-    /**
-     * @param string $type
-     * @param array  $classes
-     *
-     * @return array
-     */
-    private function checkFormClass($type, $classes)
+    private function checkFormClass(string $type, array $classes): array
     {
         return array_filter($classes, static function ($subclass) use ($type) {
```
phansys, line 198:
```suggestion
        return array_filter($classes, static function (string $subclass) use ($type): bool {
```

VincentLanglet: The `checkFormClass` return value is only used as truthy/falsy.
So I changed the method.

WDYT ? Should I rename the function too ? `isFormClass` ?

src/Builder/FormContractor.php
```diff
@@ -197,15 +194,16 @@ private function hasAssociation(FieldDescriptionInterface $fieldDescription)
     }
 
     /**
-     * @param string $type
-     * @param array  $classes
-     *
-     * @return array
```
jordisala1991, line 203: This was returning an array, is that change to bool correct?

src/Builder/FormContractor.php
```diff
@@ -197,15 +194,16 @@ private function hasAssociation(FieldDescriptionInterface $fieldDescription)
...
+     * @param string[] $classes
      */
-    private function checkFormClass($type, $classes)
+    private function checkFormClass(string $type, array $classes): bool
```
phansys, line 199: Maybe we could add a description explaining the purpose of this method and how it behaves with valid and invalid values.

VincentLanglet: This method just check if the type is one of the classes provided.

A right name for this method should be enough IMHO.

VincentLanglet: `private function isInstanceOf(string $type, array $classes): bool` @phansys ?

phansys: I'm not convinced since the check will be performed against multiple values and it will return `true` if at least 1 matches, but I can't provide a better name.
My best effort: `containsInstanceOf()`.

VincentLanglet: The method check if `$type` is instance of one of the class provided.
`isInstanceOfOneOf` is more explicit, maybe ?

I'm not convinced by `containsInstanceOf`.
Do you mean that `$classes` contains an instance of `$type` ?

phansys: > The method check if `$type` is instance of one of the class provided.
> `isInstanceOfOneOf` is more explicit, maybe ?

Maybe `isAnyInstanceOf()` / `isTypeInArray()`? We could try to find inspiration at PHPUnit assertions.

> Do you mean that `$classes` contains an instance of `$type` ?

Yes, exactly.

VincentLanglet: I like `isAnyInstanceOf`. I made the change.

PR #1078, opened 2020-07-25
Use propertyAccessor and deprecate camelize

VincentLanglet:
## Subject

Related to https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/pull/368#discussion_r460404367

## Changelog

```markdown

### Deprecated
- deprecate `ModelManager::camelize`.
```

### DIFF
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -36,24 +36,48 @@ use Sonata\DoctrineORMAdminBundle\Admin\FieldDescription;
 use Sonata\DoctrineORMAdminBundle\Datagrid\OrderByToSelectWalker;
 use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery;
 use Sonata\Exporter\Source\DoctrineORMQuerySourceIterator;
-use Symfony\Component\Form\Exception\PropertyAccessDeniedException;
+use Symfony\Component\PropertyAccess\PropertyAccess;
+use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
 
 class ModelManager implements ModelManagerInterface, LockInterface
 {
     public const ID_SEPARATOR = '~';
+
     /**
      * @var ManagerRegistry
      */
     protected $registry;
 
     /**
+     * @var PropertyAccessorInterface
+     */
+    protected $propertyAccessor;
+
+    /**
      * @var EntityManager[]
      */
     protected $cache = [];
 
-    public function __construct(ManagerRegistry $registry)
+    /**
+     * NEXT_MAJOR: Make $propertyAccessor mandatory.
+     */
+    public function __construct(ManagerRegistry $registry, ?PropertyAccessorInterface $propertyAccessor = null)
     {
         $this->registry = $registry;
+
+        // NEXT_MAJOR: Remove this block.
+        if (null === $propertyAccessor) {
+            @trigger_error(sprintf(
+                'Constructing "%s" without passing an instance of "%s" as second argument is deprecated since'
+                .' sonata-project/doctrine-orm-admin-bundle 3.x and will be mandatory in 4.0',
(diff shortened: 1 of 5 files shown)


src/Model/ModelManager.php
```diff
@@ -36,24 +36,32 @@
...
+     */
+    protected $propertyAccessor;
+
     /**
      * @var EntityManager[]
      */
     protected $cache = [];
 
-    public function __construct(ManagerRegistry $registry)
+    public function __construct(ManagerRegistry $registry, ?PropertyAccessorInterface $propertyAccessor = null)
```
jordisala1991, line 61: I think we should change dependency injection to add this new param on constructor

src/Model/ModelManager.php
```diff
@@ -591,48 +599,17 @@ public function modelReverseTransform($class, array $array = [])
...
         foreach ($array as $name => $value) {
-            $reflection_property = false;
             // property or association ?
             if (\array_key_exists($name, $metadata->fieldMappings)) {
                 $property = $metadata->fieldMappings[$name]['fieldName'];
-                $reflection_property = $metadata->reflFields[$name];
             } elseif (\array_key_exists($name, $metadata->associationMappings)) {
                 $property = $metadata->associationMappings[$name]['fieldName'];
             } else {
                 $property = $name;
```
jordisala1991, line 609: initialize property as name to avoid last else, wdyt?

src/Model/ModelManager.php
VincentLanglet: By the way, I added it because it was done like this in this PR https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/pull/368, but we don't always add the dependency in the constructor in SonataAdmin.

I have trouble to see a real reason to do this since there is a 
```
PropertyAccess::createPropertyAccessor()
```
method.

Do you know a good reason for this optional injection ?

jordisala1991: For me there could be some reasons:

1. To have consistency with the rest of injections.
2. To avoid executing code on constructors (ideally)
3. The injected property access is configured by https://github.com/symfony/symfony/blob/0ab5eeddf97d0a277c5ad956a51365aba93b27b9/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L886 But I dont know if that is a good reason or a bad one.

Do you know what configuration we get when creating our own instance?

How do we deal normally with this property accessor?

VincentLanglet: > How do we deal normally with this property accessor?

It depends:
https://github.com/sonata-project/SonataAdminBundle/search?q=createpropertyaccessor&unscoped_q=createpropertyaccessor

Maybe should I add a deprecation if the dependency is not provided and remove the `= null` in the next major ? WDYT ?

jordisala1991: I think that would be great, unless there is a good reason to use the createPropertyAccessor I prefer to just inject the framework bundle one

src/Model/ModelManager.php
VincentLanglet: You mean ?
```
if (\array_key_exists($name, $metadata->fieldMappings)) {
    $name = $metadata->fieldMappings[$name]['fieldName'];
} elseif (\array_key_exists($name, $metadata->associationMappings)) {
    $name = $metadata->associationMappings[$name]['fieldName'];
}
```

jordisala1991: Yes

VincentLanglet: It can be confusing if someone want to use the $name property later in this function.

WDYT of the private method `getFieldName` ?

jordisala1991: Even better!

jordisala1991:
Thank you @VincentLanglet

PR #1084, opened 2020-08-03
Remove sonata_help block

nieuwenhuisen:
## Subject

**Fix for the Exception:**
Block "sonata_help" on template "form_div_layout.html.twig" does not exist.

**Deprecation message:**
The "sonata_help" option is deprecated since sonata-project/admin-bundle 3.60, to be removed in 4.0. Use "help" instead.

I am targeting this branch, because this branch gives an exception.

## Changelog

```markdown
### Changed
- Changed `{{ block('sonata_help') }}` to `{{ form_help(form) }}` in `form_admin_fields.html.twig`.
```

### DIFF
--- a/src/Resources/views/Form/form_admin_fields.html.twig
+++ b/src/Resources/views/Form/form_admin_fields.html.twig
@@ -131,7 +131,7 @@ file that was distributed with this source code.
             {{ form_widget(form, {'required':false}) }}
         </span>
 
-        {{ block('sonata_help') }}
+        {{ form_help(form) }}
 
         <div class="modal fade" id="field_dialog_{{ id }}" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" aria-hidden="true">
             <div class="modal-dialog modal-lg">


jordisala1991:
@franmomu I ping you because you were deprecating things recently about this topic, is this change correct?

VincentLanglet:
I'm surprised about the fact we keep using template from a storage bundle.
Shouldn't we move every template to SonataAdminBundle ?

franmomu:
I guess [some of them](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Resources/views/Form/form_admin_fields.html.twig#L155) are difficult to move.

jordisala1991:
This only gives error on master right? on 3.x works fine?

franmomu:
> This only gives error on master right? on 3.x works fine?

Exactly, in 3.x this block exists: https://github.com/sonata-project/SonataAdminBundle/blob/3.x/src/Resources/views/Form/form_admin_fields.html.twig#L27

jordisala1991:
Thank you @nieuwenhuisen

PR #1086, opened 2020-08-13
Remove references to deprecated help option

franmomu:
Those methods are from `Sonata\AdminBundle\Admin\BaseFieldDescription`, we shouldn't test them here.


VincentLanglet:
Thanks @franmomu

PR #1092, opened 2020-08-24
Add support for Twig 3

willemverspyck:
## Subject

## Changelog

```markdown
### Added
- Add support for Twig 3
```


jordisala1991:
There are some test that need to be fixed. I am not sure if they are broken because of this PR or on a previous one. @wbloszyk can you check if it is because the block bundle 4 support?

jaikdean:
@jordisala1991 The tests were broken when block bundle 4 support was added, see https://travis-ci.org/github/sonata-project/SonataDoctrineORMAdminBundle/jobs/720385821

jordisala1991:
Do you want to help fixing tests?

willemverspyck:
@jordisala1991 I just fixed one of the failed tests, but not sure how to handle the other 2 failed tests, because "DeprecatedAuditBlockServiceTest" extends the class "BlockServiceTestCase" from SonataBlockBundle 4 and the constructor has been changed since SonataBlockBundle 3. Shall I remove that "DeprecatedAuditBlockServiceTest" class? The class "AuditBlockService" is already tested in another file.

wbloszyk:
@willemverspyck 
Agree, `DeprecatedAuditBlockServiceTest` can be deleted.

willemverspyck:
@wbloszyk Thanks! File has been removed.

tests/Block/DeprecatedAuditBlockServiceTest.php
```diff
@@ -1,93 +0,0 @@
...
-    private $blockService;
-
-    protected function setUp(): void
-    {
-        parent::setUp();
-        $this->simpleThingsAuditReader = $this->prophesize(SimpleThingsAuditReader::class);
-
-        $this->blockService = new AuditBlockService(
-            'block.service',
-            $this->templating,
```
jordisala1991, line 43: IMO this test should not be removed, but this $this->templating should be fixed. It can be declared on this same class.

jordisala1991: And probably this will need to only execute this tests for block bundle 3, not for version 4.

wbloszyk: We can remove it becouse this arguments is not use directly in `AuditBlockService`. It is hide behind `renderResponse`. Result for templating and twig should be the same. This should be test by BlockBundle, not each Block which extends it.

wbloszyk (approved):

core23 (approved):

PR #1109, opened 2020-08-31
Add deprecations for ModelManager methods

VincentLanglet:
## Subject

These methods are deprecated in Sonata Admin.

## Changelog

```markdown
### Deprecated
- Deprecate ModelManager collections methods.
- Deprecate ModelManager::getPaginationParameters().
- Deprecate ModelManager::getSortParameters().
```

### DIFF
--- a/tests/Model/ModelManagerTest.php
+++ b/tests/Model/ModelManagerTest.php
@@ -59,10 +59,13 @@ use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\UuidBinaryEntity;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\UuidEntity;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\VersionedEntity;
 use Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Util\NonIntegerIdentifierTestClass;
+use Symfony\Bridge\PhpUnit\ExpectDeprecationTrait;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 
 class ModelManagerTest extends TestCase
 {
+    use ExpectDeprecationTrait;
+
     /**
      * @var ManagerRegistry|MockObject
      */
@@ -143,6 +146,11 @@ class ModelManagerTest extends TestCase
         $this->assertSame($em, $this->modelManager->getEntityManager('x'));
     }
 
+    /**
+     * NEXT_MAJOR: Remove this test.
+     *
+     * @group legacy
+     */
     public function testSortParameters(): void
     {
         $datagrid1 = $this->createMock(Datagrid::class);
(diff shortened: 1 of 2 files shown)


VincentLanglet:
Please review @sonata-project/contributors

tests/Model/ModelManagerTest.php
```diff
@@ -143,6 +143,12 @@ public function testInstantiateWithDeprecatedRegistryInterface(): void
         $this->assertSame($em, $this->modelManager->getEntityManager('x'));
     }
 
+    /**
+     * NEXT_MAJOR: Remove this test.
+     *
+     * @group legacy
+     * @expectedDeprecation Method Sonata\DoctrineORMAdminBundle\Model\ModelManager::getSortParameters() is deprecated since sonata-project/doctrine-orm-admin-bundle 3.x and will be removed in version 4.0.
```
franmomu, line 150: The expectedDeprecation annotation was deprecated in favor of the method inside the test: https://symfony.com/doc/current/components/phpunit_bridge.html#id1

core23 (approved):

tests/Model/ModelManagerTest.php
VincentLanglet: That's what we're using in every project currently.
Shouldn't we keep consistency ?

That won't be a real issue since the deprecation is removed in Sonata 4.0

If we start now to move every `@expectedDeprecation`, we'll end to a lot of conflict in master.

franmomu: > If we start now to move every `@expectedDeprecation`, we'll end to a lot of conflict in master.

I haven't said anything about that. IMHO if we are adding new code, I'd rather add non-deprecated one (if it's possible).

jordisala1991: I think, for the new deprecations we should follow the new format, because we have to get used to it, and we dont know when we will have new release. For older deprecation I would say to not migrate anything

jordisala1991 (dismissed):

franmomu (approved):

tests/Model/ModelManagerTest.php
```diff
@@ -143,8 +146,15 @@ public function testInstantiateWithDeprecatedRegistryInterface(): void
...
     }
 
+    /**
+     * NEXT_MAJOR: Remove this test.
+     *
+     * @group legacy
+     */
     public function testSortParameters(): void
     {
+        $this->expectDeprecation('Method Sonata\DoctrineORMAdminBundle\Model\ModelManager::getSortParameters() is deprecated since sonata-project/doctrine-orm-admin-bundle 3.x and will be removed in version 4.0.');
```
phansys, line 156: Please, place this expectation immediately before where the deprecation is intended to be triggered.

phansys (changes requested):

tests/Model/ModelManagerTest.php
VincentLanglet: done

tests/Model/ModelManagerTest.php
```diff
@@ -742,8 +751,15 @@ public function testModelReverseTransform(): void
...
 
+    /**
+     * NEXT_MAJOR: Remove this test.
+     *
+     * @group legacy
+     */
     public function testCollections(): void
     {
+        $this->expectDeprecation('Method Sonata\DoctrineORMAdminBundle\Model\ModelManager::getModelCollectionInstance() is deprecated since sonata-project/doctrine-orm-admin-bundle 3.x and will be removed in version 4.0.');
+
```
phansys, line 762: In the other tests, there is no extra LF between the expectation and the method call.

phansys (approved):

tests/Model/ModelManagerTest.php
VincentLanglet: Done

jordisala1991 (approved):

PR #1120, opened 2020-09-22
Allow empty string for filtering

peter-gribanov:
## Subject

If the filter value is empty, but the filtering type has been changed from the default value, then this is an indication that the user wants to search by empty value.
We filter by empty value if the filtering type is explicitly specified.

Part of fix for https://github.com/sonata-project/SonataAdminBundle/issues/3569.

## Changelog

```markdown
### Added
- Allow empty string for filtering
```

### DIFF
--- a/docs/reference/filter_field_definition.rst
+++ b/docs/reference/filter_field_definition.rst
@@ -67,6 +67,15 @@ Example
         }
     }
 
+StringFilter
+------------
+
+The string filter has additional options:
+
+* ``case_sensitive`` - set to ``false`` to make the search case insensitive. By default ``true`` is used;
+* ``trim`` - use one of ``Sonata\DoctrineORMAdminBundle\Filter\TRIM_*`` constants to control the clearing of blank spaces around in the value. By default ``Sonata\DoctrineORMAdminBundle\Filter\TRIM_BOTH`` is used;
+* ``allow_empty`` - set to ``true`` to enable search by empty value. By default ``false`` is used.
+
 StringListFilter
 ----------------
 
--- a/src/Filter/StringFilter.php
+++ b/src/Filter/StringFilter.php
@@ -22,6 +22,11 @@ use Sonata\AdminBundle\Form\Type\Operator\StringOperatorType;
  */
 class StringFilter extends Filter
 {
+    public const TRIM_NONE = 0;
+    public const TRIM_LEFT = 1;
+    public const TRIM_RIGHT = 2;
+    public const TRIM_BOTH = self::TRIM_LEFT | self::TRIM_RIGHT;
+
     public const CHOICES = [
         StringOperatorType::TYPE_CONTAINS => 'LIKE',
         StringOperatorType::TYPE_STARTS_WITH => 'LIKE',
@@ -31,19 +36,32 @@ class StringFilter extends Filter
         StringOperatorType::TYPE_NOT_EQUAL => '<>',
     ];
 
+    /**
+     * Filtering types do not make sense for searching by empty value.
+     */
+    private const MEANINGLESS_TYPES = [
+        StringOperatorType::TYPE_CONTAINS,
+        StringOperatorType::TYPE_STARTS_WITH,
+        StringOperatorType::TYPE_ENDS_WITH,
+        StringOperatorType::TYPE_NOT_CONTAINS,
+    ];
+
     public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $value)
     {
-        if (!$value || !\is_array($value) || !\array_key_exists('value', $value) || null === $value['value']) {
+        if (!\is_array($value) || !\array_key_exists('value', $value)) {
             return;
         }
 
-        $value['value'] = trim($value['value']);
-
-        if (0 === \strlen($value['value'])) {
-            return;
-        }
-
+        $value['value'] = $this->trim((string) ($value['value'] ?? ''));
         $type = $value['type'] ?? StringOperatorType::TYPE_CONTAINS;
+
+        // ignore empty value if it doesn't make sense
+        if ('' === $value['value'] &&
+            (!$this->getOption('allow_empty') || \in_array($type, self::MEANINGLESS_TYPES, true))
+        ) {
+            return;
+        }
+
         $operator = $this->getOperator((int) $type);
 
         // c.name > '1' => c.name OPERATOR :FIELDNAME
@@ -102,6 +120,8 @@ class StringFilter extends Filter
             // NEXT_MAJOR: Remove the format option.
             'format' => '%%%s%%',
             'case_sensitive' => true,
+            'trim' => self::TRIM_BOTH,
+            'allow_empty' => false,
         ];
     }
 
@@ -117,5 +137,20 @@ class StringFilter extends Filter
     private function getOperator(int $type): string
     {
         return self::CHOICES[$type] ?? self::CHOICES[StringOperatorType::TYPE_CONTAINS];
+    }
+
+    private function trim(string $string): string
+    {
+        $trimMode = $this->getOption('trim');
+
+        if ($trimMode & self::TRIM_LEFT) {
+            $string = ltrim($string);
+        }
+
+        if ($trimMode & self::TRIM_RIGHT) {
+            $string = rtrim($string);
+        }
+
+        return $string;
     }
 }
--- a/tests/Filter/StringFilterTest.php
+++ b/tests/Filter/StringFilterTest.php
@@ -29,110 +29,186 @@ class StringFilterTest extends TestCase
 
         $filter->filter($builder, 'alias', 'field', null);
         $filter->filter($builder, 'alias', 'field', '');
+        $filter->filter($builder, 'alias', 'field', []);
 
         $this->assertSame([], $builder->query);
         $this->assertFalse($filter->isActive());
     }
 
-    public function testNullValue(): void
+    public function getValues(): iterable
     {
-        $filter = new StringFilter();
-        $filter->initialize('field_name');
-
-        $builder = new ProxyQuery(new QueryBuilder());
-        $this->assertSame([], $builder->query);
-
-        $filter->filter($builder, 'alias', 'field', ['value' => null, 'type' => StringOperatorType::TYPE_EQUAL]);
-        $this->assertFalse($filter->isActive());
+        return [
+            'filter by normal value' => ['asd', false],
+            'not filter by empty string' => ['', false],
+            'filter by empty string' => ['', true],
+            'not filter by null' => [null, false],
+            'filter by null' => [null, true],
+            'not filter by 0' => [0, false],
+            'filter by 0' => [0, true],
+            'not filter by \'0\'' => ['0', false],
+            'filter by \'0\'' => ['0', true],
+        ];
     }
 
-    public function testContains(): void
+    /**
+     * @dataProvider getValues
+     */
+    public function testDefaultType($value, bool $allowEmpty): void
     {
         $filter = new StringFilter();
-        $filter->initialize('field_name');
+        $filter->initialize('field_name', ['allow_empty' => $allowEmpty]);
 
         $builder = new ProxyQuery(new QueryBuilder());
         $this->assertSame([], $builder->query);
 
-        $filter->filter($builder, 'alias', 'field', ['value' => 'asd', 'type' => StringOperatorType::TYPE_CONTAINS]);
-        $this->assertSame(['alias.field LIKE :field_name_0'], $builder->query);
-        $this->assertSame(['field_name_0' => '%asd%'], $builder->parameters);
+        $filter->filter($builder, 'alias', 'field', ['value' => $value, 'type' => null]);
 
-        $builder = new ProxyQuery(new QueryBuilder());
-        $this->assertSame([], $builder->query);
-
-        $filter->filter($builder, 'alias', 'field', ['value' => 'asd', 'type' => null]);
-        $this->assertSame(['alias.field LIKE :field_name_0'], $builder->query);
-        $this->assertSame(['field_name_0' => '%asd%'], $builder->parameters);
-        $this->assertTrue($filter->isActive());
+        if ('' !== (string) $value) {
+            $this->assertSame(['alias.field LIKE :field_name_0'], $builder->query);
+            $this->assertSame(['field_name_0' => sprintf('%%%s%%', $value)], $builder->parameters);
+            $this->assertTrue($filter->isActive());
+        } else {
+            $this->assertSame([], $builder->query);
+            $this->assertFalse($filter->isActive());
+        }
     }
 
-    public function testStartsWith(): void
+    /**
+     * @dataProvider getValues
+     */
+    public function testContains($value, bool $allowEmpty): void
     {
         $filter = new StringFilter();
-        $filter->initialize('field_name');
+        $filter->initialize('field_name', ['allow_empty' => $allowEmpty]);
 
         $builder = new ProxyQuery(new QueryBuilder());
         $this->assertSame([], $builder->query);
 
-        $filter->filter($builder, 'alias', 'field', ['value' => 'asd', 'type' => StringOperatorType::TYPE_STARTS_WITH]);
-        $this->assertSame(['alias.field LIKE :field_name_0'], $builder->query);
-        $this->assertSame(['field_name_0' => 'asd%'], $builder->parameters);
+        $filter->filter($builder, 'alias', 'field', ['value' => $value, 'type' => StringOperatorType::TYPE_CONTAINS]);
+
+        if ('' !== (string) $value) {
+            $this->assertSame(['alias.field LIKE :field_name_0'], $builder->query);
+            $this->assertSame(['field_name_0' => sprintf('%%%s%%', $value)], $builder->parameters);
+            $this->assertTrue($filter->isActive());
+        } else {
+            $this->assertSame([], $builder->query);
+            $this->assertFalse($filter->isActive());
+        }
     }
 
-    public function testEndsWith(): void
+    /**
+     * @dataProvider getValues
+     */
+    public function testStartsWith($value, bool $allowEmpty): void
     {
         $filter = new StringFilter();
-        $filter->initialize('field_name');
+        $filter->initialize('field_name', ['allow_empty' => $allowEmpty]);
 
         $builder = new ProxyQuery(new QueryBuilder());
         $this->assertSame([], $builder->query);
 
-        $filter->filter($builder, 'alias', 'field', ['value' => 'asd', 'type' => StringOperatorType::TYPE_ENDS_WITH]);
-        $this->assertSame(['alias.field LIKE :field_name_0'], $builder->query);
-        $this->assertSame(['field_name_0' => '%asd'], $builder->parameters);
+        $filter->filter($builder, 'alias', 'field', ['value' => $value, 'type' => StringOperatorType::TYPE_STARTS_WITH]);
+
+        if ('' !== (string) $value) {
+            $this->assertSame(['alias.field LIKE :field_name_0'], $builder->query);
+            $this->assertSame(['field_name_0' => sprintf('%s%%', $value)], $builder->parameters);
+            $this->assertTrue($filter->isActive());
+        } else {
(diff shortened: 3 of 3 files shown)


src/Filter/StringFilter.php
```diff
@@ -29,13 +29,14 @@ class StringFilter extends Filter
...
+        if (!$data || !\is_array($data) || !\array_key_exists('value', $data)) {
             return;
         }
 
-        $data['value'] = trim($data['value']);
+        $data['value'] = trim($data['value'] ?? '');
 
-        if (0 === \strlen($data['value'])) {
+        // default filter value
+        if (0 === \strlen($data['value']) && empty($data['type'])) {
```
VincentLanglet, line 39: I feel like this check shouldn't even exists.
- If the SonataAdmin decided to apply the filter, I shouldn't check again.

Moreover, if you want to allow looking for empty value, you should also also looking for value with only spaces, like `       ` ; isn't it ?

Maybe a `trim` (true, false ? Or left, right, both, none ?) option should be added.

peter-gribanov: I add ability to use spaces in filtering.

src/Filter/StringFilter.php
```diff
@@ -29,13 +29,14 @@ class StringFilter extends Filter
 
     public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
     {
-        if (!$data || !\is_array($data) || !\array_key_exists('value', $data) || null === $data['value']) {
+        if (!$data || !\is_array($data) || !\array_key_exists('value', $data)) {
             return;
         }
 
-        $data['value'] = trim($data['value']);
```
VincentLanglet, line 36: I'm not sure this can be done like this and is not considered as a BC break.

What I meant was to create a boolean option `trim`, default to true.
And then we could do
```
$value = $data['value'] ?? '';
if ($this->getOption('trim')) {
    $value = trim($value);
}
```

But we definitely should do something about value with only spaces.

Before, if you were looking for `contains`: `space-space-space Vincent`, the extra-space were trimmed. I think it was useful, because we generally looking for a string, and extra space were not what we were searching. Plus it was an easy typing mistake.

Now, if you look for `contains`: ``, you looking for empty string.
But if you're looking for `contains`: `space-space-space`, I find weird that we're looking for empty string, and not especially three consecutive spaces.

Then, to me, the check
```
if ('' === $data['value'] && empty($data['type'])) {
```
is useless and can be removed, because it's already done by SonataAdmin.

VincentLanglet (changes requested):
I would like more opinion about the trim.

src/Filter/StringFilter.php
peter-gribanov: I added modes for trim.

peter-gribanov: >  Then, to me, the check
> 
> ```
> if ('' === $data['value'] && empty($data['type'])) {
> ```
> is useless and can be removed, because it's already done by SonataAdmin.

Not certainly in that way. We only ignore cases when both the type and value of the filter are not specified. Now the SonataAdmin is really ignoring these cases, but PR https://github.com/sonata-project/SonataAdminBundle/pull/6402 will be change this behavior.

src/Filter/StringFilter.php
```diff
@@ -115,4 +121,19 @@ private function getOperator(int $type): string
     {
         return self::CHOICES[$type] ?? self::CHOICES[StringOperatorType::TYPE_CONTAINS];
     }
+
+    private function trim(string $string): string
+    {
+        $trim_mode = $this->getOption('trim', self::TRIM_BOTH);
```
VincentLanglet, line 127: Shouldn't trim be declared in `getDefaultOptions` ?

src/Filter/StringFilter.php
VincentLanglet: The SonataAdmin Pr: https://github.com/sonata-project/SonataAdminBundle/pull/6402/files
is saying "If there is a value OR a type, use the filter".
So it means that if there is no value and no type, the filter is ignored.

The check 
```
if ('' === $data['value'] && empty($data['type'])) {
```
is saying "If there is no value and no type, ignore this".

To me, this is already excluded.

src/Filter/StringFilter.php
peter-gribanov: fixed

VincentLanglet:
@peter-gribanov Since you added a trim option, most of my https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1120#discussion_r492841226 concern are solved.

There is just the following check
```
if ('' === $data['value'] && empty($data['type'])) {
```
which should be removed IMHO. But that's debatable. 

I would find weird the fact that
```
$data = [
    'value' => '    '
]
```
would lead to no result, but
```
$data = [
    'value' => '    ',
    'type' => 1
]
```
does.

Anyway, I don't recommend you to work again on this PR before having other @sonata-project/contributors opinion about the whole empty-value-filters strategy.

src/Filter/StringFilter.php
wbloszyk: IMO we should not trim string filter. What when someone want find double spaces in posts or something simmilar?

VincentLanglet: I agree that the trim is questionable but it's the current behavior.

The only BC path I see would be to add the trim option in 3.x, and change the default value in 4.0 to TRIM_NONE, if we decide that's better to not trim.

src/Filter/StringFilter.php
```diff
@@ -29,13 +34,14 @@ class StringFilter extends Filter
 
     public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data)
     {
-        if (!$data || !\is_array($data) || !\array_key_exists('value', $data) || null === $data['value']) {
+        if (!$data || !\is_array($data) || !\array_key_exists('value', $data)) {
```
wbloszyk, line 37:
```suggestion
        if (!\isset($data['value']) && !\in_array($data['value'], [null, ''])) {
```

VincentLanglet: @peter-gribanov want to allow the value `''`/`null`.

peter-gribanov: `null` and empty string is available values.

VincentLanglet:
Any opinion on this PR and the whole empty-value-filter strategy ? @sonata-project/contributors

VincentLanglet:
@peter-gribanov I start to think this behavior change could be better with an option.
```
allow_empty => false/true
```
with false as default value to not change the actual behavior.

```
// Do the trim 

if ('' === $value['value'] && !$allow_empty)) {
     return;
}

// Apply the filter
```

If you want to filter by empty string you just have to set it `true`. WDYT ?

tests/Filter/StringFilterTest.php
```diff
@@ -30,96 +30,162 @@ public function testEmpty(): void
...
-        $filter->filter($builder, 'alias', 'field', ['value' => 'asd', 'type' => null]);
-        $this->assertSame(['alias.field LIKE :field_name_0'], $builder->query);
-        $this->assertSame(['field_name_0' => '%asd%'], $builder->parameters);
-        $this->assertTrue($filter->isActive());
+        $filter->filter($builder, 'alias', 'field', ['value' => $value, 'type' => ContainsOperatorType::TYPE_CONTAINS]);
+
+        if ($value || $allowEmpty) {
+            $this->assertSame(['alias.field LIKE :field_name_0'], $builder->query);
+            $this->assertSame(['field_name_0' => sprintf('%%%s%%', $value)], $builder->parameters);
+        }
```
VincentLanglet, line 105: else ? If I understand correctly we can add 

```
else {
    $this->assertSame([], $builder->query);
}
```

tests/Filter/StringFilterTest.php
```diff
@@ -30,96 +30,162 @@ public function testEmpty(): void
...
 
-        $filter->filter($builder, 'alias', 'field', ['value' => 'asd', 'type' => StringOperatorType::TYPE_STARTS_WITH]);
-        $this->assertSame(['alias.field LIKE :field_name_0'], $builder->query);
-        $this->assertSame(['field_name_0' => 'asd%'], $builder->parameters);
+        $filter->filter($builder, 'alias', 'field', ['value' => $value, 'type' => StringOperatorType::TYPE_STARTS_WITH]);
+
+        if ($value || $allowEmpty) {
+            $this->assertSame(['alias.field LIKE :field_name_0'], $builder->query);
+            $this->assertSame(['field_name_0' => sprintf('%s%%', $value)], $builder->parameters);
+        }
```
VincentLanglet, line 124: Same

tests/Filter/StringFilterTest.php
```diff
@@ -30,96 +30,162 @@ public function testEmpty(): void
...
 
-        $filter->filter($builder, 'alias', 'field', ['value' => 'asd', 'type' => StringOperatorType::TYPE_ENDS_WITH]);
-        $this->assertSame(['alias.field LIKE :field_name_0'], $builder->query);
-        $this->assertSame(['field_name_0' => '%asd'], $builder->parameters);
+        $filter->filter($builder, 'alias', 'field', ['value' => $value, 'type' => StringOperatorType::TYPE_ENDS_WITH]);
+
+        if ($value || $allowEmpty) {
+            $this->assertSame(['alias.field LIKE :field_name_0'], $builder->query);
+            $this->assertSame(['field_name_0' => sprintf('%%%s', $value)], $builder->parameters);
+        }
```
VincentLanglet, line 143: same

src/Filter/StringFilter.php
```diff
@@ -58,6 +64,10 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $value
 
         if (StringOperatorType::TYPE_NOT_CONTAINS === $type) {
             $or->add($queryBuilder->expr()->isNull(sprintf('%s.%s', $alias, $field)));
+
+            if ('' === $value['value']) {
+                $or->add(sprintf('%s.%s <> :%s', $alias, $field, $parameterName));
+            }
```
VincentLanglet, line 70: Why did you add this ? This doesn't seem necessary to me.

`NOT LIKE ''` is working

peter-gribanov: `NOT LIKE ''` is not equal `<> ''`. The DB Query Optimizer should convert the first expression to the second, but that doesn't mean you should rely on it. It is better not to use `LIKE` where it is not needed.

src/Filter/StringFilter.php
```diff
@@ -114,8 +122,28 @@ public function getRenderSettings()
         ]];
     }
 
-    private function getOperator(int $type): string
+    private function getOperator(int $type, string $value): string
     {
+        // using an explicit comparison is preferred over LIKE
+        if ('' === $value) {
+            return StringOperatorType::TYPE_NOT_CONTAINS === $type ? '<>' : '=';
```
VincentLanglet, line 129: I'm not ok with this.

CONTAINS '' should return everything.
EQUAL '' should return = ''.
NOT CONTAINS '' should return nothing.

VincentLanglet (changes requested):
Currently there is a format used `%%%s%%`.

When there is a value 'foo' is provided we're looking for
- CONTAINS => LIKE %foo%
- START => LIKE %foo
- END => LIKE foo%
- NOT_CONTAINS => NOT LIKE %foo%
- EQUAL => = foo

With the value '', you're looking for 
- CONTAINS => LIKE %%
- START => LIKE %
- END => LIKE %
- NOT_CONTAINS => NOT LIKE %% (then you changed this to <> %% which doesn't sound good)
- EQUAL => ''

Before modifying the code and the tests, can you explain the query you expect to do ?

IMHO
- Contains '' should return everything, because every string contains empty string.
- Same for start ''.
- Same for end ''.
- Not contains should return nothing. But I think I understood why you changed to <> ; IMHO we should add `StringOperatorType::NOT_EQUAL` for this. It requires first a PR to add this: see https://github.com/sonata-project/SonataAdminBundle/pull/6134
- Equal should do `= ''`.

Can you confirm ?

peter-gribanov:
IMHO `CONTAINS => LIKE %%` will be used by default and doesn't make sense.
Same for `START => LIKE %`, `END => LIKE %` and `NOT_CONTAINS => NOT LIKE %%`.
Only `EQUAL => ''` and `NO_EQUAL => <> ''` *(not supported now)* will actually work with empty values.

Therefore i suggest `CONTAINS`, `STARTS_WITH`, `ENDS_WITH` to be considered as `EQUAL`  and `NOT_CONTAINS` as `NOT_EQUAL` for empty values.

VincentLanglet:
> IMHO `CONTAINS => LIKE %%` will be used by default and doesn't make sense.

We could change
```
$type = $value['type'] ?? StringOperatorType::TYPE_CONTAINS;
```
to default to another type in case the value is empty.

> Same for `START => LIKE %`, `END => LIKE %` and `NOT_CONTAINS => NOT LIKE %%`.

They won't be used indeed. 
But semantically, I find them more natural, so I would keep the consistency. See, in php 8
```
str_starts_with('Foo', ''); // true
str_starts_with('', ''); // true

str_ends_with('Foo', ''); // true
str_ends_with('', ''); // true
```

> Only `EQUAL => ''` and `NO_EQUAL => <> ''` _(not supported now)_ will actually work with empty values.

The initial issue was to have a way to filter by empty value, this will be one :)

src/Filter/StringFilter.php
```diff
@@ -30,29 +35,42 @@ class StringFilter extends Filter
...
-        $type = $value['type'] ?? StringOperatorType::TYPE_CONTAINS;
         $operator = $this->getOperator((int) $type);
 
         // c.name > '1' => c.name OPERATOR :FIELDNAME
         $parameterName = $this->getNewParameterName($queryBuilder);
 
         $or = $queryBuilder->expr()->orX();
 
-        if ($this->getOption('case_sensitive')) {
+        if ($this->getOption('case_sensitive') || '' === $value['value']) {
```
VincentLanglet, line 70: I understand why you add the check `|| '' === $value['value']`, but it could be the same for `' '`, `'     '`, `'0'`, '`42'`, `'$^&@'`, and lot of values so I find weird to have a special treatment here.

peter-gribanov:
> > IMHO `CONTAINS => LIKE %%` will be used by default and doesn't make sense.
>
> We could change
> ```php
> $type = $value['type'] ?? StringOperatorType::TYPE_CONTAINS;
> ```

This is not good, because the default behavior is changed from the previous version. In the current version, the empty value is ignored and the filter is not applied. The new version (with this PR) will apply a filter for empty values. This can cause problems with BC.

VincentLanglet:
@peter-gribanov I opened the PR to add the NOT_EQUAL operator
https://github.com/sonata-project/SonataAdminBundle/pull/6585

VincentLanglet:
The support for the NOT_EQUAL operator is now merged in this bundle.
You just have to rebase, and this PR should be ready to merge.

VincentLanglet (changes requested):
Can you use `@dataProvider getValues` for the `testNotEquals` too ?

And the PR will be perfect IMHO.

VincentLanglet (dismissed):

src/Filter/StringFilter.php
```diff
@@ -118,4 +138,19 @@ private function getOperator(int $type): string
     {
         return self::CHOICES[$type] ?? self::CHOICES[StringOperatorType::TYPE_CONTAINS];
     }
+
+    private function trim(string $string): string
+    {
+        $trim_mode = $this->getOption('trim');
```
phansys, line 144:
```suggestion
        $trimMode = $this->getOption('trim');
```

src/Filter/StringFilter.php
```diff
@@ -118,4 +138,19 @@ private function getOperator(int $type): string
...
+
+        if ($trim_mode & self::TRIM_LEFT) {
```
phansys, line 146:
```suggestion
        if ($trimMode & self::TRIM_LEFT) {
```

src/Filter/StringFilter.php
```diff
@@ -118,4 +138,19 @@ private function getOperator(int $type): string
...
+            $string = ltrim($string);
+        }
+
+        if ($trim_mode & self::TRIM_RIGHT) {
```
phansys, line 150:
```suggestion
        if ($trimMode & self::TRIM_RIGHT) {
```

tests/Filter/StringFilterTest.php
```diff
@@ -29,110 +29,182 @@ public function testEmpty(): void
...
-
-        $builder = new ProxyQuery(new QueryBuilder());
-        $this->assertSame([], $builder->query);
+        $filter->filter($builder, 'alias', 'field', ['value' => $value, 'type' => StringOperatorType::TYPE_CONTAINS]);
 
-        $filter->filter($builder, 'alias', 'field', ['value' => 'asd', 'type' => null]);
-        $this->assertSame(['alias.field LIKE :field_name_0'], $builder->query);
-        $this->assertSame(['field_name_0' => '%asd%'], $builder->parameters);
-        $this->assertTrue($filter->isActive());
+        if ($value) {
```
phansys, line 85: IMO, we should make this condition explicit, otherwise values like `0` or `'0'` will be interpreted as false. Indeed, these values should be added to the list of the provided values.

docs/reference/filter_field_definition.rst
```diff
@@ -67,6 +67,15 @@ Example
         }
     }
 
+StringFilter
+------------
+
+The string filter has additional options:
+
+* ``case_sensitive`` - set to ``false`` to make the search case insensitive;
```
phansys, line 75: All of these options should indicate its default value or its default behavior.

docs/reference/filter_field_definition.rst
```diff
@@ -67,6 +67,15 @@ Example
...
+* ``trim`` - use one of ``Sonata\DoctrineORMAdminBundle\Filter\TRIM_*`` to control the clearing of spaces in the value. As default used ``Sonata\DoctrineORMAdminBundle\Filter\TRIM_BOTH``;
```
phansys, line 76:
```suggestion
* ``trim`` - use one of ``Sonata\DoctrineORMAdminBundle\Filter\TRIM_*`` constants to control the clearing of blank spaces around the value. By default ``Sonata\DoctrineORMAdminBundle\Filter\TRIM_BOTH`` is used;
```

phansys (changes requested):

tests/Filter/StringFilterTest.php
peter-gribanov: If we try to use `0`, we get error:

> Argument 1 passed to Sonata\DoctrineORMAdminBundle\Tests\Filter\StringFilterTest::testNotEquals() must be of the type string or null, int given

phansys: I missed that you were declaring `?string` for the value in the test method.
I think this is ok only if the filter really can't handle other types (like the integer in my comment).

VincentLanglet: I would say that the string filter is made to deal with string only.

VincentLanglet (dismissed):

tests/Filter/StringFilterTest.php
phansys: >I would say that the string filter is made to deal with string only.

I agree, that's the intention. But since the method is public, there is no check preventing other types and the provided value is being casted as string, I think a test covering these cases could help to avoid unintended behaviors.
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/80b37e6336b3fb9293da35fde06a4b1eef8e1835/src/Filter/StringFilter.php#L55

peter-gribanov: I add data type casting

VincentLanglet (approved):

tests/Filter/StringFilterTest.php
phansys: Calls to `ltrim()` and `rtrim()` were already performing the casting.

phansys (approved):

tests/Filter/StringFilterTest.php
peter-gribanov: No, because a string from the `StringFilter::trim()` method comes to them. The typecast had to be done before calling `StringFilter::trim()`.

VincentLanglet:
Thanks @peter-gribanov

PR #1127, opened 2020-09-29
Add more phpdoc

core23:
## Subject

## Changelog

```markdown
### Added
- Add more phpdoc
```

## To do

- [x] Add phpstan to CI

### DIFF
--- a/src/Admin/FieldDescription.php
+++ b/src/Admin/FieldDescription.php
@@ -15,6 +15,9 @@ namespace Sonata\DoctrineORMAdminBundle\Admin;
 
 use Sonata\AdminBundle\Admin\BaseFieldDescription;
 
+/**
+ * @final since sonata-project/doctrine-orm-admin-bundle 3.x
+ */
 class FieldDescription extends BaseFieldDescription
 {
     public function __construct()
@@ -52,9 +55,6 @@ class FieldDescription extends BaseFieldDescription
         return $this->getTargetModel();
     }
 
-    /**
-     * @final since sonata-project/doctrine-orm-admin-bundle 3.20.
-     */
     public function getTargetModel(): ?string
     {
         if ($this->associationMapping) {
--- a/src/Filter/ModelAutocompleteFilter.php
+++ b/src/Filter/ModelAutocompleteFilter.php
@@ -20,22 +20,25 @@ use Sonata\AdminBundle\Form\Type\Filter\DefaultType;
 use Sonata\AdminBundle\Form\Type\ModelAutocompleteType;
 use Sonata\AdminBundle\Form\Type\Operator\EqualOperatorType;
 
+/**
+ * @final since sonata-project/doctrine-orm-admin-bundle 3.x
+ */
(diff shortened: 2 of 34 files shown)


src/Filter/ModelAutocompleteFilter.php
```diff
@@ -67,52 +70,59 @@ public function getRenderSettings()
...
+        $queryBuilder->setParameter($parameterName, $value['value']);
     }
 
-    protected function association(ProxyQueryInterface $queryBuilder, $data)
+    /**
+     * @param mixed[] $value
+     *
+     * @return array
+     *
+     * @phpstan-return array(string, bool)
```
VincentLanglet, line 123: I don't know this syntax, can't say if it's correct.

VincentLanglet (commented):
IMHO we should add phpstan before adding some `@phpstan-` phpdoc.

src/Admin/FieldDescription.php
```diff
@@ -15,6 +15,9 @@
 
 use Sonata\AdminBundle\Admin\BaseFieldDescription;
 
+/**
+ * @final since sonata-project/doctrine-orm-admin-bundle 3.x
```
jordisala1991, line 19: IMO we should ensure that all final we want to add is not extended anywhere in any of the sonata bundles. We have some cases where we marked something final for Next major and it cant happen because is extended somewhere else. Have you check all of them?

src/Filter/ModelAutocompleteFilter.php
greg0ire: Did you mean to write this:
```suggestion
     * @phpstan-return array{string, bool}
```
?

src/Admin/FieldDescription.php
core23: AFAIK we do not extend any class from the doctrine specific admin bundles.

src/Model/ModelManager.php
```diff
@@ -98,11 +103,10 @@ public function getMetadata($class)
...
      *
-     * @return array(
-     *                \Doctrine\ORM\Mapping\ClassMetadata $parentMetadata,
-     *                string $lastPropertyName,
-     *                array $parentAssociationMappings
-     *                )
+     * @return array
+     *
+     * @phpstan-param class-string $baseClass
+     * @phpstan-return array{\Doctrine\ORM\Mapping\ClassMetadata, string, array}|null
```
VincentLanglet, line 109:
```suggestion
     * @phpstan-return array{\Doctrine\ORM\Mapping\ClassMetadata, string, array}
```

VincentLanglet (changes requested):

VincentLanglet (approved):

franmomu (approved):

OskarStark:
Thank you @core23

OskarStark:
Are you fine with `patch` label?

core23:
Shouldn't we use `minor`, because we finalize some components?

PR #1128, opened 2020-09-30
Add phpstan to CI

core23:
## Subject

When adding phpstan, I also discovered a `Variable $objectIdIterator might not be defined.` error.

## Changelog

```markdown
### Fixed
- Fix null error in `ObjectAclManipulator`
```

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -1,0 +1,14 @@
+on: [push, pull_request]
+name: Quality assurance
+jobs:
+  phpstan:
+    name: PHPStan
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: PHPStan
+        uses: "docker://oskarstark/phpstan-ga"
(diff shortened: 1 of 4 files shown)


.github/workflows/qa.yaml
```diff
@@ -0,0 +1,14 @@
+on: [push, pull_request]
+name: Quality assurance
+jobs:
+  phpstan:
+    name: PHPStan
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
```
OskarStark, line 8: v2 instead master

OskarStark (dismissed):

.github/workflows/qa.yaml
core23: Fixed

OskarStark (approved):

OskarStark:
Thanks 🙏

VincentLanglet:
We should allow to run phpstan locally too.

PR #1136, opened 2020-10-03
Cast the value as string before calling `explode()` at `ModelManager::addIdentifiersToQuery()`

phansys:
## Subject
Cast the value as string before calling `explode()` at `ModelManager::addIdentifiersToQuery()`.

## Changelog

```markdown
### Fixed
- `TypeError` thrown by `explode()` when receiving non string values as argument 2 from argument 3 at `ModelManager::addIdentifiersToQuery()`.
```

### DIFF
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -502,7 +502,7 @@ class ModelManager implements ModelManagerInterface, LockInterface
     }
 
     /**
-     * @phpstan-param non-empty-array<string> $idx
+     * @phpstan-param non-empty-array<string|int> $idx
      *
      * @throws \InvalidArgumentException if value passed as argument 3 is an empty array
      */
@@ -521,7 +521,7 @@ class ModelManager implements ModelManagerInterface, LockInterface
         $prefix = uniqid();
         $sqls = [];
         foreach ($idx as $pos => $id) {
-            $ids = explode(self::ID_SEPARATOR, $id);
+            $ids = explode(self::ID_SEPARATOR, (string) $id);
 
             $ands = [];
             foreach ($fieldNames as $posName => $name) {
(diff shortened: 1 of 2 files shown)


VincentLanglet:
Ids are an array, should we improve phpdoc about acceptable values ?

Is it always int|string or can it be an object ?

If so, what if there is no toString method ?

phansys:
> Ids are an array, should we improve phpdoc about acceptable values ?

I've added some annotations, but maybe they should be moved to `ModelManagerInterface`. 
> Is it always int|string or can it be an object ?

I don't know if we are allowing non scalar ids, but in that case we could check if the `__toString()` method exists in an object.

VincentLanglet:
> > Ids are an array, should we improve phpdoc about acceptable values ?
> 
> I've added some annotations, but maybe they should be moved to `ModelManagerInterface`.

I agree it's better on the interface. For instance, on master we have
```
/**
 * @param string[] $idx
 *
 * @phpstan-param class-string $class
 */
public function addIdentifiersToQuery(string $class, ProxyQueryInterface $query, array $idx): void;
```
So, we should decide on the interface if we're allowing or not `int`.
Allowing `int` doesn't seem to be a bad idea.

> 
> > Is it always int|string or can it be an object ?
> 
> I don't know if we are allowing non scalar ids, but in that case we could check if the `__toString()` method exists in an object.

We currently have the following signature on master
```
/**
 * @param mixed $id
 *
 * @return object|null the object with id or null if not found
 *
 * @phpstan-template T of object
 * @phpstan-param class-string<T> $class
 * @phpstan-return T|null
 */
public function find(string $class, $id): ?object;
```

And we had the following discussion with @jordisala1991 https://github.com/sonata-project/SonataAdminBundle/pull/6261/files#r467393918

Restricting to `int|string` for both `find` and `addIdentifierToQuery` does not seems to be a bad idea to me. 
WDYT ?

VincentLanglet (approved):

VincentLanglet (approved):

wbloszyk (approved):

VincentLanglet:
Thanks @phansys

PR #1159, opened 2020-10-12
Add deprecation if callback return type is not a boolean

VincentLanglet:
## Subject

## Changelog

```markdown
### Deprecated
- Using a callback filter with a callback option which does not return a boolean
```

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -1,6 +1,40 @@
 UPGRADE 3.x
 ===========
 
+UPGRADE FROM 3.x to 3.x
+=======================
+
+### Sonata\DoctrineORMAdminBundle\Filter\CallbackFilter
+
+Deprecate the usage of a callback which does not return a boolean value.
+
+Previously, this was valid:
+```php
+function callback(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $value)
+{
+    if (!$value['value']) {
+        return;
+    }
+
+    // ...
+
+    return true;
+}
+```
+To remove the deprecation, please update the code this way:
+```php
+function callback(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $value): bool
+{
+    if (!$value['value']) {
+        return false;
+    }
+
+    // ...
+
+    return true;
+}
+```
+
 UPGRADE FROM 3.22 to 3.23
 =========================
 
--- a/docs/reference/filter_field_definition.rst
+++ b/docs/reference/filter_field_definition.rst
@@ -255,9 +255,9 @@ In this example, ``getWithOpenCommentField`` and ``getWithOpenCommentFilter`` im
                 ->add('author')
                 ->add('with_open_comments', CallbackFilter::class, [
     //                'callback'   => [$this, 'getWithOpenCommentFilter'],
-                    'callback' => function($queryBuilder, $alias, $field, $value) {
+                    'callback' => static function(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $value): bool {
                         if (!$value['value']) {
-                            return;
+                            return false;
                         }
 
                         $queryBuilder
@@ -274,7 +274,7 @@ In this example, ``getWithOpenCommentField`` and ``getWithOpenCommentFilter`` im
         public function getWithOpenCommentFilter($queryBuilder, $alias, $field, $value)
         {
             if (!$value['value']) {
-                return;
+                return false;
             }
 
             $queryBuilder
--- a/src/Filter/CallbackFilter.php
+++ b/src/Filter/CallbackFilter.php
@@ -29,7 +29,22 @@ class CallbackFilter extends Filter
             throw new \RuntimeException(sprintf('Please provide a valid callback option "filter" for field "%s"', $this->getName()));
         }
 
-        $this->active = \call_user_func($this->getOption('callback'), $queryBuilder, $alias, $field, $value);
+        $isActive = \call_user_func($this->getOption('callback'), $queryBuilder, $alias, $field, $value);
+        if (!\is_bool($isActive)) {
+            @trigger_error(
+                'Using another return type than boolean for the callback option is deprecated'
+                .' since sonata-project/doctrine-orm-admin-bundle 3.x and will throw an exception in version 4.0.',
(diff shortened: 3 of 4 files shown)


docs/reference/filter_field_definition.rst
```diff
@@ -257,7 +257,7 @@ In this example, ``getWithOpenCommentField`` and ``getWithOpenCommentFilter`` im
     //                'callback'   => [$this, 'getWithOpenCommentFilter'],
                     'callback' => function($queryBuilder, $alias, $field, $value) {
```
phansys, line 258:
```suggestion
                    'callback' => static function(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $value): bool {
```

tests/Filter/CallbackFilterTest.php
```diff
@@ -108,4 +108,29 @@ public function testApplyMethod(): void
...
+        $filter->initialize('field_name', [
+            'callback' => static function ($builder, $alias, $field, $value) {
+                $builder->andWhere(sprintf('CUSTOM QUERY %s.%s', $alias, $field));
+                $builder->setParameter('value', $value);
+
+                return 1;
+            },
+        ]);
+
+        $filter->filter($builder, 'alias', 'field', 'myValue');
```
phansys, line 131: Shouldn't we expect a deprecation in this call?

phansys (changes requested):

tests/Filter/CallbackFilterTest.php
VincentLanglet: Yes indeed, I forgot to add one.

src/Filter/CallbackFilter.php
```diff
@@ -29,7 +29,18 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $value
             throw new \RuntimeException(sprintf('Please provide a valid callback option "filter" for field "%s"', $this->getName()));
         }
 
-        $this->active = \call_user_func($this->getOption('callback'), $queryBuilder, $alias, $field, $value);
+        $isActive = \call_user_func($this->getOption('callback'), $queryBuilder, $alias, $field, $value);
```
phansys, line 32: I know this is not directly related to this PR, but IMO we should add a test case using an array callback.

VincentLanglet: Like this
```
$filter->initialize('field_name', [
    'callback' => [$this, 'customCallback'],
]);
```
?

There is already one.

tests/Filter/CallbackFilterTest.php
phansys: Great! Didn't see it before.

src/Filter/CallbackFilter.php
phansys: Great! Didn't see it before.

phansys (dismissed):

src/Filter/CallbackFilter.php
```diff
@@ -29,7 +29,18 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $value
...
+        if (!\is_bool($isActive)) {
+            @trigger_error(
+                'Using another return type than boolean for the callback option is deprecated'
+                .' since sonata-project/doctrine-orm-admin-bundle 3.x and will throw an exception in version 4.0.',
+                E_USER_DEPRECATED
+            );
+
+            // NEXT_MAJOR: Throw an Exception instead.
```
franmomu, line 40: I think it's easier to define what exception would be now that we have context and just uncomment it in `4.x`. 

This would ease removing it in `4.x`, maybe someone else does it and has to read the whole thing to figure out.

VincentLanglet: It should be ok now

phansys (approved):

phansys (approved):

src/Filter/CallbackFilter.php
```diff
@@ -29,7 +29,22 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $value
...
+            @trigger_error(
+                'Using another return type than boolean for the callback option is deprecated'
+                .' since sonata-project/doctrine-orm-admin-bundle 3.x and will throw an exception in version 4.0.',
+                E_USER_DEPRECATED
+            );
+
+            // NEXT_MAJOR: Uncomment the following code instead of the deprecation.
+//            throw new \UnexpectedValueException(sprintf(
+//                'The callback should return a boolean, %s returned',
+//                \is_object($isActive) ? 'instance of "'.\get_class($value).'"' : '"'.\gettype($value).'"'
```
franmomu, line 43:
```suggestion
//                \is_object($isActive) ? 'instance of "'.\get_class($isActive).'"' : '"'.\gettype($isActive).'"'
```

VincentLanglet: Fixed

UPGRADE-3.x.md
```diff
@@ -1,6 +1,40 @@
...
+UPGRADE FROM 3.x to 3.x
+=======================
+
+### Sonata\DoctrineORMAdminBundle\Filter\CallbackFilter
+
+Deprecate the usage of a callback which does not return a boolean value.
+
+Previously, this was valid:
+```php
+function callback(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $value)
```
phansys, line 13:
```suggestion
static function callback(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $value)
```

UPGRADE-3.x.md
```diff
@@ -1,6 +1,40 @@
...
+    }
+
+    // ...
+
+    return true;
+}
+```
+To remove the deprecation, please update the code this way:
+```php
+function callback(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $value): bool
```
phansys, line 26:
```suggestion
static function callback(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $value): bool
```

phansys (approved):

UPGRADE-3.x.md
VincentLanglet: This code is out of a class in my example, I prefer to avoid using static then.

franmomu (approved):

PR #1161, opened 2020-10-18
Change default route name for association links to show.

VincentLanglet:
Related to https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1160

A global option can still be implemented, but IMHO `show` has more sens than `edit` for the default route name.

If accepted, the same PR should be done to https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle

### DIFF
--- a/tests/Model/ModelManagerTest.php
+++ b/tests/Model/ModelManagerTest.php
@@ -724,6 +724,54 @@ class ModelManagerTest extends TestCase
         $this->modelManager->getEntityManager(VersionedEntity::class);
     }
 
+    public function testGetNewFieldDescriptionInstance(): void
+    {
+        $modelManager = $this->getMockBuilder(ModelManager::class)
+            ->disableOriginalConstructor()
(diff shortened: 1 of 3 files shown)


wbloszyk (approved):

core23 (approved):

tests/Model/ModelManagerTest.php
```diff
@@ -724,6 +724,29 @@ public function testGetEntityManagerException(): void
         $this->modelManager->getEntityManager(VersionedEntity::class);
     }
 
+    public function testGetNewFieldDescriptionInstance(): void
```
phansys, line 727: Could we add a test for the case where the name is set explicitly to a value different than its default?

VincentLanglet: Done @phansys

wbloszyk (approved):

phansys (approved):

PR #1170, opened 2020-10-24
Remove fieldDescription deprecation

VincentLanglet:
## Subject

BC

## Changelog

```markdown
### Deprecated
- Instantiate a FieldDescription without passing the name as first argument
```

### DIFF
--- a/src/Admin/FieldDescription.php
+++ b/src/Admin/FieldDescription.php
@@ -20,11 +20,6 @@ use Sonata\AdminBundle\Admin\BaseFieldDescription;
  */
 class FieldDescription extends BaseFieldDescription
 {
-    public function __construct()
-    {
-        $this->parentAssociationMappings = [];
-    }
-
     public function setAssociationMapping($associationMapping)
     {
         if (!\is_array($associationMapping)) {
@@ -35,6 +30,7 @@ class FieldDescription extends BaseFieldDescription
 
         $this->type = $this->type ?: $associationMapping['type'];
         $this->mappingType = $this->mappingType ?: $associationMapping['type'];
+        // NEXT_MAJOR: Remove the next line.
         $this->fieldName = $associationMapping['fieldName'];
     }
 
@@ -74,6 +70,7 @@ class FieldDescription extends BaseFieldDescription
 
         $this->type = $this->type ?: $fieldMapping['type'];
         $this->mappingType = $this->mappingType ?: $fieldMapping['type'];
+        // NEXT_MAJOR: Remove the next line.
         $this->fieldName = $this->fieldName ?: $fieldMapping['fieldName'];
     }
 
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "doctrine/doctrine-bundle": "^1.8 || ^2.0",
         "doctrine/orm": "^2.5",
         "doctrine/persistence": "^1.3.4 || ^2.0",
-        "sonata-project/admin-bundle": "^3.71",
+        "sonata-project/admin-bundle": "^3.78.1",
         "sonata-project/exporter": "^1.11.0 || ^2.0",
         "sonata-project/form-extensions": "^0.1 || ^1.4",
         "symfony/config": "^4.4",
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -162,9 +162,7 @@ class ModelManager implements ModelManagerInterface, LockInterface
 
         [$metadata, $propertyName, $parentAssociationMappings] = $this->getParentMetadataForProperty($class, $name);
 
-        $fieldDescription = new FieldDescription();
-        $fieldDescription->setName($name);
-        $fieldDescription->setOptions($options);
+        $fieldDescription = new FieldDescription($name, $options);
         $fieldDescription->setParentAssociationMappings($parentAssociationMappings);
 
         if (isset($metadata->associationMappings[$propertyName])) {
--- a/tests/Admin/FieldDescriptionTest.php
+++ b/tests/Admin/FieldDescriptionTest.php
@@ -24,8 +24,7 @@ class FieldDescriptionTest extends TestCase
 {
     public function testOptions(): void
     {
-        $field = new FieldDescription();
-        $field->setOptions([
+        $field = new FieldDescription('name', [
             'template' => 'foo',
             'type' => 'bar',
             'misc' => 'foobar',
@@ -87,7 +86,7 @@ class FieldDescriptionTest extends TestCase
 
     public function testAssociationMapping(): void
     {
-        $field = new FieldDescription();
+        $field = new FieldDescription('name');
         $field->setAssociationMapping([
             'type' => 'integer',
             'fieldName' => 'position',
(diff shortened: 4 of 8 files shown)


src/Admin/FieldDescription.php
```diff
@@ -20,11 +20,6 @@
  */
 class FieldDescription extends BaseFieldDescription
 {
-    public function __construct()
-    {
-        $this->parentAssociationMappings = [];
```
VincentLanglet, line 25: This is already the default value. And the __construct() override disallow to pass a name as argument.

src/Admin/FieldDescription.php
```diff
@@ -35,6 +30,7 @@ public function setAssociationMapping($associationMapping)
 
         $this->type = $this->type ?: $associationMapping['type'];
         $this->mappingType = $this->mappingType ?: $associationMapping['type'];
+        // NEXT_MAJOR: Remove the next line.
```
VincentLanglet, line 33: The next line is uselss because the fieldName is supposed to be set to find the association mapping.
Plus in next major, the field name will be set with `new FieldDescription('FieldName')`

src/Admin/FieldDescription.php
```diff
@@ -74,6 +70,7 @@ public function setFieldMapping($fieldMapping)
 
         $this->type = $this->type ?: $fieldMapping['type'];
         $this->mappingType = $this->mappingType ?: $fieldMapping['type'];
+        // NEXT_MAJOR: Remove the next line.
```
VincentLanglet, line 73: The next line is uselss because the fieldName is supposed to be set to find the association mapping.
Plus in next major, the field name will be set with `new FieldDescription('FieldName')`

VincentLanglet:
What is wrong with static analysis ? @sonata-project/contributors

greg0ire:
You have the click the button :wink: 

Peek 2020-10-24 19-01

That issue is typical of circular dependencies, here is how we solve it at `doctrine/persistence`: https://github.com/doctrine/persistence/blob/6e297e5cd2d8aedda49a5bc0fa39c6682583dc67/.github/workflows/continuous-integration.yml#L8

Basically, I think you will have to let Composer know what version of `sonata-project/doctrine-orm-admin-bundle` otherwise `No version set (parsed as 1.0.0)`. 

@OskarStark I think the Github Action should fail when Composer fails, why is that not the case?

VincentLanglet:
> Basically, I think you will have to let Composer know what version of `sonata-project/doctrine-orm-admin-bundle` otherwise `No version set (parsed as 1.0.0)`.

How can I do this ?

greg0ire:
Use `COMPOSER_ROOT_VERSION: "3.24.0". Btw, I just noticed that you tagged the latest version of this packages as `3.24` instead of `3.24.0`… I'm not sure of the implications of that :sweat_smile:

VincentLanglet:
> Use `COMPOSER_ROOT_VERSION: "3.24.0"`. 

This won't be a long term fix, or we will have to update the version every time.

And why the composer would work for test but not for quality assurance ?
Is there something wrong with the action @OskarStark ?

> BTW, I just noticed that you tagged the latest version of this packages as `3.24` instead of `3.24.0`… I'm not sure of the implications of that 😅

Should be fix.

greg0ire:
> And why the composer would work for test but not for quality assurance ?

Good question. Also, why does it fail only now and not before? I notice that the Composer command is not the same, but that's it. Note that Composer 2 has been released, maybe it has to do with that?

> By setting this var you can specify the version of the root package, if it cannot be guessed from VCS info and is not present in composer.json.

This means the difference probably lies in that guess. For some reason, Composer does not have access at the branch name in the case of the QA workflow.

jordisala1991:
Does any of this helps? https://github.com/composer/composer/issues/8579#issuecomment-583071754

AFAIK it was only needed on block bundle 3.x because it has circular dependency with admin 3.x (and the provided solution is not working on every case, I copied it from the old travis config I will try to use somethinh from that comment).

jordisala1991:
Fixed here for blockbundle: https://github.com/sonata-project/SonataBlockBundle/pull/773

VincentLanglet:
> Fixed here for blockbundle: [sonata-project/SonataBlockBundle#773](https://github.com/sonata-project/SonataBlockBundle/pull/773)

I will wait for the dev-kit pr then :)

jordisala1991:
The problem is with this Pr: https://github.com/sonata-project/SonataAdminBundle/pull/6438

I gave a reason to not add that conflict, and it also creates cyclic dependency, so IMho we should revert that part and add the requirement here

jordisala1991:
On the block bundle 3.x is needed because of another cyclic dependency (and we should avoid them), fixed in 4.x

VincentLanglet:
> The problem is with this Pr: https://github.com/sonata-project/SonataAdminBundle/pull/6438
> 
> I gave a reason to not add that conflict, and it also creates cyclic dependency, so IMho we should revert that part and add the requirement here

It's not the same to add a requirement here or add a conflict to AdminBundle.

With a requirement, 3.25 orm cant be install without 3.78. But 3.78 can be install with 3.23 witch I wanted to avoid.

I dont understand why you call this a cyclic dependency. Its not because there is a conflict between admin and orm that admin needs orm.

jordisala1991:
> > The problem is with this Pr: https://github.com/sonata-project/SonataAdminBundle/pull/6438
> > 
> > I gave a reason to not add that conflict, and it also creates cyclic dependency, so IMho we should revert that part and add the requirement here
> 
> It's not the same to add a requirement here or add a conflict to AdminBundle.
> 
> With a requirement, 3.25 orm cant be install without 3.78. But 3.78 can be install with 3.23 witch I wanted to avoid.
> 
> I dont understand why you call this a cyclic dependency. Its not because there is a conflict between admin and orm that admin needs orm.

Something is wrong if the admin bundle (which does not require orm or mongodb persistence , because it is the other way around) is relying on a specific behavior of a specific version of the persistence. We either have a behavioral bc break or I cant understand. Is the admin bundle who defines the interfaces and the expected behavior, and the persistence who implements those contracts...

VincentLanglet:
> Something is wrong if the admin bundle (which does not require orm or mongodb persistence , because it is the other way around) is relying on a specific behavior of a specific version of the persistence. We either have a behavioral bc break or I cant understand. Is the admin bundle who defines the interfaces and the expected behavior, and the persistence who implements those contracts...

To resume the situation: 
Previously the admin bundle was always using a function A, but it was buggy.
Now, the admin is using a function A or B, depending on an option foo = true or false.

The right behavior is the following,
- If foo = true, the function B should be use.
- If foo = false, the function A should be use.

BUT, the persistence bundle was not passing the option, and the default value was true.
So, if you use an old version of persistence bundle, with the new version of admin bundle, you'll use function B even if the option foo is false (when you were using correctly the function A before).

In practice, the impact of the function A vs the function B is so minimal that I start to think that the best solution would be indeed to remove the conflict.

VincentLanglet:
Ready to review now @sonata-project/contributors

greg0ire (approved):

src/Admin/FieldDescription.php
```diff
@@ -35,6 +30,7 @@ public function setAssociationMapping($associationMapping)
 
         $this->type = $this->type ?: $associationMapping['type'];
```
phansys, line 31: Off topic: if these ternaries are intended to "catch" only null values (instead of other falsy values), we should use `??` instead.

phansys (approved):

PR #1183, opened 2020-11-05
DevKit updates for master branch

SonataCI:


jordisala1991:
That build need a bump for phpunit bridge to 5.1.4 at least.

There could be some errors to fix, but it will be easy. Probably this weekend I will try to fix remaining builds. If you can link them here I can find them easily

OskarStark (approved):

PR #477, opened 2015-08-10
Date filter should filter records for the whole day

pulzarraider:
This PR takes code from issues #346, #345, #285 from original authors: @cassianotartari, @sirian. I fixed some minor issues and added into this PR so it can be merged into current code without conflicts.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #346, #345, #285
| License       | MIT

PR #1189, opened 2020-11-07
Fix date handling through `DateRangeFilter`

phansys:
## Subject
Fix how dates from `DateRangeType` are handled by `DateRangeFilter`.
    
When the `DateRangeType` uses different values between "view_timezone" and "model_timezone" options, the values that are processed by `AbstractDateFilter::filter()` are `\DateTime` objects (start and end) holding the same timestamp than the view data, but represented in the timezone configured for the model data (thus, the offset of these objects is set to the model timezone and the date and time references are adjusted accordingly).
Previous to this commit, these date and time conversions were becoming imprecise since the `\DateTime` objects used in the filter were being manipulated to set the time part to "00:00:00" (in case of start) and to "23:59:59" (in case of end), which produces wrong results given the offset between the view and the model data is lost in these operations.
This commit removes the manipulation of the start time completely, and changes the manipulation done for the end time, by adding "23:59:59" hours to the received value, since both dates are expressed originally at "00:00:00" in the view timezone.

I am targeting this branch, because these changes are intended to respect BC.

Related to #477.

## Changelog

```markdown
### Fixed
- Setting wrong date and time references at `AbstractDateFilter::filter()`.
```

## To do

- [x] Update the tests;
- [x] Add the corresponding entries for the changelog;
- [x] Add details to the commit message and explain better what this PR fixes.

### DIFF
--- a/src/Filter/AbstractDateFilter.php
+++ b/src/Filter/AbstractDateFilter.php
@@ -65,13 +65,15 @@ abstract class AbstractDateFilter extends Filter
             }
 
             // date filter should filter records for the whole days
-            if (false === $this->time) {
-                if ($value['value']['start'] instanceof \DateTime) {
-                    $value['value']['start']->setTime(0, 0, 0);
-                }
-                if ($value['value']['end'] instanceof \DateTime) {
-                    $value['value']['end']->setTime(23, 59, 59);
-                }
+            if (false === $this->time && $value['value']['end'] instanceof \DateTime) {
+                // since the received `\DateTime` object  uses the model timezone to represent
+                // the value submitted by the view (which can use a different timezone) and this
+                // value is intended to contain a time in the begining of a date (IE, if the model
+                // object is configured to use UTC timezone, the view object "2020-11-07 00:00:00.0-03:00"
+                // is transformed to "2020-11-07 03:00:00.0+00:00" in the model object), we increment
+                // the time part by adding "23:59:59" in order to cover the whole end date and get proper
+                // results from queries like "o.created_at <= :date_end".
+                $value['value']['end']->modify('+23 hours 59 minutes 59 seconds');
             }
 
             // transform types
--- a/tests/Filter/DateRangeFilterTest.php
+++ b/tests/Filter/DateRangeFilterTest.php
@@ -20,7 +20,7 @@ use Sonata\DoctrineORMAdminBundle\Filter\DateRangeFilter;
 /**
  * @author Patrick Landolt <patrick.landolt@artack.ch>
  */
-class DateRangeFilterTest extends TestCase
+final class DateRangeFilterTest extends TestCase
 {
     public function testFilterEmpty(): void
     {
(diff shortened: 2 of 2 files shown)


tests/Filter/DateRangeFilterTest.php
```diff
@@ -119,4 +119,67 @@ public function testFilterEndDate(): void
...
+            ],
+        ]);
+
+        $this->assertTrue($filter->isActive());
+        $this->assertSame(['alias.field <= :field_name_1'], $builder->query);
+        $this->assertSame(['field_name_1' => $modelEndDateTime], $builder->parameters);
+        $this->assertSame($expectedEndDateTime->getTimestamp(), $modelEndDateTime->getTimestamp());
+    }
+
+    public function provideDates(): iterable
```
greg0ire, line 153:
```suggestion
    /**
     * @return \Generator<array{\DateTimeImmutable, \DateTime, \DateTimeZone}>
     */
    public function provideDates(): iterable
```

tests/Filter/DateRangeFilterTest.php
```diff
@@ -119,4 +119,67 @@ public function testFilterEndDate(): void
         $this->assertSame(['field_name_1' => $endDateTime], $builder->parameters);
         $this->assertTrue($filter->isActive());
     }
+
+    /**
+     * @dataProvider provideDates
+     */
+    public function testFilterEndDateCoversWholeDay(\DateTimeImmutable $expectedEndDateTime, \DateTime $viewEndDateTime, \DateTimeZone $modelTimeZone): void
+    {
```
greg0ire, line 127:
```suggestion
    public function testFilterEndDateCoversWholeDay(
        \DateTimeImmutable $expectedEndDateTime,
        \DateTime $viewEndDateTime,
        \DateTimeZone $modelTimeZone
    ): void {
```

greg0ire (commented):
Great commit message! Just a few nitpicks and this should be good to go!

greg0ire (approved):

VincentLanglet (approved):

VincentLanglet (commented):
Should we add a comment ?

src/Filter/AbstractDateFilter.php
```diff
@@ -66,11 +66,8 @@ public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $value
 
             // date filter should filter records for the whole days
             if (false === $this->time) {
-                if ($value['value']['start'] instanceof \DateTime) {
-                    $value['value']['start']->setTime(0, 0, 0);
-                }
                 if ($value['value']['end'] instanceof \DateTime) {
-                    $value['value']['end']->setTime(23, 59, 59);
+                    $value['value']['end']->modify('+23 hours 59 minutes 59 seconds');
```
phansys, line 70: Something like this @VincentLanglet?
```suggestion
                    // since the received `\DateTime` object  uses the model timezone to represent
                    // the value submitted by the view (which can use a different timezone) and this
                    // value is intended to contain a time in the beginning of a date (IE, if the model
                    // object is configured to use UTC timezone, the view object "2020-11-07 00:00:00.0-03:00"
                    // is transformed to "2020-11-07 03:00:00.0+00:00" in the model object), we increment
                    // the time part by adding "23:59:59" in order to cover the whole end date and get proper
                    // results from queries like "o.created_at <= :date_end".
                    $value['value']['end']->modify('+23 hours 59 minutes 59 seconds');
```

VincentLanglet: Seems great indeed !

phansys: It's added.

VincentLanglet (approved):

franmomu (approved):

PR #1192, opened 2020-11-08
Fix not to call an id method without object.

ggabrovski:
## Subject

In next major of **SonataAdminBundle** only object will be allowed as argument for the `id` method, so we need to skip calling this method without object.


VincentLanglet:
Since it's BC you can do it on the 3.x branch ;)

ggabrovski:
OK, I will close this PR and make new one to 3.x

PR #1193, opened 2020-11-08
Fix not to call an id method without object.

ggabrovski:
## Subject
In next major of **SonataAdminBundle** only object will be allowed as argument for the `id` method, so we need to skip calling this method without object.

### DIFF
--- a/src/Resources/views/Form/form_admin_fields.html.twig
+++ b/src/Resources/views/Form/form_admin_fields.html.twig
@@ -56,7 +56,7 @@ file that was distributed with this source code.
 {% block sonata_type_model_list_widget %}
     <div id="field_container_{{ id }}" class="field-container">
         <span id="field_widget_{{ id }}" class="field-short-description">
-            {% if sonata_admin.field_description.associationadmin.id(sonata_admin.value) %}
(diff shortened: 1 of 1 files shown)


VincentLanglet (approved):

src/Resources/views/Form/form_admin_fields.html.twig
```diff
@@ -56,7 +56,7 @@ file that was distributed with this source code.
 {% block sonata_type_model_list_widget %}
     <div id="field_container_{{ id }}" class="field-container">
         <span id="field_widget_{{ id }}" class="field-short-description">
-            {% if sonata_admin.field_description.associationadmin.id(sonata_admin.value) %}
+            {% if sonata_admin.value and sonata_admin.field_description.associationadmin.id(sonata_admin.value) %}
```
phansys, line 59:
```suggestion
            {% if sonata_admin.value and sonata_admin.field_description.associationadmin.id(sonata_admin.value) is not null %}
```

phansys (changes requested):

phansys (approved):

VincentLanglet (approved):

VincentLanglet:
Thanks @ggabrovski

PR #1194, opened 2020-11-08
Remove the deprecation from getNormalizedIdentifier

ggabrovski:
## Subject

Remove the deprecation in `getNormalizedIdentifier`, because of allow null as return type for `getUrlSafeIdentifier`, `getNormalizedIdentifier` and `id` in **SonataAdmin**.

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -4,6 +4,10 @@ UPGRADE 3.x
 UPGRADE FROM 3.24 to 3.25
 =========================
 
+### Sonata\DoctrineORMAdminBundle\Model\ModelManager
+
+Previously passing an object which is in state new or removed as argument 1 for `getNormalizedIdentifier()` was deprecated and would throw an exception in 4.0. Since throwing an exception is not allowed (and returning `null` is still allowed), the deprecation is removed.
+
 ### Added full support for `\DateTimeImmutable` in filters extending `Sonata\DoctrineORMAdminBundle\Filter\AbstractDateFilter`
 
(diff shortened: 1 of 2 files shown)


phansys (commented):
We should make this decision clear at `UPGRADE-3.x.md`.

VincentLanglet:
> We should make this decision clear at `UPGRADE-3.x.md`.

Previously we wrote
```
UPGRADE FROM 3.19 to 3.20
=========================

### Sonata\DoctrineORMAdminBundle\Model\ModelManager

Deprecated passing `null` or an object which is in state new or removed as argument 1 for `getNormalizedIdentifier()`.
```

What's phansys expect is to add something like 
```
UPGRADE FROM 3.x to 3.x
=======================

### Sonata\DoctrineORMAdminBundle\Model\ModelManager

Previously passing an object which is in state new or removed as argument 1 for `getNormalizedIdentifier()` was deprecated and would throw an exception in 4.0. Since throwing an exception is not allowed (and returning `null` is still allowed), the deprecation is removed.
```

You confirm @phansys ?

phansys:
>You confirm @phansys ?

I agree with that message. Thank you.

VincentLanglet:
@ggabrovski Do you have some time to add a line to the Upgrade note ?

You also need to rebase 3.x to fix the build.

ggabrovski:
I made it, sorry for the mistake.

VincentLanglet (approved):

phansys (approved):

phansys:
Thank you @ggabrovski!

PR #1190, opened 2020-11-07
Add support for `\DateTimeImmutable` at `AbstractDateFilter::filter()`

phansys:
## Subject
Add support for `\DateTimeImmutable` at `AbstractDateFilter::filter()`.

I am targeting this branch, because this change respects BC.

## Changelog

```markdown
### Added
- Support for `\DateTimeImmutable` at `AbstractDateFilter::filter()`.
```

PR #1196, opened 2020-11-09
Add upgrade note for #1190

phansys:
## Subject
Add missing upgrade note for #1190.

I am targeting this branch, because this change does not affect BC.

Related to #1190.

I'm not adding a changelog about this change since the related PR is not present in any release yet.

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -4,6 +4,19 @@ UPGRADE 3.x
 UPGRADE FROM 3.x to 3.x
 =======================
 
+### Added full support for `\DateTimeImmutable` in filters extending `Sonata\DoctrineORMAdminBundle\Filter\AbstractDateFilter`
+
+- `Sonata\DoctrineORMAdminBundle\Filter\DateFilter`
+- `Sonata\DoctrineORMAdminBundle\Filter\DateRangeFilter`
+- `Sonata\DoctrineORMAdminBundle\Filter\DateTimeFilter`
+- `Sonata\DoctrineORMAdminBundle\Filter\DateTimeRangeFilter`
+- `Sonata\DoctrineORMAdminBundle\Filter\TimeFilter`
+
+Previous to this change, only the instances of `\DateTime` were manipulated in these
+filters to set the time under determined circumstances. If you are using them with instances
+of `\DateTimeImmutable`, be aware of this change in order to confirm if you must update
+your implementation.
+
 ### Sonata\DoctrineORMAdminBundle\Filter\CallbackFilter
 
 Deprecate the usage of a callback which does not return a boolean value.


VincentLanglet (approved):

franmomu:
I thought the `UPGRADE-x.md` file was meant to add things a user has to do in order to upgrade to version `x`, I see this more in the changelog of the original PR.

phansys:
I understand your point @franmomu.
My concern is about the change in the behavior for cases where an instance of `\DateTimeImmutable` is already passed in projects using previous versions.
By instance, the time manipulation done at `AbstractDateFilter::filter()` was not affecting the passed value if it was an instance of `\DateTimeImmutable`, but now it does:

https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/d6fc51cc14530dc30145f30873d0e4778a8d4b88/src/Filter/AbstractDateFilter.php#L67-L77

franmomu:
Ah, that makes sense! but should the upgrade note reflect these changes? otherwise looks like just support was added (it is what happened to me).

VincentLanglet:
We should be more clear about what we should put to the UPGRADE note.
Same issue here https://github.com/sonata-project/SonataAdminBundle/pull/6585

If we add new feature to upgrade note, adding them to the changelog seems useless.

To me, we should just add to the upgrade note things that will need change from the developer like deprecations (or mandatory BC-break)
Bug fix and new feature shouldn't be added.

phansys:
Personally, I see the upgrade document as the first place where I should go to check if something related to a package stopped to work or if some previous behavior seems to be changed. If I can't find something related there, I check the changelog.
>Ah, that makes sense! but should the upgrade note reflect these changes?

I agree, I'll try to update the message today.

VincentLanglet:
> Personally, I see the upgrade document as the first place where I should go to check if something related to a package stopped to work or if some previous behavior seems to be changed.

If something need to be added to the upgrade document, there should always be with a "HOW-TO" then.

phansys:
>If something need to be added to the upgrade document, there should always be with a "HOW-TO" then.

Should we take the following premise as part of the "HOW-TO" you're proposing?
>If you are using these filters with instances of `\DateTimeImmutable`, be aware of this change in order to determine if you must update your implementation.

VincentLanglet:
> > If something need to be added to the upgrade document, there should always be with a "HOW-TO" then.
> 
> Should we take the following premise as part of the "HOW-TO" you're proposing?
> 
> > If you are using these filters with instances of `\DateTimeImmutable`, be aware of this change in order to determine if you must update your implementation.

I would say yes.

But in this case, I don't understand you because `\DateTimeImmutable` wasn't supported yet.
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1190

phansys:
>But in this case, I don't understand you because \DateTimeImmutable wasn't supported yet.
So why someone would use it with \DateTimeImmutable.

I'd say that they were already supported (the filter doesn't have any check forbidding the type), but not in a fully or a proper way (they weren't receiving the same treatment as `\DateTime` instances).
That's why I consider this as a MINOR instead of PATCH, but we could discuss about this point.

VincentLanglet:
> I'd say that they were already supported (the filter doesn't have any check forbidding the type), but not in a fully or a proper way (they weren't receiving the same treatment as `\DateTime` instances).
> That's why I consider this as a MINOR instead of PATCH, but we could discuss about this point.

Ok.

phansys:
>Ah, that makes sense! but should the upgrade note reflect these changes? otherwise looks like just support was added (it is what happened to me).

The note was updated @franmomu :+1:

VincentLanglet (approved):

franmomu (dismissed):

VincentLanglet:
You may need to rebase

VincentLanglet (approved):

OskarStark (approved):

OskarStark:
@franmomu please give us a final review here, thanks

franmomu (approved):

franmomu:
Thanks @phansys!

PR #1198, opened 2020-11-12
Use correctly phpstan-symfony

VincentLanglet:
Pedantic

I'm not sure, but it seems like it was installed but not used ?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,6 +54,7 @@
         "sonata-project/admin-bundle-persistency-layer": "1.0.0"
     },
     "require-dev": {
+        "phpstan/extension-installer": "^1.0",
         "phpstan/phpstan": "^0.12.52",
         "phpstan/phpstan-symfony": "^0.12.10",
         "psalm/plugin-symfony": "^2.0",


OskarStark:
can use phpstan/extension-installer instead?

VincentLanglet:
@OskarStark @franmomu started to do it without the extension-installer
See https://github.com/sonata-project/SonataTranslationBundle/blob/2.x/phpstan.neon.dist#L3

IMHO we should do the same way for all projects. But we can use extension-installer everywhere instead.
Updating all the bundle with phpstan should be easy.

OskarStark:
> MHO we should do the same way for all projects. But we can use extension-installer everywhere instead. 

I would vote for this. WDYT @franmomu ?

franmomu:
> > MHO we should do the same way for all projects. But we can use extension-installer everywhere instead.
> 
> I would vote for this. WDYT @franmomu ?

yeah, I'm fine with the extension installer, I just did it this way to not require another package.

VincentLanglet:
Done @franmomu @OskarStark 

I also done https://github.com/sonata-project/SonataAdminBundle/pull/6592

greg0ire (commented):
I think you added `phpstan/phpstan-symfony` in https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1172, but it does not seem to be needed for now. Was it "just in case"? Because I think we might as well remove it…

VincentLanglet:
> Was it "just in case"?

Not sure, kinda maybe. Isn't something like a bonus ?
It works without, but with we may detect some supplementary issue.

It's like phpstan-phpunit, without phpstan works well, but with we can detect some issues in our tests.

greg0ire:
>  It works without, but with we may detect some supplementary issue.

I think it's more about not detecting extra issues, for instance with the console component, phpstan might complain that we are using an input argument as an integer without knowing, because the return type is `int|string|array<string>|whatever`. Or if you do `$container->get('some_service')`, phpstan might complain again because it does not know the type .

I don't think you can find a feature we use or are likely to use in this bundle in [the list of features improved by this extension](https://github.com/phpstan/phpstan-symfony#phpstan-symfony-framework-extensions-and-rules). The only 2 features that seem to be about discovering issues are:

Notifies you when you try to get an unregistered service from the container.
Notifies you when you try to get a private service from the container.

We don't get services from the container in this bundle, and are unlikely to do so in the future as it is a bad practice.

But maybe I'm missing the bigger picture? Is the plan to add this extension to all packages that end with `Bundle`? It does not have a lot of dependencies so that would probably not be that big of a deal.

VincentLanglet:
> But maybe I'm missing the bigger picture? Is the plan to add this extension to all packages that end with `Bundle`?It does not have a lot of dependencies so that would probably not be that big of a deal. 

This package also provide stubs, in order to fix/improve the phpdoc from Symfony. Since there is more and more stub provided, if we currently not using them, one day we will.

Plus, We're using psalm-symfony, so I find weird that we're not using phpstan-symfony.

I think indeed it could be simpler to add symfony plugin to every symfony related project.

greg0ire (approved):
Sounds sensible enough.

core23 (approved):

greg0ire:
Thanks @VincentLanglet !

PR #1199, opened 2020-11-12
Add deprecation for filter method

VincentLanglet:
## Subject

The method `filter` won't be in the interface anymore so we can restrict the typehint.

## Changelog

```markdown
### Deprecated
- Deprecate passing an instance of `ProxyQueryInterface` which is not an instance of `Sonata\DoctrineORMAdminBundle\Datagrid::ProxyQuery` as argument 1 to the `Sonata\DoctrineORMAdminBundle\Filter\Filter::filter()` method.
```

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -1,6 +1,12 @@
 UPGRADE 3.x
 ===========
 
+### Sonata\DoctrineORMAdminBundle\Filter\Filter
+
+Deprecate passing an instance of `Sonata\AdminBundle\Datagrid\ProxyQueryInterface`
+which is not an instance of `Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery` as
+argument 1 to the `Sonata\DoctrineORMAdminBundle\Filter\Filter::filter()` method.
+
 UPGRADE FROM 3.25 to 3.26
 =========================
 
--- a/src/Filter/AbstractDateFilter.php
+++ b/src/Filter/AbstractDateFilter.php
@@ -13,13 +13,14 @@ declare(strict_types=1);
 
 namespace Sonata\DoctrineORMAdminBundle\Filter;
 
-use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
+use Sonata\AdminBundle\Datagrid\ProxyQueryInterface as BaseProxyQueryInterface;
 use Sonata\AdminBundle\Form\Type\Filter\DateRangeType;
 use Sonata\AdminBundle\Form\Type\Filter\DateTimeRangeType;
 use Sonata\AdminBundle\Form\Type\Filter\DateTimeType;
 use Sonata\AdminBundle\Form\Type\Filter\DateType;
 use Sonata\AdminBundle\Form\Type\Operator\DateOperatorType;
 use Sonata\AdminBundle\Form\Type\Operator\DateRangeOperatorType;
+use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface;
 
 abstract class AbstractDateFilter extends Filter
 {
@@ -47,8 +48,19 @@ abstract class AbstractDateFilter extends Filter
      */
     protected $time = false;
 
-    public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $value)
+    public function filter(BaseProxyQueryInterface $query, $alias, $field, $value)
     {
+        /* NEXT_MAJOR: Remove this deprecation and update the typehint */
+        if (!$query instanceof ProxyQueryInterface) {
+            @trigger_error(sprintf(
+                'Passing %s as argument 1 to %s() is deprecated since sonata-project/doctrine-orm-admin-bundle 3.x'
+                .' and will throw a \TypeError error in version 4.0. You MUST pass an instance of %s instead.',
+                \get_class($query),
+                __METHOD__,
+                ProxyQueryInterface::class
+            ));
+        }
+
         // check data sanity
         if (!$value || !\is_array($value) || !\array_key_exists('value', $value)) {
             return;
@@ -85,27 +97,27 @@ abstract class AbstractDateFilter extends Filter
             // default type for range filter
             $value['type'] = !isset($value['type']) || !is_numeric($value['type']) ? DateRangeOperatorType::TYPE_BETWEEN : $value['type'];
 
-            $startDateParameterName = $this->getNewParameterName($queryBuilder);
-            $endDateParameterName = $this->getNewParameterName($queryBuilder);
+            $startDateParameterName = $this->getNewParameterName($query);
+            $endDateParameterName = $this->getNewParameterName($query);
 
             if (DateRangeOperatorType::TYPE_NOT_BETWEEN === $value['type']) {
-                $this->applyWhere($queryBuilder, sprintf('%s.%s < :%s OR %s.%s > :%s', $alias, $field, $startDateParameterName, $alias, $field, $endDateParameterName));
+                $this->applyWhere($query, sprintf('%s.%s < :%s OR %s.%s > :%s', $alias, $field, $startDateParameterName, $alias, $field, $endDateParameterName));
             } else {
                 if ($value['value']['start']) {
-                    $this->applyWhere($queryBuilder, sprintf('%s.%s %s :%s', $alias, $field, '>=', $startDateParameterName));
+                    $this->applyWhere($query, sprintf('%s.%s %s :%s', $alias, $field, '>=', $startDateParameterName));
                 }
 
                 if ($value['value']['end']) {
-                    $this->applyWhere($queryBuilder, sprintf('%s.%s %s :%s', $alias, $field, '<=', $endDateParameterName));
+                    $this->applyWhere($query, sprintf('%s.%s %s :%s', $alias, $field, '<=', $endDateParameterName));
                 }
             }
 
             if ($value['value']['start']) {
-                $queryBuilder->setParameter($startDateParameterName, $value['value']['start']);
+                $query->getQueryBuilder()->setParameter($startDateParameterName, $value['value']['start']);
             }
 
             if ($value['value']['end']) {
-                $queryBuilder->setParameter($endDateParameterName, $value['value']['end']);
+                $query->getQueryBuilder()->setParameter($endDateParameterName, $value['value']['end']);
             }
         } else {
             if (!$value['value']) {
@@ -125,33 +137,33 @@ abstract class AbstractDateFilter extends Filter
 
             // null / not null only check for col
             if (\in_array($operator, ['NULL', 'NOT NULL'], true)) {
-                $this->applyWhere($queryBuilder, sprintf('%s.%s IS %s ', $alias, $field, $operator));
+                $this->applyWhere($query, sprintf('%s.%s IS %s ', $alias, $field, $operator));
 
                 return;
             }
 
-            $parameterName = $this->getNewParameterName($queryBuilder);
+            $parameterName = $this->getNewParameterName($query);
 
             // date filter should filter records for the whole day
             if (false === $this->time && DateOperatorType::TYPE_EQUAL === $value['type']) {
-                $this->applyWhere($queryBuilder, sprintf('%s.%s %s :%s', $alias, $field, '>=', $parameterName));
-                $queryBuilder->setParameter($parameterName, $value['value']);
+                $this->applyWhere($query, sprintf('%s.%s %s :%s', $alias, $field, '>=', $parameterName));
+                $query->getQueryBuilder()->setParameter($parameterName, $value['value']);
 
-                $endDateParameterName = $this->getNewParameterName($queryBuilder);
-                $this->applyWhere($queryBuilder, sprintf('%s.%s %s :%s', $alias, $field, '<', $endDateParameterName));
+                $endDateParameterName = $this->getNewParameterName($query);
+                $this->applyWhere($query, sprintf('%s.%s %s :%s', $alias, $field, '<', $endDateParameterName));
                 if ('timestamp' === $this->getOption('input_type')) {
                     $endValue = strtotime('+1 day', $value['value']);
                 } else {
                     $endValue = clone $value['value'];
                     $endValue->add(new \DateInterval('P1D'));
                 }
-                $queryBuilder->setParameter($endDateParameterName, $endValue);
+                $query->getQueryBuilder()->setParameter($endDateParameterName, $endValue);
 
                 return;
             }
 
-            $this->applyWhere($queryBuilder, sprintf('%s.%s %s :%s', $alias, $field, $operator, $parameterName));
-            $queryBuilder->setParameter($parameterName, $value['value']);
+            $this->applyWhere($query, sprintf('%s.%s %s :%s', $alias, $field, $operator, $parameterName));
+            $query->getQueryBuilder()->setParameter($parameterName, $value['value']);
         }
     }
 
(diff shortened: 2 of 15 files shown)


src/Filter/AbstractDateFilter.php
```diff
@@ -49,6 +50,17 @@ abstract class AbstractDateFilter extends Filter
 
     public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $value)
     {
+        /* NEXT_MAJOR: Remove this deprecation and update the typehing */
+        if ($queryBuilder instanceof ProxyQuery) {
+            @trigger_error(sprintf(
+                'Passing %s as argument 1 to %s() is deprecated since sonata-project/doctrine-orm-admin-bundle 3.x'
+                .' and will throw a \TypeError in version 4.0. You must pass an instance of %s instead.',
```
phansys, line 57:
```suggestion
                .' and will throw a \TypeError error in version 4.0. You MUST pass an instance of %s instead.',
```

UPGRADE-3.x.md
```diff
@@ -4,6 +4,12 @@ UPGRADE 3.x
 UPGRADE FROM 3.x to 3.x
 =======================
 
+### Sonata\DoctrineORMAdminBundle\Filter\Filter
+
+Deprecate passing an instance of `ProxyQueryInterface` which is not an instance of
+`Sonata\DoctrineORMAdminBundle\Datagrid::ProxyQuery` as argument 1 to the
```
phansys, line 10:
```suggestion
`Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery` as argument 1 to the
```

phansys (changes requested):
Will we make a contract with the concrete class `Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery`?
I don't understand why.

VincentLanglet:
> Will we make a contract with the concrete class `Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery`?
> I don't understand why.

Because we're already kinda doing it by using method like
- `setParameter`
- `expr`
- `entityJoin` (which will be removed from the interface) 
- `getUniqueParameterId` (which will be removed from the interface)

UPGRADE-3.x.md
```diff
@@ -4,6 +4,12 @@ UPGRADE 3.x
 UPGRADE FROM 3.x to 3.x
 =======================
 
+### Sonata\DoctrineORMAdminBundle\Filter\Filter
+
+Deprecate passing an instance of `ProxyQueryInterface` which is not an instance of
```
phansys, line 9:
```suggestion
Deprecate passing an instance of `Sonata\AdminBundle\Datagrid\ProxyQueryInterface` which is not an instance of
```

phansys (changes requested):
Should we not create a local interface (maybe extending `Sonata\AdminBundle\Datagrid\ProxyQueryInterface`) with the methods that are used here? Or it is not worth?

VincentLanglet:
> Should not we create a local interface (maybe extending `Sonata\AdminBundle\Datagrid\ProxyQueryInterface`) with the methods that are used here? Or it is not worth?

Do you use these filters with another ProxyQuery ?
I don't. And I'm not sure someone does.

I would wait for someone creating an issue with a real use case to create this interface instead.

UPGRADE-3.x.md
VincentLanglet: Done

phansys:
>Do you use these filters with another ProxyQuery ?

Yes, indeed I use a variety of custom model managers with their own proxy queries (and so).
In my case, these model managers are for external data sources (mainly ReST APIs). Currently, since I have not enough time to plan a deep refactoring, the classes I use to support these managers are extending or decorating classes from the ORM manager in order to simplify the amount of code I need (I just override the parts I need).

VincentLanglet:
> > Do you use these filters with another ProxyQuery ?
> 
> Yes, indeed I use a variety of custom model managers with their own proxy queries (and so).

How are defined your ProxyQueries ?
The ProxyQuery is currently not marked as final, so if you're extending them, there is no issue.

The day we will mark this class as final, we should add a `DoctrineORMProxyQueryInterface`.

> In my case, these model managers are for external data sources (mainly ReST APIs). Currently, since I have not enough time to plan a deep refactoring, the classes I use to support these managers are extending or decorating classes from the ORM manager in order to simplify the amount of code I need (I just override the parts I need).

ModelManager will be final in next Major (https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Model/ModelManager.php#L43).
Will it impact you ?

src/Filter/Filter.php
```diff
@@ -24,6 +24,17 @@ abstract class Filter extends BaseFilter
...
+    /**
+     * NEXT_MAJOR change $queryBuilder type for ProxyQuery.
+     *
+     * Apply the filter to the QueryBuilder instance.
+     *
+     * @param string  $alias
+     * @param string  $field
+     * @param mixed[] $value
+     */
+    abstract public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $value);
```
core23, line 36: This is a BC break. You MUST implement the new method, if you create a custom filter

core23 (changes requested):

src/Filter/Filter.php
VincentLanglet: This is not, because the method is not new. It's in the FilterInterface.

It will be removed in the next major, but will still be used in this persistence bundle.

phansys:
>How are defined your ProxyQueries ?

My `ProxyQuery` classes are currently extending `Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery` (but please, be aware that they are not related in any way to the ORM manager). I'm currently using inheritance just for laziness, but if they were defined correctly, they should not inherit from other model managers.

>ModelManager will be final in next Major (https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Model/ModelManager.php#L43).
Will it impact you ?

Sure, I'm aware of this and that is right.

VincentLanglet:
> > How are defined your ProxyQueries ?
> 
> My `ProxyQuery` classes are currently extending `Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery`

Then you won't have any issue with this PR

> (but please, be aware that they are not related in any way to the ORM manager). I'm currently using inheritance just for laziness, but if they were defined correctly, they should not inherit from other model managers.

If they are not related in any way to the ORM manager, you should have your own bundle. If you override everything, you could have use SonataDoctrineMongoDBAdminBundle too 😅

First, for our filter, the methods 
```
setParameter
expr
entityJoin (which will be removed from the interface)
getUniqueParameterId
```
Are needed. Do you override all of them ?
Then if I look at others methods of this bundle, there is a lot of `QueryBuilder` specific methods that we use.
In `SonataDoctrineMongoDBAdminBundle`, @franmomu is doing a lot of work to improve the psalm level. 
If we do the same here, we will have error because we're using method which are missing in ProxyQueryInterface.
So we need to find a solution, and restricting the ProxyQuery type seems to be a good solution to me.

phansys:
>If you override everything, you could have use SonataDoctrineMongoDBAdminBundle too

Yes, I could override any existing implementation.

>First, for our filter, the methods
>
>setParameter
expr
entityJoin (which will be removed from the interface)
getUniqueParameterId
Are needed. Do you override all of them ?

In my use case, I'm overriding `setParameter()` and `entityJoin()` (although I think the last one is not required).

VincentLanglet:
Since the ProxyQuery is not final ; do you still have a blocking point ? @phansys

phansys:
> Since the ProxyQuery is not final ; do you still have a blocking point ? @phansys

My intention is not to set a blockint point based on my specific and invalid use case. But I think every extension point we provide must be based on a interface.
So, to resume my opinion, if we consider `ProxyQuery` an extension point, we should provide an interface defining the methods we consider as part of that contract. The implementation must be final and must implement that interface.

VincentLanglet:
> So, to resume my opinion, if we consider `ProxyQuery` an extension point, we should provide an interface defining the methods we consider as part of that contract. The implementation must be final and must implement that interface.

I would agree with this, theorically, but
- There is no `QueryBuilderInterface` we can extends.
- If we declare our `DoctrineProxyQueryInterface` with all the methods from the `QueryBuilder`, we'll have to add them to our `ProxyQuery` since `__call` is not enough to implement an Interface.

We're currently coding in this repository like the ProxyQueryInterface was a ProxyQuery, (in the same way, @franmomu is adding some `assert($proxyQuery instanceof DocumentManager)` everywhere in another persistence bundle) so let's just update the code to reflect the reality. It's a big lie to keep the ProxyQueryInterface typehint when we're not respecting this interface.

If not, this commit: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/commit/4f63cc946147f95b791a60a481b2227baa8a076a#diff-d72576a660958913b8abcaf203e1b6c855f9f4bdfd93c5c2715fd4c4e96e3d70
Should have been a BC-break, because the filter was using 
```
$queryBuilder->expr()->notIn()
$queryBuilder->expr()->in()
```
And not use some new methods 
```
$queryBuilder->expr()->andX()
$queryBuilder->expr()-> isNotNull()
```
Which may not exists. Proof is this didn't exist in the QueryBuilder used for the tests.

And I'm pretty sure I can find bigger BC-break.

This PR doesn't forbid to do all the interface-work later. It's just here to start "saying the truth".

phansys:
>This PR doesn't forbid to do all the interface-work later. It's just here to start "saying the truth".

I agree with the path we are trying to follow to provide transparency. I just think that we should be clear about the lack of a proper interface for these methods. So, I think we must provide `TODO` or `NEXT_MAJOR` comments explaining what we should do to fix the already known issues later.

VincentLanglet:
> > This PR doesn't forbid to do all the interface-work later. It's just here to start "saying the truth".
> 
> I agree with the path we are trying to follow to provide transparency. I just think that we should be clear about the lack of a proper interface for these methods. So, I think we must provide `TODO` or `NEXT_MAJOR` comments explaining what we should do to fix the already known issues later.

An issue since the ideal place for this: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1216

VincentLanglet:
Friendly ping @core23 @phansys @sonata-project/contributors

src/Filter/AbstractDateFilter.php
```diff
@@ -49,6 +50,17 @@ abstract class AbstractDateFilter extends Filter
 
     public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $value)
     {
+        /* NEXT_MAJOR: Remove this deprecation and update the typehint */
+        if ($queryBuilder instanceof ProxyQuery) {
```
core23, line 54: You should invert this check (and the others)

core23 (changes requested):

src/Filter/AbstractDateFilter.php
VincentLanglet: Yes indeed ! It's fixed, thanks.

core23 (dismissed):

phansys:
> Friendly ping @core23 @phansys @sonata-project/contributors

Before considering this PR RTM, I think we must take a decision about the situation exposed at #1216 first. Otherwise, we will be triggering a deprecation that maybe later may change. Not a big deal, but if is possible, I'd like to avoid changes in what we document as upgrade notes.

VincentLanglet:
> > Friendly ping @core23 @phansys @sonata-project/contributors
> 
> Before considering this PR RTM, I think we must take a decision about the situation exposed at #1216 first. Otherwise, we will be triggering a deprecation that maybe later may change. Not a big deal, but if is possible, I'd like to avoid changes in what we document as upgrade notes.

I'll first merge https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1218
And then update this PR.

VincentLanglet:
@phansys The PR is ready to merge, IMHO

src/Filter/BooleanFilter.php
```diff
@@ -23,35 +24,46 @@
...
+
+        if (!$data || !\is_array($data) || !\array_key_exists('type', $data) || !\array_key_exists('value', $data)) {
             return;
         }
 
-        if (\is_array($value['value'])) {
-            $values = [];
-            foreach ($value['value'] as $v) {
+        if (\is_array($data['value'])) {
+            $datas = [];
```
franmomu, line 45: I think this is better as `$values`.

src/Filter/ChoiceFilter.php
```diff
@@ -54,8 +66,19 @@ public function getRenderSettings()
         ]];
     }
 
-    private function filterWithMultipleValues(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $data = []): void
+    private function filterWithMultipleValues(BaseProxyQueryInterface $query, string $alias, string $field, array $data = []): void
     {
+        /* NEXT_MAJOR: Remove this deprecation and update the typehint */
+        if (!$query instanceof ProxyQueryInterface) {
```
franmomu, line 72: It's not necessary to trigger a deprecation in this private method, just the `NEXT_MAJOR` comment.

src/Filter/ChoiceFilter.php
```diff
@@ -65,51 +88,62 @@ private function filterWithMultipleValues(ProxyQueryInterface $queryBuilder, str
...
+            $this->applyWhere($query, $query->getQueryBuilder()->expr()->orX()->addMultiple($orConditions));
+            $query->getQueryBuilder()->setParameter($parameterName, $data['value']);
         }
     }
 
-    private function filterWithSingleValue(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $data = []): void
+    private function filterWithSingleValue(BaseProxyQueryInterface $query, string $alias, string $field, array $data = []): void
     {
+        /* NEXT_MAJOR: Remove this deprecation and update the typehint */
+        if (!$query instanceof ProxyQueryInterface) {
```
franmomu, line 118: It's not necessary to trigger a deprecation in this private method, just the `NEXT_MAJOR` comment.

src/Filter/BooleanFilter.php
VincentLanglet: Done

src/Filter/ChoiceFilter.php
VincentLanglet: Indeed, I also discovered I also added `NEXT_MAJOR` comments for something I did on master https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1011

src/Filter/ChoiceFilter.php
```diff
@@ -54,62 +66,70 @@ public function getRenderSettings()
         ]];
     }
 
-    private function filterWithMultipleValues(ProxyQueryInterface $queryBuilder, string $alias, string $field, array $data = []): void
+    /**
+     * NEXT_MAJOR: Change the typehint to ProxyQueryInterface.
+     */
+    private function filterWithMultipleValues(BaseProxyQueryInterface $query, string $alias, string $field, array $data = []): void
```
franmomu, line 72: It's not related to this PR, but `array $data = []` (the default value) I think it doesn't make sense since next line is checking `count($data['value'])`, it would be nice to change this array to a VO, I'll create an issue.

franmomu (approved):

src/Filter/ChoiceFilter.php
VincentLanglet: Nice catch. This PR is big enough indeed, so I would prefer an issue indeed. :)

VincentLanglet:
@phansys Is it ok for you now ? I'm using the newly created interface instead of ProxyQuery.
Your approval is needed to merge this.

@greg0ire would it be conceivable to dismiss the change requested since I followed them ?
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1199#pullrequestreview-530575230

greg0ire:
I suppose so.

PR #1200, opened 2020-11-12
Refactor filter tests to use a Mock instead of a fake QueryBuilder

VincentLanglet:
This will allow to typehint ProxyQuery constructor

Pedantic


greg0ire:
There is a conflict

VincentLanglet:
> There is a conflict

Indeed, it was with the previous merged PR. Fixed now :)

OskarStark:
Typo in PR title and first commit:

Refacto-> Refactor

VincentLanglet:
> Typo in PR title and first commit:
> 
> Refacto-> Refactor

Fixed

tests/Datagrid/PagerTest.php
```diff
@@ -54,50 +52,4 @@ public function testComputeNbResultForCompositeId(string $className): void
...
-        $proxyQuery->setDistinct($distinct);
-
-        $queryBuilder->expects($this->once())
-            ->method('resetDQLPart')
-            ->willReturn($proxyQuery);
-
-        $pager = new Pager();
-        $pager->setQuery($proxyQuery);
-        $pager->computeNbResult();
-    }
```
greg0ire, line 102: Why is all this removed?

greg0ire (commented):
> This will allow to typehint ProxyQuery constructor

Sorry, I don't see why we can't use a type declaration right now?

VincentLanglet:
> > This will allow to typehint ProxyQuery constructor
> 
> Sorry, I don't see why we can't use a type declaration right now?

The `new QueryBuilder()` used in test is a fake QueryBuilder
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/tests/Filter/QueryBuilder.php

Adding a type declaraiton to the constructor will return an error for all our current TestFilter.

tests/Datagrid/PagerTest.php
VincentLanglet: - This test is using a `Sonata\DoctrineORMAdminBundle\Tests\Filter\QueryBuilder` instead of a Doctrine one.
- This test, which seems complicated, is not really testing something (or something useful), there are some `expects` about the implementation, nothing about the result of this method.
- There is already another test on the `computeNbResult` which is doing better assertions.

greg0ire: Please mention this in the body of your commit message.

VincentLanglet: Done @greg0ire

greg0ire (approved):

franmomu (approved):

PR #1207, opened 2020-11-18
Add NotEmptyFilter

VincentLanglet:
## Subject 

I'm adding a `inverse` option in order to use the EmptyFilter as a NotEmptyFilter.
You may want to have a filter called `Has foo`, instead of a `Has no foo` one.

## Changelog

```markdown
### Added
- Added an option `inverse` for the EmptyFilter.
```

### DIFF
--- a/docs/reference/filter_field_definition.rst
+++ b/docs/reference/filter_field_definition.rst
@@ -192,6 +192,8 @@ Empty
         }
     }
 
+The ``inverse`` option can be used to filter values that are not empty.
+
 Advanced usage
 --------------
 
--- a/src/Filter/EmptyFilter.php
+++ b/src/Filter/EmptyFilter.php
@@ -31,19 +31,18 @@ final class EmptyFilter extends Filter
             return;
         }
 
-        if (BooleanType::TYPE_YES === (int) $value['value']) {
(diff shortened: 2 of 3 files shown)


docs/reference/filter_field_definition.rst
```diff
@@ -192,6 +192,8 @@ Empty
         }
     }
 
+This filter has an additional option ``inverse`` in order to be used as a ``NotEmptyFilter``.
```
phansys, line 195:
```suggestion
The ``inverse`` option can be used to filter values that are not empty.
```
BTW, we should add an example using this option.

VincentLanglet: I made the change.

But I don't think an example would be useful

This option can be used like any other filter option.
```
final class PostAdmin extends AbstractAdmin
    {
        {
            $datagridMapper
                ->add('deleted', EmptyFilter::class, ['field_name' => 'deletedAt', 'inverse' => true]);
        }
    }
```

phansys (approved):

core23 (approved):

VincentLanglet:
If you have some time, take a look at this issue: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1208

PR #996, opened 2020-03-02
Concat value of complex primary key for correct calculate total pages in Datagrid

peter-gribanov:
## Subject

Closes #497

## Changelog

```markdown
### Fixed
- Concat value of complex primary key for correct calculate total pages in Datagrid
```

PR #1217, opened 2020-11-25
Update Pager.php

Warxcell:
## Subject

I have entity with composite key, just created admin for it and I found this bug.

I'm using: sonata-project/doctrine-orm-admin-bundle:3.24.0, doctrine/orm:2.7.4

I am targeting this branch, because it's bug fix.

```markdown
### Changed
- Use Doctrine ORM Paginator to count in Pager.

### Fixed
- Support of composite key for computeNbResult
```

# Notes

I think it can be used as single method of generating query, no matter if it's composite or not.

### DIFF
--- a/src/Datagrid/Pager.php
+++ b/src/Datagrid/Pager.php
@@ -203,7 +203,7 @@ class Pager extends BasePager
         }
 
         if (!$this->getQuery() instanceof ProxyQueryInterface) {
-            throw new \InvalidArgumentException(
+            throw new \TypeError(
                 sprintf('The datagrid query MUST implement %s.', ProxyQueryInterface::class)
             );
         }
--- a/src/Datagrid/Pager.php
+++ b/src/Datagrid/Pager.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Sonata\DoctrineORMAdminBundle\Datagrid;
 
 use Doctrine\ORM\Query;
+use Doctrine\ORM\Tools\Pagination\Paginator;
 use Sonata\AdminBundle\Datagrid\Pager as BasePager;
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
 
@@ -28,6 +29,10 @@ class Pager extends BasePager
 {
     /**
      * Use separator in CONCAT() function for correct determinate similar records.
+     *
+     * NEXT_MAJOR: remove this property.
+     *
+     * @deprecated since sonata-project/doctrine-orm-admin-bundle 3.x and will be removed in 4.0
      */
     public const CONCAT_SEPARATOR = '|';
 
@@ -51,17 +56,20 @@ class Pager extends BasePager
     /**
      * NEXT_MAJOR: remove this method.
      *
-     * @deprecated since sonata-project/doctrine-orm-admin-bundle 3.27
-     *
      * @return int
+     *
+     * @deprecated since sonata-project/doctrine-orm-admin-bundle 3.27
      */
     public function computeNbResult()
     {
         if ('sonata_deprecation_mute' !== (\func_get_args()[0] ?? null)) {
-            @trigger_error(sprintf(
-                'The %s() method is deprecated since sonata-project/doctrine-orm-admin-bundle 3.27 and will be removed in 4.0.',
-                __METHOD__,
-            ), E_USER_DEPRECATED);
+            @trigger_error(
+                sprintf(
+                    'The %s() method is deprecated since sonata-project/doctrine-orm-admin-bundle 3.27 and will be removed in 4.0.',
+                    __METHOD__,
+                ),
+                E_USER_DEPRECATED
+            );
         }
 
         return $this->computeResultsCount();
@@ -86,10 +94,13 @@ class Pager extends BasePager
             return $this->resultsCount;
         }
 
-        @trigger_error(sprintf(
-            'Relying on the protected property "%s::$nbResults" and its getter/setter is deprecated since sonata-project/doctrine-orm-admin-bundle 3.27 and will fail 4.0. Use "countResults()" and "setResultsCount()" instead.',
-            self::class,
-        ), E_USER_DEPRECATED);
+        @trigger_error(
+            sprintf(
+                'Relying on the protected property "%s::$nbResults" and its getter/setter is deprecated since sonata-project/doctrine-orm-admin-bundle 3.27 and will fail 4.0. Use "countResults()" and "setResultsCount()" instead.',
+                self::class,
+            ),
+            E_USER_DEPRECATED
+        );
 
         return $deprecatedCount;
     }
@@ -97,17 +108,20 @@ class Pager extends BasePager
     /**
      * NEXT_MAJOR: remove this method.
      *
-     * @deprecated since sonata-project/doctrine-orm-admin-bundle 3.27
-     *
      * @return int
+     *
+     * @deprecated since sonata-project/doctrine-orm-admin-bundle 3.27
      */
     public function getNbResults()
     {
         if ('sonata_deprecation_mute' !== (\func_get_args()[0] ?? null)) {
-            @trigger_error(sprintf(
-                'The %s() method is deprecated since sonata-project/doctrine-orm-admin-bundle 3.27 and will be removed in 4.0. Use "countResults()" instead.',
-                __METHOD__,
-            ), E_USER_DEPRECATED);
+            @trigger_error(
+                sprintf(
+                    'The %s() method is deprecated since sonata-project/doctrine-orm-admin-bundle 3.27 and will be removed in 4.0. Use "countResults()" instead.',
+                    __METHOD__,
+                ),
+                E_USER_DEPRECATED
+            );
         }
 
         return $this->nbResults;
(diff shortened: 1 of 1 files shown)


VincentLanglet (changes requested):
- The Changelog should be a generic messaging talking to every user, not the error message you have in your specific project.
- The tests are failing
- Maybe @peter-gribanov can help https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/996#issuecomment-594409886

peter-gribanov:
I'm not sure what the `IDENTITY()` function works with composite PKs. If it supports composite keys, then you don't need to edit method `countCompositePrimaryKey()`, but delete method `countCompositePrimaryKey()` and `countSinglePrimaryKey()` and rewrite the place of its call. That is, it is necessary to revert changes in my PR #996 with minor modifications.
And methods `getCountColumn()` and `setCountColumn()` cease to make sense.

Warxcell:
ok, it looks like if composite key is of scalars, concat needs to be used, otherwise IDENTITY. Will do some research and further modifications.

Warxcell:
What about using integrated ORM Paginator?

```
public function computeNbResult()
{

        $paginator = new Paginator($this->getQuery()->getQueryBuilder());

        return count($paginator);
}
```

src/Datagrid/Pager.php
```diff
@@ -26,11 +26,6 @@
  */
 class Pager extends BasePager
 {
-    /**
-     * Use separator in CONCAT() function for correct determinate similar records.
-     */
-    public const CONCAT_SEPARATOR = '|';
```
VincentLanglet, line 32: This is a BC-break.

You have to keep it, and update the phpdoc
```
/*
 * NEXT_MAJOR: Remove this constant.
 *
 * @deprecated since sonata-project/doctrine-orm-admin-bundle 3.23 and will be removed in version 4.0.
 */
```

VincentLanglet (changes requested):

src/Datagrid/Pager.php
Warxcell: Isn't it for internal use only?

VincentLanglet: It's a public property and there is no @internal ; so you cant be sure someone didnt used it

src/Datagrid/Pager.php
```diff
@@ -26,6 +26,15 @@
  */
 class Pager extends BasePager
 {
+    /**
+     * Use separator in CONCAT() function for correct determinate similar records.
+     *
+     * NEXT_MAJOR: remove this property.
+     *
+     * @deprecated since sonata-project/doctrine-orm-admin-bundle 2.4 and will be removed in 4.0
```
VincentLanglet, line 34:
```suggestion
     * @deprecated since sonata-project/doctrine-orm-admin-bundle 3.x and will be removed in 4.0
```

VincentLanglet (commented):
I removed the same test than you in https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1200/files

If you rebase 3.x, it should be removed from the diff.

Warxcell:
> I removed the same test than you in https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1200/files
> 
> If you rebase 3.x, it should be removed from the diff.
> 
> And do you think you'll be able to add a test of the `computeNbResult` method ?

Since it's using 3rd party class, I don't think there is much to test. I see that test exists for composite key.

src/Datagrid/Pager.php
```diff
@@ -43,22 +47,12 @@ class Pager extends BasePager
...
-            $countQuery->setParameters($this->getParameters());
-        }
-
-        if (\count($this->getCountColumn()) > 1) {
-            $this->countCompositePrimaryKey($countQuery);
-        } else {
-            $this->countSinglePrimaryKey($countQuery);
+        if (!$this->getQuery() instanceof ProxyQuery) {
+            throw new \LogicException('Only '.ProxyQuery::class.' is supported');
         }
```
VincentLanglet, line 52: @phansys Here we have the same problem we discussed https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1199#pullrequestreview-529410825

The change proposed by @Warxcell (using a paginator) fix the issue for the ProxyQuery, but using `getQueryBuilder` could be considered as a BC-break since it's doesn't exist in the `ProxyQueryInterface` class.

VincentLanglet, line 52: There is now a `Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface.

```suggestion
        if (!$this->getQuery() instanceof ProxyQueryInterface) {
            throw new \InvalidValueException(sprintf('The datagrid query MUST implements %s', .ProxyQueryInterface::class.'));
        }
```
@phansys may have a better exception type and exception message.

VincentLanglet:
@Warxcell The failing build (with the naming issue) should be fixed with the PR: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1199

You just need to wait for the merge

VincentLanglet:
Hi @Warxcell I think you have to rebase 3.x to fix the build.

franmomu:
I've never used composite keys with Sonata, maybe it would be interesting to create a functional test. Since https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1242, it should be easy to create an [Entity](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/tree/9f9f21b630a28b83b467cf2b62dd4ae34e18e7a6/tests/App/Entity) with composite key with its [Admin](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/tree/9f9f21b630a28b83b467cf2b62dd4ae34e18e7a6/tests/App/Admin) and check [some CRUD operations or whatever is needed to test](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/tree/9f9f21b630a28b83b467cf2b62dd4ae34e18e7a6/tests/Functional).

OskarStark:
I like your idea 💡 @franmomu

src/Datagrid/Pager.php
```diff
@@ -43,22 +47,14 @@ class Pager extends BasePager
...
-            $countQuery->setParameters($this->getParameters());
-        }
-
-        if (\count($this->getCountColumn()) > 1) {
-            $this->countCompositePrimaryKey($countQuery);
-        } else {
-            $this->countSinglePrimaryKey($countQuery);
+        if (!$this->getQuery() instanceof ProxyQueryInterface) {
+            throw new \InvalidArgumentException(
+                sprintf('The datagrid query MUST implements %s', ProxyQueryInterface::class)
```
phansys, line 52:
```suggestion
                sprintf('The datagrid query MUST implement %s.', ProxyQueryInterface::class)
```

phansys (changes requested):

src/Datagrid/Pager.php
```diff
@@ -43,22 +47,14 @@ class Pager extends BasePager
...
-        if (\count($this->getParameters()) > 0) {
-            $countQuery->setParameters($this->getParameters());
-        }
-
-        if (\count($this->getCountColumn()) > 1) {
-            $this->countCompositePrimaryKey($countQuery);
-        } else {
-            $this->countSinglePrimaryKey($countQuery);
+        if (!$this->getQuery() instanceof ProxyQueryInterface) {
+            throw new \InvalidArgumentException(
```
phansys, line 51: Since the checked value is not an argument of this method, I think we should use `\TypeError` instead,

VincentLanglet (approved):

src/Datagrid/Pager.php
```diff
@@ -143,64 +157,36 @@ public function init()
...
-    }
-
     private function computeResultsCount(): int
     {
-        $countQuery = clone $this->getQuery();
-
-        if (\count($this->getParameters()) > 0) {
-            $countQuery->setParameters($this->getParameters());
+        if (!$this->getQuery() instanceof ProxyQueryInterface) {
+            throw new \InvalidArgumentException(
```
phansys, line 183:
```suggestion
            throw new \TypeError(
```

src/Datagrid/Pager.php
```diff
@@ -143,64 +157,36 @@ public function init()
...
-        ));
-    }
-
     private function computeResultsCount(): int
     {
-        $countQuery = clone $this->getQuery();
-
-        if (\count($this->getParameters()) > 0) {
-            $countQuery->setParameters($this->getParameters());
+        if (!$this->getQuery() instanceof ProxyQueryInterface) {
```
VincentLanglet, line 182: What's the purpose of this check ?

`getQuery` is returning `ProxyQueryInterface|null`.
It seems better to write `null !== $this->getQuery()`.
And then, the message should explain that the query wasn't set.

If the goal is to check that the method `getQueryBuilder` exists, 
you should not use `Sonata\AdminBundle\Datagrid\ProxyQueryInterface` but `Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface`

VincentLanglet (changes requested):

src/Datagrid/Pager.php
VincentLanglet: > But then, maybe we should directly add these checks in `setQuery/getQuery`.

Maybe Pager should be generic 
```
/**
 * @template Query of ProxyQueryInterface
 */
```

VincentLanglet: I created https://github.com/sonata-project/SonataAdminBundle/pull/6759

VincentLanglet:
> I've never used composite keys with Sonata, maybe it would be interesting to create a functional test. Since #1242, it should be easy to create an [Entity](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/tree/9f9f21b630a28b83b467cf2b62dd4ae34e18e7a6/tests/App/Entity) with composite key with its [Admin](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/tree/9f9f21b630a28b83b467cf2b62dd4ae34e18e7a6/tests/App/Admin) and check [some CRUD operations or whatever is needed to test](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/tree/9f9f21b630a28b83b467cf2b62dd4ae34e18e7a6/tests/Functional).

I added these functional tests here: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1268
and finish this PR to see if everything works fine with this new way to count composite primary keys.

VincentLanglet:
Closed by https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1268 ; Thanks @Warxcell

Issue #1188, opened 2020-11-05
Add param and return type hint


VincentLanglet:
### ToDo

- [ ] Admin
- [ ] Block
- [ ] Builder
- [ ] Datagrid
- [ ] DependencyInjection
- [ ] Filter
- [ ] Guesser
- [ ] Model
- [ ] Util

PR #1218, opened 2020-11-27
Introduce DoctrineORMProxyQueryInterface

### LINKED ISSUE #1216: This project is using methods which does not exist in ProxyQueryInterface without exposing a custom interface
Related to the discussion: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1199#issuecomment-732333757

When methods have `ProxyQueryInterface` param, we're using them like it was always a `ProxyQuery`.
We should add a type-check before.

Currently the `ProxyQuery` is not final, so
```
$query instanceof ProxyQuery
```
Can do the job, but exposing an interface `DoctrineProxyQueryInterface extends ProxyQueryInterface` could be better in order to declare `ProxyQuery` as final and allow others implementation.

The main issue with the DoctrineProxyQueryInterface (which should basically be ProxyQueryInterface + QueryBuilder) are:
- There is no QueryBuilderInterface we can extends.
- If we declare our DoctrineProxyQueryInterface with all the methods from the QueryBuilder, we'll have to add them to our ProxyQuery since __call is not enough to implement an Interface.

VincentLanglet:
## Subject

Then I'll replace every magic `foo` call by `getQueryBuilder->foo()`

Closes #1216.

## Changelog

```markdown
### Added
- Added `Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface`
```

### DIFF
--- a/src/Datagrid/ProxyQueryInterface.php
+++ b/src/Datagrid/ProxyQueryInterface.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DoctrineORMAdminBundle\Datagrid;
+
+use Doctrine\ORM\QueryBuilder;
+use Sonata\AdminBundle\Datagrid\ProxyQueryInterface as BaseProxyQueryInterface;
+
+interface ProxyQueryInterface extends BaseProxyQueryInterface
+{
+    /**
+     * @return QueryBuilder
+     */
+    public function getQueryBuilder();
+}
--- a/src/Datagrid/ProxyQuery.php
+++ b/src/Datagrid/ProxyQuery.php
@@ -18,7 +18,6 @@ use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
-use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
 
 /**
  * This class try to unify the query usage with Doctrine.


src/Datagrid/DoctrineORMProxyQueryInterface.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Sonata\DoctrineORMAdminBundle\Datagrid;
+
+use Doctrine\ORM\QueryBuilder;
+use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
+
+interface DoctrineORMProxyQueryInterface extends ProxyQueryInterface
```
phansys, line 8:
```suggestion
use Sonata\AdminBundle\Datagrid\ProxyQueryInterface as BaseProxyQueryInterface;

interface ProxyQueryInterface extends BaseProxyQueryInterface
```

core23 (approved):

phansys (approved):

src/Datagrid/ProxyQueryInterface.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Sonata\DoctrineORMAdminBundle\Datagrid;
+
+use Doctrine\ORM\QueryBuilder;
+use Sonata\AdminBundle\Datagrid\ProxyQueryInterface as BaseProxyQueryInterface;
+
+interface ProxyQueryInterface extends BaseProxyQueryInterface
+{
+    /**
+     * @return QueryBuilder
```
phansys, line 22: Should we add a `NEXT_MAJOR` comment indicating that this docblock must be moved to the return type declaration?

core23: 👎 That should be clear. We need to do this for every method.

phansys: Good. It's just the panic for the lack of the interface.

VincentLanglet: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1188

PR #1236, opened 2020-12-18
Fix deprecation of EmptyFilter

pbories:
## Subject
EmptyFilter

### DIFF
--- a/src/Filter/EmptyFilter.php
+++ b/src/Filter/EmptyFilter.php
@@ -13,16 +13,19 @@ declare(strict_types=1);
 
 namespace Sonata\DoctrineORMAdminBundle\Filter;
 
-// NEXT_MAJOR: remove this file
-@trigger_error(sprintf(
-    'The %1$s\EmptyFilter class is deprecated since version 3.x and will be removed in 4.0.'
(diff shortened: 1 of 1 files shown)


VincentLanglet (dismissed):

src/Filter/EmptyFilter.php
```diff
@@ -13,16 +13,18 @@
...
 /**
  * @psalm-suppress InvalidExtendClass
  */
 final class EmptyFilter extends NullFilter
 {
+    public function __construct()
+    {
+        // NEXT_MAJOR: remove this file
+        @trigger_error(sprintf(
+            'The %1$s\EmptyFilter class is deprecated since version 3.x and will be removed in 4.0.'
```
core23, line 25: You should use ``__CLASS__`` instead of hardcoded string

VincentLanglet: sprintf('The %s class is deprecated since version 3.x and will be removed in 4.0. Use %s instead.', __CLASS__, NullFilter::class)

core23 (approved):

VincentLanglet (approved):

PR #1241, opened 2020-12-22
Set the correct fieldName to get the correct value

### LINKED ISSUE #1219: FieldDescription::getValue does not allow to chain custom getter
## Feature Request

Hi,
I need to chain custom getters to add a field in a listMapper.

For instance I have 2 classes `Man` and `Dog` with the following methods :
```
Class Man {
    getDogs();

    // custom getter
    getFavoriteDog();
}
```
```
class Dog {
    getName();

    // custom getter
    getFormattedName();
}
```

`add('favoriteDog')` will work on ManAdmin and `add('formattedName')` will work on DogAdmin.
But `add('favoriteDog.formattedName')` will not display anything on ManAdmin.

From what I saw, the problem comes from FieldDescription::getValue which search only the first getter.

Does anyone have an idea to fix it without creating a new custom getter in my class `Man` ?

VincentLanglet:
## Subject

I am targeting this branch, because the wrong field name is a bug, so this PR would be a bug fix.

When using the following code 
```
->add('foo.bar')
```

If the subject has a relation to a Foo entity, which has a relation to a Bar entity, the fieldDescription has the following data:
- fieldName: bar
- parentAssociationMapping: An array containing foo.

If it's an embedded entity, the fieldDescription had the following data:
- fieldName: bar
- no parentAssociationMapping.

That's why the there was extra code for embedded fields.
- in getValue
- in fixFieldDescription

But both of these code could be avoid if we correctly set the fieldDescription fieldName when looking for parentAssociationMapping. I propose to change this to:
- fieldName: foo.bar
- no parentAssociationMapping.

Then, no code is needed in fixFieldDescription. And in getValue, we just have to iterate on `explode('.', $fieldName)`.

Bonus point: it works with custom getters too.

Closes #1219.

## Changelog

```markdown
### Changed
- When using embedded fields or fake field 'foo.bar' (with custom getters in the entity), `FieldDescription::fieldName` is changed from `bar` to the correct value `foo.bar`.

### Fixed
- Support for embedded and custom getters by the FieldDescription.
```

### DIFF
--- a/src/Admin/FieldDescription.php
+++ b/src/Admin/FieldDescription.php
@@ -105,16 +105,6 @@ class FieldDescription extends BaseFieldDescription
             $object = $this->getFieldValue($object, $parentAssociationMapping['fieldName']);
         }
 
-        $fieldMapping = $this->getFieldMapping();
-        // Support embedded object for mapping
-        // Ref: https://www.doctrine-project.org/projects/doctrine-orm/en/latest/tutorials/embeddables.html
-        if (isset($fieldMapping['declaredField'])) {
-            $parentFields = explode('.', $fieldMapping['declaredField']);
-            foreach ($parentFields as $parentField) {
-                $object = $this->getFieldValue($object, $parentField);
-            }
-        }
-
-        return $this->getFieldValue($object, $this->fieldName);
+        return $this->getFieldValue($object, $this->getFieldName());
     }
 }
--- a/src/Builder/DatagridBuilder.php
+++ b/src/Builder/DatagridBuilder.php
@@ -92,6 +92,7 @@ class DatagridBuilder implements DatagridBuilderInterface
                     $fieldDescription->setOption('global_search', $fieldDescription->getOption('global_search', true)); // always search on string field only
                 }
 
+                // NEXT_MAJOR: Remove this, the fieldName should be correctly set at the creation.
                 if (!empty($embeddedClasses = $metadata->embeddedClasses)
                     && isset($fieldMapping['declaredField'])
                     && \array_key_exists($fieldMapping['declaredField'], $embeddedClasses)
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "doctrine/doctrine-bundle": "^1.8 || ^2.0",
         "doctrine/orm": "^2.5",
         "doctrine/persistence": "^1.3.4 || ^2.0",
-        "sonata-project/admin-bundle": "^3.83",
+        "sonata-project/admin-bundle": "^3.84",
         "sonata-project/exporter": "^1.11.0 || ^2.0",
         "sonata-project/form-extensions": "^0.1 || ^1.4",
         "symfony/config": "^4.4",
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -198,7 +198,8 @@ class ModelManager implements ModelManagerInterface, LockInterface
             $options,
             $metadata->fieldMappings[$propertyName] ?? [],
             $metadata->associationMappings[$propertyName] ?? [],
-            $parentAssociationMappings
+            $parentAssociationMappings,
+            $propertyName
         );
     }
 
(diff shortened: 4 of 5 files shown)


wbloszyk (dismissed):

src/Admin/FieldDescription.php
```diff
@@ -93,19 +94,53 @@ public function isIdentifier()
     public function getValue($object)
     {
         foreach ($this->parentAssociationMappings as $parentAssociationMapping) {
-            $object = $this->getFieldValue($object, $parentAssociationMapping['fieldName']);
+            $object = $this->getChildValue($object, $parentAssociationMapping['fieldName']);
         }
 
+        /*
+         * NEXT_MAJOR: Remove this code and the related tests.
+         * The support for embedded object come from the next foreach if the fieldName is correctly set.
```
greg0ire, line 102:
```suggestion
         * The support for embedded objects comes from the next foreach if the fieldName is correctly set.
```

src/Admin/FieldDescription.php
```diff
@@ -93,19 +94,53 @@ public function isIdentifier()
...
+    private function getChildValue(?object $object, string $fieldName): ?object
+    {
+        if (null === $object) {
+            return null;
+        }
+
+        $child = $this->getFieldValue($object, $fieldName);
+        if (null !== $child && !\is_object($child)) {
+            throw new NoValueException(sprintf(
+                'Unexpected value when accessing to the property "%s" on the class "%s" for the field "%s".'
```
greg0ire, line 135:
```suggestion
                'Unexpected value when accessing to property "%s" on class "%s" for field "%s".'
```

src/Admin/FieldDescription.php
```diff
@@ -93,19 +94,53 @@ public function isIdentifier()
...
+                .' Expected object|null, got %s.',
```
greg0ire, line 136: If you need to split the string in the code, it's probably too long in the output exception message. How about using linkebreaks in the code AND in the output?

```php
<<<'EXCEPTION'
Unexpected value when accessing to the property "%s" on the class "%s" for the field "%s".
Expected object|null, got %s.
EXCEPTION
```

VincentLanglet:
@greg0ire I moved the support for fieldName with `.` in SonataAdmin instead: https://github.com/sonata-project/SonataAdminBundle/pull/6724

src/Admin/FieldDescription.php
```diff
@@ -105,16 +105,6 @@ public function getValue($object)
...
 
-        $fieldMapping = $this->getFieldMapping();
-        // Support embedded object for mapping
-        // Ref: https://www.doctrine-project.org/projects/doctrine-orm/en/latest/tutorials/embeddables.html
-        if (isset($fieldMapping['declaredField'])) {
-            $parentFields = explode('.', $fieldMapping['declaredField']);
-            foreach ($parentFields as $parentField) {
-                $object = $this->getFieldValue($object, $parentField);
-            }
-        }
```
VincentLanglet, line 116: This is not needed anymore (and will be buggy instead) if the correct fieldName is set in `ModelManager::getNewFieldDescriptionInstance`.

VincentLanglet:
The PR seems ready to me :)

src/Builder/DatagridBuilder.php
```diff
@@ -92,6 +92,7 @@ public function fixFieldDescription(AdminInterface $admin, FieldDescriptionInter
                     $fieldDescription->setOption('global_search', $fieldDescription->getOption('global_search', true)); // always search on string field only
                 }
 
+                // NEXT_MAJOR: Remove this, the fieldName should be correctly be set at the creation.
```
OskarStark, line 95:
```suggestion
                // NEXT_MAJOR: Remove this, the fieldName should be correctly set at the creation.
```

OskarStark (approved):

src/Builder/DatagridBuilder.php
VincentLanglet: Fixed

OskarStark (approved):

PR #1242, opened 2020-12-26
Add functional tests

franmomu:
This adds some functional tests using symfony/panther to test CRUD operations as well as some embedded and many to one mapping. 

I had this half way done.

## To do

- [x] Create a PR in dev-kit to ignore `phpunit.xml.dist` and `bootstrap.php` and custom gitignore path: `/tests/App/public/bundles`.


VincentLanglet:
> ## To do
> * [ ]  Create a PR in dev-kit to ignore `phpunit.xml.dist` and `bootstrap.php` and custom gitignore path: `/tests/App/public/bundles`.

Can't we avoid this with an option (and maybe a `make pretest` command) ?
```
panther: true
```

VincentLanglet (approved):

OskarStark:
Thanks!

PR #1243, opened 2021-01-02
Use ModelManagerException in ModelManager

VincentLanglet:
## Subject

In SonataAdmin, some methods have phpdoc about exception thrown. Some don't.
I wanted to add `@throws` for every methods throwing an exception in the interface, so I looked at the persistence bundle when exception are thrown and which kind.

Seems there is a `ModelManagerException`, I assume that most of the exception thrown should be this one.

This PR is more a way to open a discussion about multiple subject:
- `$this->getEntityManager($class)` or `$this->getMetadata($class)` are throwing an exception when the class has no entity manager / metadata. Currently `getEntityManager` throw a `RuntimeException` which is not documented in the interface. I think we should instead one of the two solutions
  - throw an InvalidArgument because it means the call was made which a wrong class and LogicException "should lead directly to a fix in your code." according to https://www.php.net/manual/en/class.logicexception.php. (I personally prefer this choice)
  - throw a ModelManagerException because it's more specific. (But it needs to add a `@throws` tag for every method calling these methods)
- `create`, `update`, ... methods have catch `\PDOException` and `DBALException` but not for others exception (And I didn't find a method throwing a `DBALException`). I think we should catch every exception instead to always throw a `ModelManagerException`.
- `lock` is doing nothing if the metadata is not versioned. I think it should throw an exception instead (Which one ?) because it shouldn't do nothing silently.
- `getIdentifierValues` could throw an exception because of the call `$connection->getDatabasePlatform()`, I think we can catch it instead like I did in this PR.
- `getModelInstance` could throw a `\ReflectionException` so I added a TryCatch, but I think we should deprecate this method and move it to SonataAdmin instead, since there is nothing database-related ; WDYT ?

I will update this PR (and create one adding the `@throws` tag to the Interface in SonataAdmin, after your opinion @sonata-project/contributors :) 

## Changelog

```markdown
### Added
- Added some `Class::newMethod()` to do great stuff.

### Changed

### Deprecated

### Removed

### Fixed

### Security
```

### DIFF
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -14,12 +14,10 @@ declare(strict_types=1);
 namespace Sonata\DoctrineORMAdminBundle\Model;
 
 use Doctrine\Common\Util\ClassUtils;
-use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\LockMode;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\EntityManager;
-use Doctrine\ORM\OptimisticLockException;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
 use Doctrine\ORM\UnitOfWork;
@@ -90,6 +88,8 @@ class ModelManager implements ModelManagerInterface, LockInterface
      *
      * @param string $class
      *
+     * @throws ModelManagerException if the entity manager or the metadata is not found
+     *
      * @return ClassMetadata
      *
      * @phpstan-param class-string $class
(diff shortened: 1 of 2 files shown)


src/Model/ModelManager.php
```diff
@@ -250,25 +250,23 @@ public function delete($object)
...
 
     public function getLockVersion($object)
     {
-        // NEXT_MAJOR: Remove `sonata_deprecation_mute`.
-        $metadata = $this->getMetadata(ClassUtils::getClass($object), 'sonata_deprecation_mute');
+        try {
+            // NEXT_MAJOR: Remove `sonata_deprecation_mute`.
+            $metadata = $this->getMetadata(ClassUtils::getClass($object), 'sonata_deprecation_mute');
+        } catch (\Exception $exception) {
+            return null;
```
core23, line 268: Catching an exception without logging is dangerous!

core23 (changes requested):

src/Model/ModelManager.php
VincentLanglet: Please take a look at the explanation https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1243#issue-547793912
I would be happy to get your opinion about the different point.

For this case, theoretically if there is no entityManager/metadata, the object has no lockVersion.
To me, other solution would be,
- if `getMetadata` throw a `ModelManagerException`, I catch it, then I throw instead a `LockException` seems this method is in the LockInterface.
- if `getMetadata` throw only `\InvalidArgumentException`, I don't catch it.
So we should first decide about the listed points.

PR #1248, opened 2021-01-03
Avoid Pager deprecation

VincentLanglet:
## Subject

We may not need to add the countColumn property/methods with the following PR
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1217

## Changelog

```markdown
### Fixed
- Unavoidable deprecation in Pager.
```

### DIFF
--- a/src/Datagrid/Pager.php
+++ b/src/Datagrid/Pager.php
@@ -39,6 +39,11 @@ class Pager extends BasePager
     protected $queryBuilder = null;
 
     /**
+     * @var string[]
+     */
+    protected $countColumn = ['id'];
+
+    /**
      * @var int
      */
     private $resultsCount = 0;
@@ -89,13 +94,6 @@ class Pager extends BasePager
         return $deprecatedCount;
     }
 
-    public function setResultsCount(int $count): void
-    {
-        $this->resultsCount = $count;
-        // NEXT_MAJOR: Remove this line.
-        $this->setNbResults($count, 'sonata_deprecation_mute');
-    }
-
     /**
      * NEXT_MAJOR: remove this method.
      *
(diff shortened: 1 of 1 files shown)


OskarStark (dismissed):

franmomu:
There are some cs and phpstan errors

VincentLanglet:
Yes, I didn't fix them because I was thinking about trying to merge https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1217 before @franmomu ; what do you think about this PR ?

src/Datagrid/Pager.php
```diff
@@ -89,13 +94,6 @@ public function countResults(): int
         return $deprecatedCount;
     }
 
-    public function setResultsCount(int $count): void
```
VincentLanglet, line 92: Changing this from public to private is BC since it was not released.

src/Datagrid/Pager.php
```diff
@@ -117,15 +115,22 @@ public function getNbResults()
...
         $this->setResultsCount($this->computeNbResult('sonata_deprecation_mute'));
 
         $this->getQuery()->setFirstResult(null);
         $this->getQuery()->setMaxResults(null);
 
-        if (\count($this->getParameters()) > 0) {
-            $this->getQuery()->setParameters($this->getParameters());
+        // NEXT_MAJOR: Remove this code.
+        if (\count($this->getParameters('sonata_deprecation_mute')) > 0) {
+            @trigger_error(
```
franmomu, line 128: I guess this would need an upgrade note (BTW we missed the upgrade note for https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1257, I'll add it to MongoDB when creating a similar PR to this)

VincentLanglet: I don't think an upgrade note is needed here.

`setParameters` and all similar method are already deprecated by SonataAdmin, so you should not use them anymore.
Here we're just saying that we won't use them, to respect the SonataAdmin deprecation.

If someone use them, he would already have got the deprecation from SonataAdmin.

franmomu: So maybe triggering a new deprecation it is not needed. 

In general, if we trigger a new deprecation, the upgrade note should be updated.

src/Datagrid/Pager.php
```diff
@@ -188,8 +209,13 @@ private function computeResultsCount(): int
     {
         $countQuery = clone $this->getQuery();
 
-        if (\count($this->getParameters()) > 0) {
-            $countQuery->setParameters($this->getParameters());
+        if (\count($this->getParameters('sonata_deprecation_mute')) > 0) {
```
franmomu, line 212: A `NEXT_MAJOR` comment should be added

src/Datagrid/Pager.php
VincentLanglet: Done

src/Datagrid/Pager.php
VincentLanglet: Done

franmomu (approved):

phansys (approved):

PR #1254, opened 2021-01-05
Delete button is independent of the ability to delete an object

axzx:
## Subject

The user may not have access to delete the object in question, but should be allowed to remove the selection from the list.
The Delete button does not delete the object, so we don't have to check if the user has the ability to delete the object.

## Changelog

```markdown
### Fixed
- Fix for checking if delete button is to be displayed
```

### DIFF
--- a/src/Resources/views/Form/form_admin_fields.html.twig
+++ b/src/Resources/views/Form/form_admin_fields.html.twig
@@ -113,7 +113,7 @@ file that was distributed with this source code.
             </span>
 
             <span class="btn-group">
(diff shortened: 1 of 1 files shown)


axzx:
> You're changing a lot of things in this template, it's not easy to review.
> Can you do separate commit ?

done
I will do new PR with code formatting only

VincentLanglet (approved):

VincentLanglet:
Thanks @axzx.

Can you update the changelog of your first message.
To me, it should be in the `Fixed` category.

Something like 
```
## Fixed
- Correctly display the remove-selection button from ...
```
But I don't know when this code is used.

VincentLanglet (approved):

OskarStark (approved):

PR #1257, opened 2021-01-06
handle SonataAdminBundle Pager::nbresults deprecation

dmaicher:
## Subject

I am targeting this branch, because its BC.

## Changelog

```markdown
### Added
- implemented `Sonata\AdminBundle\Datagrid\PagerInterface::countResults()`

### Deprecated
- `Sonata\DoctrineORMAdminBundle\Datagrid\Pager::computeNbResult()`
- `Sonata\DoctrineORMAdminBundle\Datagrid\Pager::getNbResults()`
- `Sonata\DoctrineORMAdminBundle\Datagrid\Pager::setNbResults()`
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "doctrine/doctrine-bundle": "^1.8 || ^2.0",
         "doctrine/orm": "^2.5",
         "doctrine/persistence": "^1.3.4 || ^2.0",
-        "sonata-project/admin-bundle": "^3.84",
+        "sonata-project/admin-bundle": "^3.86",
         "sonata-project/exporter": "^1.11.0 || ^2.0",
         "sonata-project/form-extensions": "^0.1 || ^1.4",
         "symfony/config": "^4.4",
--- a/src/Datagrid/Pager.php
+++ b/src/Datagrid/Pager.php
@@ -39,26 +39,27 @@ class Pager extends BasePager
     protected $queryBuilder = null;
 
     /**
+     * @var int
+     */
+    private $resultsCount = 0;
+
+    /**
+     * NEXT_MAJOR: remove this method.
+     *
+     * @deprecated since sonata-project/doctrine-orm-admin-bundle 3.27
+     *
      * @return int
      */
     public function computeNbResult()
     {
-        $countQuery = clone $this->getQuery();
-
-        if (\count($this->getParameters()) > 0) {
-            $countQuery->setParameters($this->getParameters());
+        if ('sonata_deprecation_mute' !== (\func_get_args()[0] ?? null)) {
+            @trigger_error(sprintf(
+                'The %s() method is deprecated since sonata-project/doctrine-orm-admin-bundle 3.27 and will be removed in 4.0.',
+                __METHOD__,
+            ), E_USER_DEPRECATED);
         }
 
-        if (\count($this->getCountColumn()) > 1) {
-            $this->countCompositePrimaryKey($countQuery);
-        } else {
-            $this->countSinglePrimaryKey($countQuery);
-        }
-
-        return array_sum(array_column(
-            $countQuery->resetDQLPart('orderBy')->getQuery()->getResult(Query::HYDRATE_SCALAR),
-            'cnt'
-        ));
+        return $this->computeResultsCount();
     }
 
     public function getResults($hydrationMode = Query::HYDRATE_OBJECT)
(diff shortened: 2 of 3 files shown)


dmaicher:
@VincentLanglet before I continue: The `Pager` here is marked final via annotation. Do we keep BC then for people who extended this class? What's the policy there?

https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Datagrid/Pager.php#L25

VincentLanglet:
Yes, I think we still keep BC for method with `@final`

Be careful, your implementation of `countResults` is wrong.
`getResults` is returning the results of the current page !

```php
private $resultsCount;

public function setResultsCount($count) {
    $this->resultsCount = $count;
    // NEXT_MAJOR: Remove this line.
    $this->setNbResults($count, 'sonata_deprecation_mute'); // You need to override this method or make a new PR to sonataAdmin to allow to avoid the deprecation.
}

public function countResults() {
    $deprecatedCount = $this->getNbResults('sonata_deprecation_mute'); // You need to override this method or make a new PR to sonataAdmin to allow to avoid the deprecation.

    if ($deprecatedCount !== $this->resultsCount) {
        // trigger deprecation

        return $deprecatedCount;
    }

    return $this->resultsCount;
}
```
Maybe we can do something like this ?

I think it would be good to deprecate `getResults` in favor of `getPageResults` in SonataAdmin too.
cc @sonata-project/contributors 

And if we're changing `nbResult`, we should rename `computeNbResult`.
But after rereading, the `computeNbResult` method should be private. So you should just
- Add a comment `NEXT_MAJOR: restrict visibility to private and rename to `computeResultsCount`
- Add a deprecation in this method.
- Making the call in `init` with `sonata_deprecation_mute`.

dmaicher:
> Be careful, your implementation of `countResults` is wrong.
> `getResults` is returning the results of the current page !

I don't follow. Currently I just call `getNbResults` (which will trigger a deprecation though).

Will look into it then with keeping BC.

VincentLanglet:
> > Be careful, your implementation of `countResults` is wrong.
> > `getResults` is returning the results of the current page !
> 
> I don't follow. Currently I just call `getNbResults` (which will trigger a deprecation though).
> 
> Will look into it then with keeping BC.

My bad, i'm tired, I read 
```
count($this->getResults())
```

But I just discovered that
```
count($this->getResults()) !== $this->countResults()
```

src/Datagrid/Pager.php
```diff
@@ -71,11 +68,70 @@ public function getQuery()
...
+    {
+        // NEXT_MAJOR: just return "$this->resultsCount" directly.
+        $deprecatedCount = $this->getNbResults('sonata_deprecation_mute');
+
+        if ($deprecatedCount === $this->resultsCount) {
+            return $this->resultsCount;
+        }
+
+        @trigger_error(sprintf(
+            'Manipulating the protected member "%s::$nbResults" without calling "setResultsCount" is deprecated since sonata-project/doctrine-orm-admin-bundle 3.27 and will fail 4.0.',
```
dmaicher, line 81: any better idea for a message here? not so happy with that one

composer.json
```diff
@@ -29,7 +29,7 @@
         "doctrine/doctrine-bundle": "^1.8 || ^2.0",
         "doctrine/orm": "^2.5",
         "doctrine/persistence": "^1.3.4 || ^2.0",
-        "sonata-project/admin-bundle": "^3.84",
+        "sonata-project/admin-bundle": "3.x-dev",
```
dmaicher, line 32: todo: revert

src/Datagrid/Pager.php
dmaicher: What about

```
'Relying on the protected member "%s::$nbResults" and its getter/setter is deprecated since sonata-project/doctrine-orm-admin-bundle 3.27 and will fail 4.0. Use "countResults()" and "setResultsCount()" instead.',

Comment (VincentLanglet) [reply src/Datagrid/Pager.php:81]: Seems better indeed

Comment (phansys) [reply src/Datagrid/Pager.php:81]: Just a little observation: we usually use "property" instead of "member".

File: src/Datagrid/Pager.php [hunk 3]
```diff
@@ -84,18 +122,36 @@ public function init()
...
+    /**
+     * NEXT_MAJOR: remove this method.
+     *
+     * @param int $nb
+     */
+    protected function setNbResults($nb)
+    {
+        if ('sonata_deprecation_mute' !== (\func_get_args()[1] ?? null)) {
+            @trigger_error(sprintf(
+                'The %s method is deprecated since sonata-project/doctrine-orm-admin-bundle 3.27 and will be removed in 4.0. Use "setResultsCount" instead.',
```
phansys, line 146:
```suggestion
                'The "%s()" method is deprecated since sonata-project/doctrine-orm-admin-bundle 3.27 and will be removed in 4.0. Use "setResultsCount()" instead.',
```

phansys (changes requested):

dmaicher:
Once SonataAdminBundle 3.86 was released we need to bump the minimum in composer.json here before we release 3.27

VincentLanglet (dismissed):

OskarStark:
Can you bump it in this PR?

VincentLanglet:
> Can you bump it in this PR?

Need https://github.com/sonata-project/SonataAdminBundle/pull/6756

VincentLanglet:
3.86 is released :)

VincentLanglet (approved):

OskarStark:
@phansys I merged this, because your suggestion was merged 👍🏻

phansys:
> @phansys I merged this, because your suggestion was merged 👍🏻

Thank you!

PR #1259, opened 2021-01-11
Code formatting in views

axzx:
## Subject

I am targeting this branch, because no bc-break.

## Changelog

```markdown
### Changed
Code formatting in views and change from `<a>` with empty href to button.
```

### DIFF
--- a/src/Resources/views/CRUD/edit_orm_many_to_one.html.twig
+++ b/src/Resources/views/CRUD/edit_orm_many_to_one.html.twig
@@ -35,11 +35,11 @@ file that was distributed with this source code.
                 {% endif %}
             </span>
             <span style="display: none" >
-                {{ form_widget(form) }}
+                {{- form_widget(form) -}}
             </span>
         {% else %}
             <span id="field_widget_{{ id }}" >
-                {{ form_widget(form) }}
+                {{- form_widget(form) -}}
             </span>
         {% endif %}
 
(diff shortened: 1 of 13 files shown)


src/Resources/views/CRUD/edit_orm_many_to_one.html.twig
```diff
@@ -73,14 +73,14 @@ file that was distributed with this source code.
             {% endif %}
 
             {% if sonata_admin.edit == 'list' and sonata_admin.field_description.associationadmin.hasRoute('delete') and sonata_admin.field_description.associationadmin.hasAccess('delete') and btn_delete %}
-                <a  href=""
+                <button
```
franmomu, line 76: hmmm it is unlikely to happen, but I think this could be considered BC break as someone may rely in this being an `a`.

core23: I'm not sure if our BC promise covers such html changes. Otherwise we can't change anything in our templates without BC breaks. Same for JS/CSS

I'm okay with the `button` change

VincentLanglet: I'm okay too, but we should have better explanations in the Changelog.

franmomu:
Off-topic: Maybe we could add https://github.com/friendsoftwig/twigcs to our GA (in another PR).

franmomu (approved):

core23 (approved):

OskarStark:
Thank you 😊

PR #1264, opened 2021-01-12
Merge 3.x

dmaicher:
## Subject

I am targeting this branch, because we need to merge 3.x into master.


composer.json
```diff
@@ -54,14 +54,16 @@
...
     },
     "require-dev": {
+        "doctrine/doctrine-fixtures-bundle": "^3.4",
         "phpstan/extension-installer": "^1.0",
         "phpstan/phpstan": "^0.12.52",
         "phpstan/phpstan-symfony": "^0.12.10",
         "psalm/plugin-symfony": "^2.0",
         "simplethings/entity-audit-bundle": "^0.9 || ^1.0",
-        "sonata-project/block-bundle": "^4.2",
+        "sonata-project/block-bundle": "^4.2",        "symfony/panther": "^0.8.0",
```
VincentLanglet, line 63: Something weird here. Missing new line ?

dmaicher:
@VincentLanglet @franmomu So I cannot debug those failing functional tests locally. I don't have any browser in my dockerized dev environment.

Wondering: why are those tests actually using a real browser? are they testing some JS stuff as well?

dmaicher:
Ok now we can see the 500 error page output.

```
Return value of Sonata\AdminBundle\Admin\BaseFieldDescription::getType() must be of the type string or null, int returned

  at /home/runner/work/SonataDoctrineORMAdminBundle/SonataDoctrineORMAdminBundle/vendor/sonata-project/admin-bundle/src/Admin/BaseFieldDescription.php:261
  at Sonata\AdminBundle\Admin\BaseFieldDescription-&gt;getType()
     (/home/runner/work/SonataDoctrineORMAdminBundle/SonataDoctrineORMAdminBundle/src/Builder/FormContractor.php:71)
  at Sonata\DoctrineORMAdminBundle\Builder\FormContractor-&gt;fixFieldDescription()
     (/home/runner/work/SonataDoctrineORMAdminBundle/SonataDoctrineORMAdminBundle/vendor/sonata-project/admin-bundle/src/Form/FormMapper.php:117)
  at Sonata\AdminBundle\Form\FormMapper-&gt;add()
     (Admin/BookAdmin.php:62)
```

any idea why this would be an int? :see_no_evil:

VincentLanglet:
Maybe because of https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/master/src/Admin/FieldDescription.php#L69
or https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/master/src/Admin/FieldDescription.php#L30

You could try to cast to string

src/Admin/FieldDescription.php
```diff
@@ -67,7 +67,7 @@ public function setFieldMapping(array $fieldMapping): void
         $this->fieldMapping = $fieldMapping;
 
         $this->type = $this->type ?: (string) $fieldMapping['type'];
-        $this->mappingType = $this->mappingType ?: $fieldMapping['type'];
+        $this->mappingType ?: $fieldMapping['type'];
```
VincentLanglet: Not sure it was the intended change

dmaicher:
@VincentLanglet ok casting fixed the 500 error. Now some selector is not there :confused: A bit hard to say whats happening there

franmomu:
It is set directly from `ClassMetadata`:

https://github.com/sonata-project/SonataAdminBundle/issues/6749

> IMO this is wrong since they could change any value and, in this case, association mapping in ORM is an int value and our getType will only admit string in 4.x.

And one way to solve it I think: https://github.com/sonata-project/SonataAdminBundle/issues/6761

Casting to string shouldn't be a solution.

VincentLanglet:
> It is set directly from `ClassMetadata`:
> 
> [sonata-project/SonataAdminBundle#6749](https://github.com/sonata-project/SonataAdminBundle/issues/6749)
> 
> > IMO this is wrong since they could change any value and, in this case, association mapping in ORM is an int value and our getType will only admit string in 4.x.
> 
> And one way to solve it I think: [sonata-project/SonataAdminBundle#6761](https://github.com/sonata-project/SonataAdminBundle/issues/6761)
> 
> Casting to string shouldn't be a solution.

It's more tricky.

We don't care about the type for association mapping ; the only check is
```
if (!$fieldDescription->getType()) {
            throw new \RuntimeException(sprintf(
                'Please define a type for field `%s` in `%s`',
                $fieldDescription->getName(),
                \get_class($admin)
            ));
        }
```

But then we're using the mappingType to attach the admin class.
For association mapping we don't rely on the type but we use the mapping type.
Casting to string avoid the TypeError exception, and change nothing since we're not using it.

Best could be to try to try to remove the `$this->type = $this->type ?: $associationMapping['type']` line.
But let's wait another PR

VincentLanglet:
Test are already failing on 3.x https://github.com/sonata-project/SonataDoctrineORMAdminBundle/runs/1689211188

VincentLanglet:
@dmaicher Can you change 
```
.sonata-ba-list-field-text[objectid="category_novel"]
```
to
```
.sonata-ba-list-field-string[objectid="category_novel"]
```

?

I think it's because of this: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Guesser/TypeGuesser.php#L107-L108

src/Admin/FieldDescription.php
```diff
@@ -27,7 +27,7 @@ public function setAssociationMapping(array $associationMapping): void
     {
         $this->associationMapping = $associationMapping;
 
-        $this->type = $this->type ?: $associationMapping['type'];
+        $this->type = $this->type ?: (string) $associationMapping['type'];
```
dmaicher, line 30: so we keep this cast to string for now?

src/Admin/FieldDescription.php
```diff
@@ -66,7 +66,7 @@ public function setFieldMapping(array $fieldMapping): void
     {
         $this->fieldMapping = $fieldMapping;
 
-        $this->type = $this->type ?: $fieldMapping['type'];
+        $this->type = $this->type ?: (string) $fieldMapping['type'];
```
dmaicher, line 69: and this one

src/Admin/FieldDescription.php
VincentLanglet: For now, I would say yes. It works like this.

VincentLanglet (approved):

VincentLanglet:
Thanks @dmaicher

PR #1265, opened 2021-01-13
Do not use null as default value

VincentLanglet:
## Subject

`getOptions` is using `array_key_exist` so
```
$this->getOptions('field_type', DateTimeFilter::class)
```
will return `null` instead of `DateTimeFiler::class` when the type is guessed by our guesser.

This issue was reported on slack by someone using the master branch and he tried the fix for me.

## Changelog

```markdown
### Fixed
- Do not provide a default `null` `field_type` option for Filter
```

### DIFF
--- a/src/Filter/NumberFilter.php
+++ b/src/Filter/NumberFilter.php
@@ -17,6 +17,7 @@ use Sonata\AdminBundle\Datagrid\ProxyQueryInterface as BaseProxyQueryInterface;
 use Sonata\AdminBundle\Form\Type\Filter\NumberType;
 use Sonata\AdminBundle\Form\Type\Operator\NumberOperatorType;
 use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface;
+use Symfony\Component\Form\Extension\Core\Type\NumberType as FormNumberType;
 
 /**
  * @final since sonata-project/doctrine-orm-admin-bundle 3.24
@@ -59,7 +60,9 @@ class NumberFilter extends Filter
 
     public function getDefaultOptions()
     {
-        return [];
+        return [
+            'field_type' => FormNumberType::class,
+        ];
     }
 
     public function getRenderSettings()
--- a/src/Guesser/FilterTypeGuesser.php
+++ b/src/Guesser/FilterTypeGuesser.php
@@ -24,11 +24,8 @@ use Sonata\DoctrineORMAdminBundle\Filter\NumberFilter;
 use Sonata\DoctrineORMAdminBundle\Filter\StringFilter;
 use Sonata\DoctrineORMAdminBundle\Filter\TimeFilter;
 use Sonata\DoctrineORMAdminBundle\Model\MissingPropertyMetadataException;
-use Sonata\Form\Type\BooleanType;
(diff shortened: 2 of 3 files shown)


OskarStark (approved):

franmomu:
Looks like the problem is that `field_type` is not defined in 

https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/4098faf8f5da236c0300ce06f9f8abcaff8db5a2/src/Guesser/FilterTypeGuesser.php#L90-L98

I guess the `$options['field_type`]` should be added in both cases.

VincentLanglet:
> Looks like the problem is that `field_type` is not defined in
> 
> https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/4098faf8f5da236c0300ce06f9f8abcaff8db5a2/src/Guesser/FilterTypeGuesser.php#L90-L98
> 
> I guess the `$options['field_type`]` should be added in both cases.

It should instead rely on the default value like here: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Filter/DateFilter.php#L39

There is no reason to set `null`, instead we should set nothing and let the filter doing his work.

franmomu:
So, should we remove the rest of `field_type` from there?

VincentLanglet:
> So, should we remove the rest of `field_type` from there?

That's a good point.

Other filters are setting the field_type because they were no default one.

For instance, the NumberFilter is not overriding the getFieldType method
```
    public function getFieldType()
    {
        return $this->getOption('field_type', TextType::class);
    }
```

I'll fix this.

VincentLanglet:
By the way, in SonataAdmin we're not using `getFieldType` and `getFieldOption` so this could be removed from the FilterInterface.

VincentLanglet:
@franmomu Tests are updated to check that `null` is not passed.
And I removed the useless `field_type`.

src/Filter/NumberFilter.php
```diff
@@ -17,6 +17,7 @@
 use Sonata\AdminBundle\Form\Type\Filter\NumberType;
 use Sonata\AdminBundle\Form\Type\Operator\NumberOperatorType;
 use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface;
+use Symfony\Component\Form\Extension\Core\Type\NumberType as FormNumberType;
```
franmomu, line 20:
```suggestion
use Symfony\Component\Form\Extension\Core\Type\NumberType as SymfonyNumberType;
```
We usually use `Symfony` prefix I think

VincentLanglet: I found this
https://github.com/sonata-project/SonataAdminBundle/blob/3.x/src/Form/Type/Operator/ContainsOperatorType.php#L17
https://github.com/sonata-project/SonataAdminBundle/blob/3.x/src/Form/Type/Operator/DateOperatorType.php#L17
and so on

franmomu: Looks like in the `Filter` part we use `Form` I guess to not be confusing. I prefer using `Symfony` to know where it comes from, but that's another PR.

franmomu (approved):

PR #1268, opened 2021-01-17
Add composite primary keys tests and use Paginator

VincentLanglet:
I added some functional tests for composite primary keys.

https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1268/checks?check_run_id=1717007313 show the failure if I removed the code for composite keys.

Then I added the PR https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1217, to check that the code is still working fine.

## Changelog

```markdown
### Changed
- Use Doctrine ORM Paginator to count in Pager.

### Fixed
- Support of composite key for computeNbResult
```

### DIFF
--- a/src/Builder/DatagridBuilder.php
+++ b/src/Builder/DatagridBuilder.php
@@ -201,8 +201,6 @@ class DatagridBuilder implements DatagridBuilderInterface
     {
         $pager = $this->getPager($admin->getPagerType());
 
-        $pager->setCountColumn($admin->getModelManager()->getIdentifierFieldNames($admin->getClass()));
-
         $defaultOptions = ['validation_groups' => false];
         if ($this->csrfTokenEnabled) {
             $defaultOptions['csrf_protection'] = false;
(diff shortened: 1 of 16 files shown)


franmomu:
If I had understood correctly https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1217#issuecomment-734131929, the problem is when the composite keys are relations, with scalar values should work fine, can you please add a test with relations?

VincentLanglet:
> If I had understood correctly [#1217 (comment)](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1217#issuecomment-734131929), the problem is when the composite keys are relations, with scalar values should work fine, can you please add a test with relations?

I need some help. The fixtures are not loading anymore because
```
[error] Error thrown while running command "doctrine:fixtures:load --no-interaction". Message: "Reference to "product_1" does not exist"
```

Maybe I'm tired but I don't see why the CommandItemFixtures is load before ProductFixtures.
I have added a `getDependencies` method...

VincentLanglet:
Done @franmomu :)

phansys (approved):
I think these tests are fine.
I just have a concern about using real trademarks in the fixtures. I think we should avoid them if possible.

VincentLanglet:
I removed the trademarkes @phansys

franmomu (approved):

phansys (approved):

PR #1278, opened 2021-01-20
remove deprecated code & cleanup Pager

dmaicher:
Shall we do cleanup like this already on master?


OskarStark:
Yes why not 🤔

Thoughts 💭 @sonata-project/contributors ?

VincentLanglet (approved):
Sure ! Resolving NEXT_MAJOR comment will take a lot of times so we can start now.
We have to start now or we will never release 4.0.

OskarStark (approved):

PR #1280, opened 2021-01-25
create CountFilter

rgrassian:
## Subject

## Changelog

```markdown
### Added
- Added `CountFilter`.
```

### DIFF
--- a/src/Filter/CountFilter.php
+++ b/src/Filter/CountFilter.php
@@ -1,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DoctrineORMAdminBundle\Filter;
+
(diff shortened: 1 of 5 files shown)


src/Filter/CountFilter.php
```diff
@@ -0,0 +1,80 @@
...
+        return [NumberType::class, [
+            'field_type' => $this->getFieldType(),
+            'field_options' => $this->getFieldOptions(),
+            'label' => $this->getLabel(),
+        ]];
+    }
+
+    private function getOperator(int $type): string
+    {
+        return self::CHOICES[$type] ?? self::CHOICES[NumberOperatorType::TYPE_EQUAL];
```
phansys, line 78: IMO, instead of silently fall to `NumberOperatorType::TYPE_EQUAL` here, we should throw a `\OutOfBoundsException` exception here if the given type isn't registered.

VincentLanglet: It's similar https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Filter/AbstractDateFilter.php#L209 or https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Filter/NumberFilter.php

Maybe it's better to merge it like this (to keep consistency) and then open a new PR in order to add a deprecation and throw an exception in next major. WDYT ?

phansys (approved):

VincentLanglet (approved):

PR #1293, opened 2021-02-01
Add Typehint

### LINKED ISSUE #1188: Add param and return type hint

VincentLanglet:
Closes #1188


src/Datagrid/ProxyQuery.php
```diff
@@ -114,22 +114,19 @@ class ProxyQuery implements ProxyQueryInterface
...
-    public function __construct($queryBuilder)
+    public function __construct(QueryBuilder $queryBuilder)
     {
         $this->queryBuilder = $queryBuilder;
         $this->uniqueParameterId = 0;
         $this->entityJoinAliases = [];
     }
 
-    public function __call($name, $args)
+    public function __call(string $name, array $args)
```
franmomu, line 124: should we deprecate and remove this method in admin-bundle?

franmomu (approved):

src/Datagrid/ProxyQuery.php
VincentLanglet: And promote `$proxyQuery->getQueryBuilder()->...` instead ?

I think we should stop to use it in the Sonata code. 
I don't know if we should force the developers to migrate all his code too or if we can let this helper.

I you think we should, I'm ok with the deprecation.

franmomu: I prefer being explicit, but maybe it's too much for devs, maybe it's better to keep it for now.

OskarStark (approved):

PR #1308, opened 2021-02-15
fix compatiblity with latest SonataAdminBundle changes

dmaicher:
See https://github.com/sonata-project/SonataAdminBundle/commit/271e4ed5100e80ef1365c5be639a96b94ab1f302#diff-c698cbb1f2afce655c27e32ebfc8d9f37af04b8164a1d8d609da71d2858c6697R39

### DIFF
--- a/tests/Admin/FieldDescriptionTest.php
+++ b/tests/Admin/FieldDescriptionTest.php
@@ -198,10 +198,10 @@ class FieldDescriptionTest extends TestCase
 
     public function testGetValue(): void
     {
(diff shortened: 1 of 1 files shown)


VincentLanglet:
The accessor is now suppose to be a callable or the property path.
In the test, `myMethod` seems like the getter.

I think we could use a better name in the test.
The `getFoo` and `accessor => 'foo'`

VincentLanglet (approved):

VincentLanglet:
Thanks @dmaicher

PR #1314, opened 2021-02-17
Remove useless options

VincentLanglet:
## Subject

We're setting `code`, `name` and `label`.
- As explained [here](https://github.com/sonata-project/SonataAdminBundle/pull/6837#discussion_r576771541), I think that removing `code` should be considered as a bugfix. 
- The `name` option is not used.
- The `label` option is useless ; it's already set in the SonataAdminBundle, so the line is doing nothing.

I try it on my project and saw no impact.

## Changelog

```markdown
### Fixed
- Unavoidable deprecation about the `code` option.
```


dmaicher (approved):
not sure of the impact. But since you tested it should be fine :+1:

OskarStark (approved):

OskarStark:
Thanks!

Issue #1225, opened 2020-12-07
Order by computedField


VincentLanglet:
#### Sonata packages

```
$ composer show --latest 'sonata-project/*'
sonata-project/admin-bundle              3.81.1 3.82.0 The missing Symfony Admin Generator
sonata-project/block-bundle              4.4.0  4.4.0  Symfony SonataBlockBundle
sonata-project/cache                     2.0.1  2.0.1  Cache library
sonata-project/doctrine-extensions       1.10.1 1.10.1 Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.26.0 3.26.0 Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/exporter                  2.4.1  2.4.1  Lightweight Exporter library
sonata-project/form-extensions           1.7.1  1.7.1  Symfony form extensions
sonata-project/twig-extensions           1.4.1  1.4.1  Sonata twig extensions
```

## Subject

I would like to order the list by a computed sql field
```php
    protected function configureQuery(ProxyQueryInterface $query): ProxyQueryInterface
    {
        $query = parent::configureQuery($query);

        $alias = $query->getRootAlias()[0];
        $query->leftJoin("$alias.children", 'child');
        $query->leftJoin("$alias.dogs", 'dog');
        $query->addSelect('GREATEST(MAX(dog.age), MAX(child.age)) as HIDDEN maxAge');
        $query->addOrderBy('maxAge', 'DESC');

        return $query;
    }
```

This query is working well when I tried the SQL by myself.
But, it doesn't work in the Datagrid.

The issue is introduced by the `getFixedQueryBuilder()` method.https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/71ea5138cfc443a42aa0b8c692b98bee11001743/src/Datagrid/ProxyQuery.php#L366

Firstly, my hidden field is lost because it overrides the select:
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/71ea5138cfc443a42aa0b8c692b98bee11001743/src/Datagrid/ProxyQuery.php#L393
If I add manually `$queryBuilderId->addSelect(...)` it works better.

Secondly, I don't have 32 results by page because of the `distinct`https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/71ea5138cfc443a42aa0b8c692b98bee11001743/src/Datagrid/ProxyQuery.php#L396
[...]

PR #1319, opened 2021-02-25
Group by is better than distinct

VincentLanglet:
## Subject

I have a Contract entity and a Client entity.
A client can have multiple contracts. A contract can be associated to multiple clients.

In the `configureQuery` of my `ContractAdmin`, I'm using a `leftJoin('contract.client')`.
I got a weird behavior:
- The total of result was 102.
- I got 4 pages of 32 results (which mean 128 results)
- Some results were missing (in database the query was returning 133 results, but there was no page 5 on the admin)

Distinct is not a good thing to use with pagination:
If the results are:
```
-----------
contract | client
-----------
1 | 1
-----------
2 | 1
-----------
2 | 2
-----------
1 | 2
-----------
3 | 1
-----------
```

With distinct on col 1 and a limit of 2
- Page 1 return contract 1 and 2
- Page 2 return contract 2 and 1
- There is no page 3

With groupby, it's applied BEFORE the pagination, so it became
```
-----------
contract | client
-----------
1 | 1
-----------
2 | 1
-----------
3 | 1
-----------
```
and the result are the one expected.

I tried on my project and it works.

Another thing: `setDistinct` and `isDistinct` are not in the ProxyQueryInterface.
I think we should deprecate them, and always adding the groupBy in the fixQueryBuilder method.

## Changelog

```markdown
### Added
- `Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface::getDoctrineQuery()`

### Deprecated
- class `Sonata\DoctrineORMAdminBundle\Datagrid\OrderByToSelectWalker`
- `Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery::getFixedQueryBuilder()`
- `Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery::getSingleScalarResult()`

### Fixed
- Do not display multiple times the same row in the admin list and the export list.
```

### DIFF
--- a/src/Datagrid/OrderByToSelectWalker.php
+++ b/src/Datagrid/OrderByToSelectWalker.php
@@ -21,6 +21,10 @@ use Doctrine\ORM\Query\AST\SelectStatement;
 use Doctrine\ORM\Query\TreeWalkerAdapter;
 
 /**
+ * NEXT_MAJOR: Remove this file.
+ *
+ * @deprecated since sonata-project/doctrine-orm-admin-bundle 3.x
+ *
  * Finds all PathExpressions in an AST's OrderByClause, and ensures that
  * the referenced fields are present in the SelectClause of the passed AST.
  *
@@ -30,6 +34,16 @@ use Doctrine\ORM\Query\TreeWalkerAdapter;
  */
 final class OrderByToSelectWalker extends TreeWalkerAdapter
 {
+    public function __construct($query, $parserResult, array $queryComponents)
+    {
+        parent::__construct($query, $parserResult, $queryComponents);
+
+        @trigger_error(sprintf(
+            'The %s class is deprecated since sonata-project/doctrine-orm-admin-bundle 3.x and will be removed in version 4.0.',
+            __CLASS__,
+        ), \E_USER_DEPRECATED);
+    }
+
     public function walkSelectStatement(SelectStatement $AST)
     {
         if (!$AST->orderByClause instanceof OrderByClause) {
--- a/src/Datagrid/ProxyQueryInterface.php
+++ b/src/Datagrid/ProxyQueryInterface.php
@@ -31,7 +31,7 @@ interface ProxyQueryInterface extends BaseProxyQueryInterface
 
 //    /**
 //     * This method should be preferred over `$this->getQueryBuilder()->getQuery()`
-//     * since some changed are done to the query builder in order to handle all the
+//     * since some changes are done to the query builder in order to handle all the
 //     * previously called Sonata\AdminBundle\Datagrid\ProxyQueryInterface methods.
 //     *
 //     * @return Query
--- a/src/Datagrid/ProxyQueryInterface.php
+++ b/src/Datagrid/ProxyQueryInterface.php
@@ -13,13 +13,28 @@ declare(strict_types=1);
 
 namespace Sonata\DoctrineORMAdminBundle\Datagrid;
 
+use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface as BaseProxyQueryInterface;
 
+/**
+ * NEXT_MAJOR: Remove the "@method" and uncomment the methods instead.
+ *
+ * @method Query getDoctrineQuery()
+ */
 interface ProxyQueryInterface extends BaseProxyQueryInterface
 {
     /**
      * @return QueryBuilder
      */
     public function getQueryBuilder();
+
+//    /**
+//     * This method should be preferred over `$this->getQueryBuilder()->getQuery()`
+//     * since some changed are done to the query builder in order to handle all the
+//     * previously called Sonata\AdminBundle\Datagrid\ProxyQueryInterface methods.
+//     *
+//     * @return Query
+//     */
+//    public function getDoctrineQuery(): Query;
 }
--- a/src/Exporter/DataSource.php
+++ b/src/Exporter/DataSource.php
@@ -14,9 +14,10 @@ declare(strict_types=1);
 namespace Sonata\DoctrineORMAdminBundle\Exporter;
 
 use Doctrine\ORM\Query;
-use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
+use Sonata\AdminBundle\Datagrid\ProxyQueryInterface as BaseProxyQueryInterface;
 use Sonata\AdminBundle\Exporter\DataSourceInterface;
 use Sonata\DoctrineORMAdminBundle\Datagrid\OrderByToSelectWalker;
+use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface;
 use Sonata\Exporter\Source\DoctrineORMQuerySourceIterator;
 use Sonata\Exporter\Source\SourceIteratorInterface;
 
@@ -25,20 +26,38 @@ use Sonata\Exporter\Source\SourceIteratorInterface;
  */
 class DataSource implements DataSourceInterface
 {
-    public function createIterator(ProxyQueryInterface $query, array $fields): SourceIteratorInterface
+    public function createIterator(BaseProxyQueryInterface $query, array $fields): SourceIteratorInterface
     {
-        $query->select('DISTINCT '.current($query->getRootAliases()));
-        $query->setFirstResult(null);
-        $query->setMaxResults(null);
+        // NEXT_MAJOR: Keep the else part and throw a TypeError instead.
+        if (!$query instanceof ProxyQueryInterface) {
+            @trigger_error(sprintf(
+                'Passing %s as argument 1 to %s() is deprecated since sonata-project/doctrine-orm-admin-bundle 3.27'
+                .' and will throw a \TypeError error in version 4.0. You MUST pass an instance of %s instead.',
+                \get_class($query),
+                __METHOD__,
+                ProxyQueryInterface::class
+            ));
 
-        $sortBy = $query->getSortBy();
+            $query->select('DISTINCT '.current($query->getRootAliases()));
+            $query->setFirstResult(null);
+            $query->setMaxResults(null);
 
-        if (null !== $sortBy) {
-            $query->addOrderBy($sortBy, $query->getSortOrder());
-            $doctrineQuery = $query->getQuery();
-            $doctrineQuery->setHint(Query::HINT_CUSTOM_TREE_WALKERS, [OrderByToSelectWalker::class]);
+            $sortBy = $query->getSortBy();
+            if (null !== $sortBy) {
+                $query->addOrderBy($sortBy, $query->getSortOrder());
+                $doctrineQuery = $query->getQuery();
+                $doctrineQuery->setHint(Query::HINT_CUSTOM_TREE_WALKERS, [OrderByToSelectWalker::class]);
+            } else {
+                $doctrineQuery = $query->getQuery();
+            }
         } else {
-            $doctrineQuery = $query->getQuery();
+            // Distinct is need to iterate, even if group by is used
+            // @see https://github.com/doctrine/orm/issues/5868
+            $query->getQueryBuilder()->distinct();
+            $query->setFirstResult(null);
+            $query->setMaxResults(null);
+
+            $doctrineQuery = $query->getDoctrineQuery();
         }
 
         return new DoctrineORMQuerySourceIterator($doctrineQuery, $fields);
--- a/src/Exporter/DataSource.php
+++ b/src/Exporter/DataSource.php
@@ -51,7 +51,7 @@ class DataSource implements DataSourceInterface
                 $doctrineQuery = $query->getQuery();
             }
         } else {
-            // Distinct is need to iterate, even if group by is used
+            // Distinct is needed to iterate, even if group by is used
             // @see https://github.com/doctrine/orm/issues/5868
             $query->getQueryBuilder()->distinct();
             $query->setFirstResult(null);
--- a/tests/Exporter/DataSourceTest.php
+++ b/tests/Exporter/DataSourceTest.php
@@ -18,9 +18,11 @@ use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
 use PHPUnit\Framework\TestCase;
+use Sonata\AdminBundle\Datagrid\ProxyQueryInterface as BaseProxyQueryInterface;
 use Sonata\DoctrineORMAdminBundle\Datagrid\OrderByToSelectWalker;
-use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery;
+use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface;
 use Sonata\DoctrineORMAdminBundle\Exporter\DataSource;
+use Sonata\Exporter\Source\DoctrineORMQuerySourceIterator;
 
 final class DataSourceTest extends TestCase
 {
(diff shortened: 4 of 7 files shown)


greg0ire:
[This post](https://markcarrington.dev/2020/12/08/dataverse-paging-with-distinct/) suggests that you can have issues when using pagination and no sort, but what you are saying is surprising to say the least. [I would expect `LIMIT` to be performed after `DISTINCT`](https://stackoverflow.com/a/8175356/353612). You did not mention `LIMIT` though, you mentioned pagination. Is this maybe based on subqueries? If the limit happens in a subquery, and the `DISTINCT` is applied on the result, I could imagine this happening.

VincentLanglet:
> Also, your PR adds the `GROUP BY`, but does not remove the `DISTINCT`, why is that?

I didn't know if I could remove it.

> Can you maybe share the final SQL that is executed when you observe your bug?

Sure

The query page 4:
```
SELECT DISTINCT a0_.id AS id_0, a1_.usagename AS usagename_1, a1_.firstname AS firstname_2, a1_.id AS id_3, a0_.id AS id_4 FROM asv_contract a0_ LEFT JOIN asv_client_role a2_ ON a0_.id = a2_.contract_id LEFT JOIN asv_external_client_id a3_ ON a2_.external_client_id_id = a3_.id LEFT JOIN asv_client a1_ ON a3_.client_id = a1_.id WHERE (a1_.portfolio_id IS NULL OR a1_.portfolio_id IN (1, 5, 3, 4, 2)) AND a1_.usagename LIKE '%DELABIE%' ORDER BY a1_.usagename ASC, a1_.firstname ASC, a1_.id ASC, a0_.id ASC LIMIT 32 OFFSET 96;
```
I get 32 results.

If I tried
```
SELECT DISTINCT a0_.id AS id_0, a1_.usagename AS usagename_1, a1_.firstname AS firstname_2, a1_.id AS id_3, a0_.id AS id_4 FROM asv_contract a0_ LEFT JOIN asv_client_role a2_ ON a0_.id = a2_.contract_id LEFT JOIN asv_external_client_id a3_ ON a2_.external_client_id_id = a3_.id LEFT JOIN asv_client a1_ ON a3_.client_id = a1_.id WHERE (a1_.portfolio_id IS NULL OR a1_.portfolio_id IN (1, 5, 3, 4, 2)) AND a1_.usagename LIKE '%DELABIE%' ORDER BY a1_.usagename ASC, a1_.firstname ASC, a1_.id ASC, a0_.id ASC 
```
I get 133 results. (But 102 with the group by)

This is because I have multiple rows with the same id with the Distinct:
image

I know understand the following part
```php
        // for SELECT DISTINCT, ORDER BY expressions must appear in idxSelect list
        /* Consider
            SELECT DISTINCT x FROM tab ORDER BY y;
        For any particular x-value in the table there might be many different y
        values.  Which one will you use to sort that x-value in the output?
        */
        $queryId = $queryBuilderId->getQuery();
        $queryId->setHint(Query::HINT_CUSTOM_TREE_WALKERS, [OrderByToSelectWalker::class]);
```

The distinct is applied on the id AND on the fields used for the order by.
But I'm not sure it makes sens two displays multiple times the same contract in the admin because we don't know which value we should use to order.

The count is made this way:
```
SELECT COUNT(*) AS dctrn_count FROM (SELECT DISTINCT id_0 FROM (SELECT a0_.id AS id_0, a0_.admin_comment AS admin_comment_1, a0_.external_contract_id AS external_contract_id_2, a0_.effective_date AS effective_date_3, a0_.end_date AS end_date_4, a0_.updated_date AS updated_date_5, a0_.sent_to_partner_date AS sent_to_partner_date_6, a0_.created_by_grc_date AS created_by_grc_date_7, a0_.completed_by_grc_date AS completed_by_grc_date_8, a0_.pledge AS pledge_9, a0_.management_mandate AS management_mandate_10, a0_.contract_status AS contract_status_11, a0_.buy_worth AS buy_worth_12, a0_.current_advances AS current_advances_13, a0_.is_complete AS is_complete_14, a0_.has_been_matched_with_created_contract AS has_been_matched_with_created_contract_15, a0_.distribution_in_progress AS distribution_in_progress_16, a0_.cash_amount AS cash_amount_17, a0_.manual_input AS manual_input_18, a0_.bank_account_number AS bank_account_number_19, a0_.received_at AS received_at_20, a0_.current_management_mode_for_partner AS current_management_mode_for_partner_21, a0_.management_mode_change_date AS management_mode_change_date_22, a0_.business_identifier_code AS business_identifier_code_23, a0_.origin AS origin_24, a0_.pending_reasons AS pending_reasons_25 FROM asv_contract a0_ LEFT JOIN asv_client_role a1_ ON a0_.id = a1_.contract_id LEFT JOIN asv_external_client_id a2_ ON a1_.external_client_id_id = a2_.id LEFT JOIN asv_client a3_ ON a2_.client_id = a3_.id WHERE (a3_.portfolio_id IS NULL OR a3_.portfolio_id IN (1, 5, 3, 4, 2)) AND a3_.usagename LIKE '%DELABIE%' ORDER BY a3_.usagename ASC, a3_.firstname ASC, a3_.id ASC) dctrn_result) dctrn_table;
```
I get 102.

This is done by the Paginator introduced recently in the Pager, that's why I only have the bug now.
You can notice that the Paginator only apply the distinct to the id.

greg0ire:
I'm not sure what your requirements are: do you mean to display one line per contract, or should there be 2 lines if one contract happens to have 2 clients?

VincentLanglet:
> I'm not sure what your requirements are: do you mean to display one line per contract, or should there be 2 lines if one contract happens to have 2 clients? Anyway, displaying a many-to-many in a table is not going to be simple.

I wanted to
- display only the contract I'm allowed to see = the contract of the client I'm allowed to see. This is why there is a join with client, there is also a `where client.portfolio in ....`.
- order the list by the client firstname/usagename of the contract.

In the SonataAdmin, the query is only returning an array of object, `Contract[]` here. Then for each contract, I display the contract id in a cell, the contract name in a second cell and all the client of the contract in a third cell (Sonata handle without any issue the OTM or MTM relation in the ListMapper).

The only issue I get is with the query which is returning multiples times the same contract (which I don't want).
I suppose this can be debatable but since, again, we're returning `Contract[]` I don't see it would be interesting to returning multiple times the same contract since it lead to the same row in the table. (If the contract is owned by client1 and client2, I don't have a row with the contract and client1 and a row with the contract and client2 but 2 rows with the contract and client1 and client2).

And since the Pager returns a count different of the count of my query results, all the results are not displayed, so there is definitely something to do.

VincentLanglet:
I remembered about this issue https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1225
And It was also solved by changing distinct into groupBy.

src/Datagrid/OrderByToSelectWalker.php
```diff
@@ -30,6 +34,16 @@
  */
 final class OrderByToSelectWalker extends TreeWalkerAdapter
 {
+    public function __construct($query, $parserResult, array $queryComponents)
+    {
+        parent::__construct($query, $parserResult, $queryComponents);
+
+        @trigger_error(sprintf(
+            'The %s class is deprecated since version 3.x and will be removed in 4.0.',
```
phansys, line 42:
```suggestion
            'The %s class is deprecated since sonata-project/doctrine-orm-admin-bundle 3.x and will be removed in version 4.0.',
```

phansys (changes requested):

src/Datagrid/OrderByToSelectWalker.php
VincentLanglet: Done

VincentLanglet:
WDYT about also deprecating `ProxyQuery::setDistinct` and `ProxyQuery::isDistinct` ? @sonata-project/contributors

greg0ire:
> With groupby, it's applied BEFORE the pagination, so it became

```
-----------
contract | client
-----------
1 | 1
-----------
2 | 1
-----------
3 | 1
-----------
```

Ok so… we're losing a lot of information here, are we not? What's your point?

> But I'm not sure it makes sens two displays multiple times the same contract
> in the admin because we don't know which value we should use to order.

Well we can always use the id of the  joined table, can we not? To me, the
issue with displaying the same contract multiple times is that it makes no
sense in a contract list. Or maybe it should be shown like follows:

| contract id | Usage Name | First name  | client id |
|-------------|------------|-------------|-----------|
| 16538       | DELABIE    | Guillaume   | 15696     |
| 16549       | DELABIE    | Jean-Claude | 15746     |
|             | DELABIE    | Maud        | 15749     |

Not printing ids that don't change might help the user spot that this is the
same contract. BTW, please don't use screenshots of text,
it makes it super cumbersome for me to reuse your examples.


There's something peculiar in the SQL request you shared: `a0_.id` is selected
twice, once as `id_0`, then as `id_4`.

> I wanted to display only the contract I'm allowed to see = the contract of
> the client I'm allowed to see. This is why there is a join with client, there
> is also a where client.portfolio in


What does that have to do with the issue at hand? My question was more about
whether you think it's best to have one line per contract, or several if that
contract has many clients.

> I display the contract id in a cell, the contract name in a second cell and
> all the client of the contract in a third cell (Sonata handle without any
> issue the OTM or MTM relation in the ListMapper).

Ok so it's the latter, you want one line, and apparently, that already works
somehow.


> The only issue I get is with the query which is returning multiples times the
> same contract (which I don't want).

If we are talking about the SQL query, I disagree, it's normal, that's how
fetch join work.
If we are talking about the DQL query, I agree, it would be best to have one
`Contract` object hydrated with all its clients if you want to display one line
per contract.

> I suppose this can be debatable but since, again, we're returning Contract[]

You mention `Contract` here, so you must be talking about the DQL query then?

> I don't see it would be interesting to returning multiple times the same
> contract since it lead to the same row in the table. (If the contract is
> owned by client1 and client2, I don't have a row with the contract and
> client1 and a row with the contract and client2 but 2 rows with the contract
> and client1 and client2).

Ah and now you are complaining about multiple rows… so you were talking about
SQL in the end? It's unclear to me.

VincentLanglet:
> What does that have to do with the issue at hand? My question was more about
> whether you think it's best to have one line per contract, or several if that
> contract has many clients.

I think it's best to have one line per contract. That's what I was trying to explain.
The row I was talking about were the row of the Admin table, not the row of the SQL results.

> If we are talking about the SQL query, I disagree, it's normal, [that's how fetch joins work](https://ocramius.github.io/blog/doctrine-orm-optimization-hydration/).
> If we are talking about the DQL query, I agree, it would be best to have one
> `Contract` object hydrated with all its clients if you want to display one line
> per contract.

I know how a join works in SQL, I was talking about the DQL indeed.

> Ah and now you are complaining about multiple rows… so you were talking about
SQL in the end? It's unclear to me.

If the contract2 has two clients, the DQL is returning something like
```
[$contract1, $contract2, $contract3, $contract2]
```
Both $contract2 are obviously fully hydrated. So in the Admin, I get 4 row instead of 3, and the row (of the admin list) 2 and the row (of the admin list) 4 display the same information. 

This is fixed with this PR, now, the result of the query is
```
[$contract1, $contract2, $contract3]
```

VincentLanglet:
I saw https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1225#issuecomment-786833677
I tried to rewrite getFixedQueryBuilder and added comments.

I tried on my project it works well. And this seems leeeeeess complicated than it was before.
But maybe some @sonata-project/contributors will find something I missed ^^'

greg0ire (dismissed):
Although I fear this will break stuff for people, I think it's a move in the right direction, since the code is cleaner and easier to understand.

Consider targeting `master` though? I would be an occasion to rename the method to something more sane, and would give us a good reason to release version 4.0.0 of this bundle. Besides, it might be considered BC-breaky for extending classes to have so much code move from `execute()` to `getFixedQueryBuilder()`. If either of those are copy/pasted adapted to do anything, things night break.

The input is a query builder provided by the user, the output is a query builder "fixed" because it will produce something that is compatible with pagination builder. So… `getFixedQueryBuilder`

VincentLanglet:
@greg0ire I'd like to find a BC way to introduce this change instead of targeting master.

First, I found this one: creating a `getFixedQueryBuilder2` private method and deprecating `getFixedQueryBuilder`.

But then, I discovered that the `DataSource::createIterator` had similar issues to `ProxyQuery::execute`.
- The same results were displayed multiple times in the export
- The sort were different in the export than in the list (because of the order of `addSortBy` calls)

So the way we were fixing the queryBuilder need to be public in order to be called in the DataSource::createIterator method.
I think the best way is to provide a `getDoctrineQuery` method in the public Api.

WDYT about this updated PR ? :)

src/Datagrid/ProxyQueryInterface.php
```diff
@@ -13,13 +13,28 @@
...
 interface ProxyQueryInterface extends BaseProxyQueryInterface
 {
     /**
      * @return QueryBuilder
      */
     public function getQueryBuilder();
+
+//    /**
+//     * This method should be preferred over `$this->getQueryBuilder()->getQuery()`
+//     * since some changed are done to the query builder in order to handle all the
```
greg0ire, line 34:
```suggestion
//     * since some changes are done to the query builder in order to handle all the
```

src/Exporter/DataSource.php
```diff
@@ -25,20 +26,38 @@
...
+            if (null !== $sortBy) {
+                $query->addOrderBy($sortBy, $query->getSortOrder());
+                $doctrineQuery = $query->getQuery();
+                $doctrineQuery->setHint(Query::HINT_CUSTOM_TREE_WALKERS, [OrderByToSelectWalker::class]);
+            } else {
+                $doctrineQuery = $query->getQuery();
+            }
         } else {
-            $doctrineQuery = $query->getQuery();
+            // Distinct is need to iterate, even if group by is used
```
greg0ire, line 54:
```suggestion
            // Distinct is needed to iterate, even if group by is used
```

greg0ire (commented):
Looks like a good idea 👍

src/Datagrid/ProxyQueryInterface.php
VincentLanglet: done

src/Exporter/DataSource.php
VincentLanglet: done

tests/Exporter/DataSourceTest.php
```diff
@@ -86,4 +85,32 @@ public function testSortableInDataSourceIterator(
             $this->assertContains(OrderByToSelectWalker::class, $hints[$key]);
         }
     }
+
+    public function testCreateIterator()
```
greg0ire, line 89:
```suggestion
    public function testCreateIterator(): void
```

src/Datagrid/OrderByToSelectWalker.php
```diff
@@ -30,6 +34,16 @@
  */
 final class OrderByToSelectWalker extends TreeWalkerAdapter
 {
+    public function __construct($query, $parserResult, array $queryComponents)
```
phansys, line 37: IMO, `Doctrine\ORM\AbstractQuery` and `Doctrine\ORM\Query\ParserResult` should be used for the type declarations in argument 1 and 2.

phansys (changes requested):

src/Datagrid/OrderByToSelectWalker.php
VincentLanglet: There is no type used in the parent constructor.

And the OrderByToSelectWalker class is deprecated.

phansys: I saw that, but the the properties that these arguments set have these type hints declared as docblocks.

VincentLanglet: The interface TreeWalker has no type in the constructor, so I cannot add them.

phansys (approved):

greg0ire:
Thanks @VincentLanglet !

PR #1325, opened 2021-03-02
Prepare Phpstan level 2

VincentLanglet:
We're accessing public property of ClassMetadataInfo, so this cannot be a ClassMetadataInterface.

### DIFF
--- a/src/Filter/Filter.php
+++ b/src/Filter/Filter.php
@@ -26,6 +26,10 @@ abstract class Filter extends BaseFilter
 
     /**
      * Apply the filter to the QueryBuilder instance.
+     *
+     * @param mixed[] $data
+     *
+     * @phpstan-param array{type?: string|int, value?: mixed} $data
      */
     abstract public function filter(ProxyQueryInterface $query, string $alias, string $field, array $data): void;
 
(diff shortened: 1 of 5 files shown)


phansys (approved):

core23 (approved):

src/Filter/Filter.php
```diff
@@ -26,6 +26,10 @@ abstract class Filter extends BaseFilter
 
     /**
      * Apply the filter to the QueryBuilder instance.
+     *
+     * @param mixed[] $data
+     *
+     * @phpstan-param array{type?: string|int, value?: mixed} $data
```
core23, line 32: Can you port this to the stable branch?

src/Model/ModelManager.php
```diff
@@ -403,13 +405,14 @@ public function addIdentifiersToQuery(string $class, ProxyQueryInterface $query,
 
     public function batchDelete(string $class, ProxyQueryInterface $query): void
     {
-        $query->select('DISTINCT '.current($query->getRootAliases()));
+        $qb = $query->getQueryBuilder();
```
core23, line 408: Same for this

src/Filter/Filter.php
VincentLanglet: This is already in the stable branch

src/Model/ModelManager.php
VincentLanglet: Check https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1326 :)

PR #1333, opened 2021-03-03
Deprecate passing arguments to execute method

VincentLanglet:
## Changelog

```markdown
### Deprecated
- Deprecate passing arguments to `ProxyQuery::execute()` method
```

### DIFF
--- a/src/Datagrid/ProxyQuery.php
+++ b/src/Datagrid/ProxyQuery.php
@@ -191,6 +191,14 @@ class ProxyQuery implements ProxyQueryInterface
 
     public function execute(array $params = [], $hydrationMode = null)
     {
+        // NEXT_MAJOR: Remove this check and update method signature to `execute()`.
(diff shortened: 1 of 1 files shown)


jordisala1991 (approved):

wbloszyk (approved):

src/Datagrid/ProxyQuery.php
```diff
@@ -191,6 +191,14 @@ final public function isDistinct()
 
     public function execute(array $params = [], $hydrationMode = null)
     {
+        // NEXT_MAJOR: Remove this check and update method signature to `execute()`.
+        if ([] !== $params || null !== $hydrationMode) {
```
phansys, line 195: We should use `func_num_args()` instead of checking the default values.
Even passing the default values should be deprecated.

phansys: Addressed in #1334.

PR #1337, opened 2021-03-06
Remove next_major comment

VincentLanglet:


VincentLanglet:
@phansys It works ?!

phansys:
> @phansys It works ?!

Yes, I guess.
At least I don't see the self deprecation notice in the tests.
It was:
```
1x: The "Sonata\DoctrineORMAdminBundle\Filter\NullFilter" class is considered final. It may change without further notice as of its next major version. You should not extend it from "Sonata\DoctrineORMAdminBundle\Filter\EmptyFilter".
    1x in FieldDescriptionTest::testOptions from Sonata\DoctrineORMAdminBundle\Tests\Admin
```
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/runs/2043015696?check_suite_focus=true#step:7:133

VincentLanglet:
> > @phansys It works ?!
> 
> Yes, I guess.
> At least I don't see the self deprecation notice in the tests.
> It was:
> 
> ```
> 1x: The "Sonata\DoctrineORMAdminBundle\Filter\NullFilter" class is considered final. It may change without further notice as of its next major version. You should not extend it from "Sonata\DoctrineORMAdminBundle\Filter\EmptyFilter".
>     1x in FieldDescriptionTest::testOptions from Sonata\DoctrineORMAdminBundle\Tests\Admin
> ```
> 
> https://github.com/sonata-project/SonataDoctrineORMAdminBundle/runs/2043015696?check_suite_focus=true#step:7:133

Yes I saw that, but now it's not here anymore... Don't know what changed but we might remove the NEXT_MAJOR comment, wdyt ?

phansys:
Given our current configuration for `SYMFONY_DEPRECATIONS_HELPER`, I think we could safely remove the comment.
Even, when we don't have a clear idea how this test is related to the triggered deprecation.
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/df9e41a4167aeea0b753cede0673b48ce12e99af/phpunit.xml.dist#L35

PR #1350, opened 2021-03-13
Add FieldDescriptionFactory

### LINKED ISSUE #1296: ModelManagerInterface and getParentMetadataForProperty()
## Feature Request

Inside the `AbstractTypeGuesser` we are [calling](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Guesser/AbstractTypeGuesser.php#L35) `Sonata\AdminBundle\Model\ModelManagerInterface::getParentMetadataForProperty` but this method is not part of that contract. 

So we are actually expecting a `Sonata\DoctrineORMAdminBundle\Model\ModelManager` for it to work.

As @VincentLanglet [suggested](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1295#discussion_r570043905) we could 

- extend the `ModelManagerInterface` and add this method
- move this method somewhere else

Or any other ideas?

### LINKED ISSUE #1348: Add FieldDescriptionFactory + resolve TypeGuesser deprecations
We need this for releasing 4.0.

See https://github.com/sonata-project/SonataAdminBundle/pull/6854

Related ODM PR: https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/pull/531

@franmomu are you looking into this? Let me know if you need any help

VincentLanglet:
## Subject

Closes #1348.
Closes #1296.

## Changelog
```markdown
### Added
- Added `FieldDescriptionFactory` class

### Deprecated
- `AbstractTypeGuesser` class
- `FilterTypeGuesser::guessType()`
- `TypeGuesser::guessType()`
- `ModelManager:: getParentMetadataForProperty()`
- `ModelManager:: getNewFieldDescriptionInstance()`
- `ModelManager:: getModelInstance()`
```

### DIFF
--- a/src/FieldDescription/FieldDescriptionFactory.php
+++ b/src/FieldDescription/FieldDescriptionFactory.php
@@ -1,0 +1,110 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DoctrineORMAdminBundle\FieldDescription;
+
+use Doctrine\ORM\EntityManager;
+use Doctrine\Persistence\Mapping\ClassMetadata;
+use Sonata\AdminBundle\Admin\FieldDescriptionInterface;
+use Sonata\AdminBundle\FieldDescription\FieldDescriptionFactoryInterface;
+use Sonata\DoctrineORMAdminBundle\Admin\FieldDescription;
+use Symfony\Bridge\Doctrine\ManagerRegistry;
+
+final class FieldDescriptionFactory implements FieldDescriptionFactoryInterface
+{
+    /**
+     * @var ManagerRegistry
+     */
+    private $registry;
+
+    public function __construct(ManagerRegistry $registry)
+    {
+        $this->registry = $registry;
+    }
+
+    public function create(string $class, string $name, array $options = []): FieldDescriptionInterface
+    {
+        if (!isset($options['route']['name'])) {
+            $options['route']['name'] = 'show';
+        }
+
+        if (!isset($options['route']['parameters'])) {
+            $options['route']['parameters'] = [];
+        }
+
+        [$metadata, $propertyName, $parentAssociationMappings] = $this->getParentMetadataForProperty($class, $name);
+
+        return new FieldDescription(
+            $name,
+            $options,
+            $metadata->fieldMappings[$propertyName] ?? [],
+            $metadata->associationMappings[$propertyName] ?? [],
+            $parentAssociationMappings,
+            $propertyName
+        );
+    }
+
+    private function getParentMetadataForProperty(string $baseClass, string $propertyFullName): array
+    {
+        $nameElements = explode('.', $propertyFullName);
+        $lastPropertyName = array_pop($nameElements);
+        $class = $baseClass;
+        $parentAssociationMappings = [];
+
+        foreach ($nameElements as $nameElement) {
+            $metadata = $this->getMetadata($class);
+
+            if (isset($metadata->associationMappings[$nameElement])) {
+                $parentAssociationMappings[] = $metadata->associationMappings[$nameElement];
+                $class = $metadata->getAssociationTargetClass($nameElement);
+
+                continue;
+            }
+
+            break;
+        }
+
+        $properties = \array_slice($nameElements, \count($parentAssociationMappings));
+        $properties[] = $lastPropertyName;
+
+        return [
+            $this->getMetadata($class),
+            implode('.', $properties),
+            $parentAssociationMappings,
+        ];
+    }
+
+    /**
+     * @param class-string $class
+     */
+    private function getMetadata(string $class): ClassMetadata
+    {
+        return $this->getEntityManager($class)->getMetadataFactory()->getMetadataFor($class);
+    }
+
+    /**
+     * @param class-string $class
+     *
+     * @throw \UnexpectedValueException
+     */
+    private function getEntityManager(string $class): EntityManager
+    {
+        $em = $this->registry->getManagerForClass($class);
+
+        if (!$em instanceof EntityManager) {
+            throw new \UnexpectedValueException(sprintf('No entity manager defined for class "%s".', $class));
+        }
+
+        return $em;
+    }
+}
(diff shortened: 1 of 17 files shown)


src/FieldDescription/FieldDescriptionFactory.php
```diff
@@ -0,0 +1,132 @@
...
+     * @param class-string $class
+     *
+     * @throw \RuntimeException
+     */
+    private function getEntityManager(string $class): EntityManager
+    {
+        $em = $this->registry->getManagerForClass($class);
+
+        if (!$em instanceof EntityManager) {
+            throw new \RuntimeException(sprintf('No entity manager defined for class %s', $class));
```
phansys, line 127: IMO, if we must throw an exception based on an unmet condition from `instanceof`, the used type should be `\TypeError`.

VincentLanglet: It's more a `if (null === $em)`check, but the instanceof is for phpstan.  `getManagerForClass` is flagged as returning a `ObjectManager|null` but this can't happen when the registry is a `ManagerRegistry`.

phansys: I understand. Is there a criteria we can use around these checks? I mean, I saw some cases where we are using `assert()` for scenarios like this.

VincentLanglet: Since it's possible to implement the interface with a method which return an ObjectManager, I think it's safer to keep the instanceof check.

Currently the message is more a `Entity manager not found` message.
If we throw a typeError, the message should be different.

phansys: What I suggest is `assert($em instanceof EntityManager)`. IMO, the important thing here is to be consistent with the way we handle these cases.

VincentLanglet: We often use exception.

Imo the rule should be
- If it's possible -> exception
- If it's not -> assert

phansys: There are cases where there is no possible to use exceptions?
Note that I'm referring to cases where these constructions are required to make the CS tools happy, rather than the users.

VincentLanglet: > There are cases where there is no possible to use exceptions?

I mean, there are cases like here where `!$em instanceof EntityManager` can be true, so it's better to use an exception than an assert.
And there are case like https://github.com/sonata-project/SonataAdminBundle/blob/3.x/src/DependencyInjection/Compiler/AddDependencyCallsCompilerPass.php#L114 where the assert is always true. But phpstan doesn't know this. Using an exception would be pretty useless.

VincentLanglet: IMHO `UnexpectedValueException` is made for this.

src/FieldDescription/FieldDescriptionFactory.php
```diff
@@ -0,0 +1,132 @@
...
+     * @param class-string $class
+     *
+     * @throw \UnexpectedValueException
+     */
+    private function getEntityManager(string $class): EntityManager
+    {
+        $em = $this->registry->getManagerForClass($class);
+
+        if (!$em instanceof EntityManager) {
+            throw new \UnexpectedValueException(sprintf('No entity manager defined for class %s', $class));
```
phansys, line 127:
```suggestion
            throw new \UnexpectedValueException(sprintf('No entity manager defined for class "%s".', $class));
```

phansys (approved):

jordisala1991 (approved):

OskarStark (approved):

franmomu:
Hi, I'm sorry I'm being quite busy these weeks (I hope at the end of this week this changes), about this, it is fine as it is and it works, but after giving some thought: 

There are like 4 types of field description depending of its use (_form_, _list_, _show_ and _filter_, _list_ and _show_ use the same types), the factory is guessing the one for _show_ and _list_, so I wasn't sure if we should guess this type using a generic factory or leave the specific guessing where it is needed.

VincentLanglet:
> Hi, I'm sorry I'm being quite busy these weeks (I hope at the end of this week this changes), about this, it is fine as it is and it works, but after giving some thought:
> 
> There are like 4 types of field description depending of its use (_form_, _list_, _show_ and _filter_, _list_ and _show_ use the same types), the factory is guessing the one for _show_ and _list_, so I wasn't sure if we should guess this type using a generic factory or leave the specific guessing where it is needed.

Indeed, I think this will break the filters in master branch, since the default type set will be the one use by show/list.
I'll make the change to remove the guesser from the FieldDescriptionFactory.

VincentLanglet:
WDYT of this new version @sonata-project/contributors

VincentLanglet:
Friendly ping @sonata-project/contributors
I have some free time so I'd like to work on fixing the master branch.

phansys:
> Friendly ping @sonata-project/contributors
> I have some free time so I'd like to work on fixing the master branch.

I just leave one comment: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1350#pullrequestreview-612773068.

src/FieldDescription/FieldDescriptionFactory.php
VincentLanglet: Done

VincentLanglet:
> > Friendly ping @sonata-project/contributors
> > I have some free time so I'd like to work on fixing the master branch.
> 
> I just leave one comment: [#1350 (review)](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1350#pullrequestreview-612773068).

Sorry, didn't see it ; it's fixed.

phansys (approved):

jordisala1991 (approved):

tests/Guesser/TypeGuesserTest.php
```diff
@@ -229,4 +250,113 @@ public function noAssociationData()
...
+        $this->assertSame($expectedType, $guess->getType());
+        $this->assertSame($expectedOptions, $guess->getOptions());
+        $this->assertSame($expectedConfidence, $guess->getConfidence());
+    }
+
+    public function guessDataProvider(): iterable
+    {
+        yield [
+            null,
+            TemplateRegistry::TYPE_TEXT,
```
wbloszyk, line 280:
```suggestion
            // NEXT_MAJOR: FieldDescriptionInterface::TYPE_STRING,
            TemplateRegistry::TYPE_TEXT,
```

tests/Guesser/TypeGuesserTest.php
```diff
@@ -229,4 +250,113 @@ public function noAssociationData()
...
+        yield [
+            'integer',
+            FieldDescriptionInterface::TYPE_INTEGER,
+            [],
+            Guess::MEDIUM_CONFIDENCE,
+        ];
+
+        yield [
+            'string',
+            TemplateRegistry::TYPE_TEXT,
```
wbloszyk, line 336:
```suggestion
            // NEXT_MAJOR: FieldDescriptionInterface::TYPE_STRING,
            TemplateRegistry::TYPE_TEXT,
```

wbloszyk (approved):
I add 2 suggestions, anyway you can ignore it when you want fix it direcly in master after merge this PR.

PR #1354, opened 2021-03-16
Add a test for fetch join queries

franmomu:
This test works on `3.30`, I haven't check why, but looking at the changelog I guess it's because of the changes introduced in https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1319.

It is produced when using a fetch-join query in `AdminInterface::configureQuery `, in this case the author has two books, but now it returns just 1.

Basically, the number of books should be 2:

image

### DIFF
--- a/src/Datagrid/ProxyQuery.php
+++ b/src/Datagrid/ProxyQuery.php
@@ -253,18 +253,15 @@ class ProxyQuery implements ProxyQueryInterface
             }
         }
 
-        $queryBuilder->resetDQLPart('groupBy');
-
         foreach ($identifierFields as $identifierField) {
             $field = $rootAlias.'.'.$identifierField;
 
-            $queryBuilder->addGroupBy($field);
             if (!\in_array($field, $existingOrders, true)) {
                 $queryBuilder->addOrderBy($field, $this->getSortOrder());
             }
         }
 
-        return $queryBuilder->getQuery();
+        return $queryBuilder->distinct()->getQuery();
     }
 
     public function setSortBy($parentAssociationMappings, $fieldMapping)
--- a/tests/App/Admin/AuthorAdmin.php
+++ b/tests/App/Admin/AuthorAdmin.php
@@ -14,9 +14,13 @@ declare(strict_types=1);
 namespace Sonata\DoctrineORMAdminBundle\Tests\App\Admin;
 
 use Sonata\AdminBundle\Admin\AbstractAdmin;
+use Sonata\AdminBundle\Admin\FieldDescriptionInterface;
 use Sonata\AdminBundle\Datagrid\DatagridMapper;
 use Sonata\AdminBundle\Datagrid\ListMapper;
+use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
 use Sonata\AdminBundle\Form\FormMapper;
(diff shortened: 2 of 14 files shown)


VincentLanglet:
@franmomu Yes indeed, this change was introduced by #1319 ; but it was considered as a bug fix:
```
### Fixed
- Do not display multiple times the same row in the admin list and the export list.
```

You're looking for the `AuthorAdmin` so the author should be display only once.
If you want one like by book, you should display a `BookAdmin` IMHO.

One of the argument in favor of this change was the following:
With an author A who wrote the books A1 and A2, and an author B who wrote a books B1: 
- The Pager were displaying only one page
- The Datagrid were displaying two line for author A.
So the author B were never displayed.

franmomu:
> You're looking for the `AuthorAdmin` so the author should be display only once.
> If you want one like by book, you should display a `BookAdmin` IMHO.

I've updated the first comment with a screenshot to make the current problem clearer. 

Basically, if you use a fetch-join query in order to reduce the number of queries, then the object retrieved is not well populated.

VincentLanglet:
> > You're looking for the `AuthorAdmin` so the author should be display only once.
> > If you want one like by book, you should display a `BookAdmin` IMHO.
> 
> I've updated the first comment with a screenshot to make the current problem clearer.

I don't see any, can you try again to add the screenshot ?

franmomu:
I guess I didn't press `update comment` 🤦

VincentLanglet:
Removing `->addSelect('book')` fix the issue.
It's not *directly* related to the join.

What's your reason for adding a select ? I use a lot of join, but never used `addSelect`.

franmomu:
For [fetch joins](https://www.doctrine-project.org/projects/doctrine-orm/en/2.8/reference/dql-doctrine-query-language.html#joins), I used them quite often to reduce the number of queries in simple queries. 

Let's say you show 32 authors, without the fetch-join, there will be 32 "extra" queries to fetch the related books, with fetch-join just 2. 

If for example we add another Entity `Reader` and a book as many Readers and we want to show how many readers an author has, following the previous example showing 32 authors, each author with 10 books, without fetch-join, there will be 32 + 320 extra queries, while adding fetch-joins will be just 2.

VincentLanglet:
This could be the reason behind `getFixedQueryBuilder`
- One query was made to get all the id displayed.
- One query was made with the condition `Where id in (:id_to_display)`.

Maybe it's possible to introduce this behavior back in `execute`.

But the changed made fixed two issues: 
- https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1225
- Duplicate rows in the list
I'd like to avoid getting regressions on these...

The second one was fixed by the change `distinct => groupBy`. 
The first one is because we sometimes overriding the `select` part.
- When we were overriding to select only ids (fixed)
- In the export: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Exporter/DataSource.php#L41

Do you have a solution ?

franmomu:
TBH I have no idea, I'm not familiar with that code.

VincentLanglet:
Closing in favor of https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1368

PR #1358, opened 2021-03-20
Fix filter query when using "OR" conditions

phansys:
## Subject
Fix filter query when using "OR" conditions.

I am targeting this branch, because these changes respect BC.

Rework of #925, keeping BC.
Fixes sonata-project/SonataAdminBundle#2850.
Fixes sonata-project/SonataAdminBundle#3368.
Fixes sonata-project/SonataAdminBundle#5569.

## Changelog

```markdown
### Fixed
- Resulting `WHERE` clause from `Filter::applyWhere()` when using `OR` conditions on queries that already have previous conditions.
```

## To do

- [x] Find a way to tell the filter it must add the `AND WHERE` clause for first time without using a static variable;
- [x] Add tests.

### DIFF
--- a/src/Filter/Filter.php
+++ b/src/Filter/Filter.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\DoctrineORMAdminBundle\Filter;
 
+use Doctrine\ORM\Query\Expr\Orx;
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface as BaseProxyQueryInterface;
 use Sonata\AdminBundle\Filter\Filter as BaseFilter;
 use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface;
@@ -23,6 +24,15 @@ abstract class Filter extends BaseFilter
      * @var bool
      */
     protected $active = false;
+
+    /**
+     * Holds an array of `orX` expressions used by each admin when the condition
+     * equals the value on `FilterInterface::CONDITION_OR`, using the admin code
+     * as index.
+     *
+     * @var array<string, Orx>
+     */
+    private static $orExpressionsByAdmin = [];
 
     /**
      * NEXT_MAJOR change $query type for Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface.
@@ -96,7 +106,7 @@ abstract class Filter extends BaseFilter
         }
 
         if (self::CONDITION_OR === $this->getCondition()) {
-            $query->getQueryBuilder()->orWhere($parameter);
+            $this->addOrParameter($query, $parameter);
         } else {
             $query->getQueryBuilder()->andWhere($parameter);
         }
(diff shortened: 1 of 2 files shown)


phansys:
We could evaluate if in the next major, we could provide the `Xor` expression used for the search from the datagrid at the AdminBundle.
By now, I think we could rely in this workaround.
@VincentLanglet, since you was experiencing this issue, could you please test these changes in a real project?

VincentLanglet (commented):
If I understand correctly.

If you've add a `where(Foo)` is the configure query.
And you're using a filter1 with AND, a filter3 with OR, a filter2 with AND and a filter4 with OR.

You will get where Foo AND filter1,
then where Foo AND filter1 AND (filter3)
then where Foo AND filter1 AND (filter3) AND filter2
then where Foo AND filter1 AND (filter3 OR filter4) AND filter2

So what you're calling a root filter condition, could more related to the Or. `getOrFilterCondition` ? Same for the marker.

Would it be possible to add test unit ?

phansys:
>So what you're calling a root filter condition, could more related to the Or. getOrFilterCondition ? Same for the marker.

Let's say you have the following datagrid query:
```sql
SELECT ...
FROM model m
WHERE
    m.enabled = true
    AND m.updated_at IS NULL
;
```
The changes introduced in this PR will perform these actions:
1. Add an `AND` condition inside the `WHERE` clause, using what I called, a marker, enclosed with parentheses:
```sql
SELECT ...
FROM model m
WHERE
    m.enabled = true
    AND m.updated_at IS NULL
    AND (
        "sonata_admin.datagrid.filter_query.marker" IS NULL
    )
;
```
This marker is used later for all the filters to identify the `AND` group where the `OR` conditions must be placed:
```sql
SELECT ...
FROM model m
WHERE
    m.enabled = true
    AND m.updated_at IS NULL
    AND (
        "sonata_admin.datagrid.filter_query.marker" IS NULL
        OR m.name LIKE "%my search criteria"
        OR m.lastname LIKE "%my search criteria"
    )
;
```
>Would it be possible to add test unit ?

Yes, that's in my To Do list, but I'd like to confirm if the results are the expected in other real project before spending time on them. In my case, it is working fine, but please let me know if you have chance to check it on your side.
Thank you in advance.

VincentLanglet:
Yes, so the method `getRootCondition` is returning the part
```
    (
        "sonata_admin.datagrid.filter_query.marker" IS NULL
        OR m.name LIKE "%my search criteria"
        OR m.lastname LIKE "%my search criteria"
    )
```
I don't consider this as the root condition, but only the one related to OrFilter.

phansys:
>I don't consider this as the root condition, but only the one related to OrFilter.

Since `getRootCondition()` is declared in the class `Filter`, I feel naming it `getOrFilterCondition()` like a redundancy, but I have no problem with that change. My main concern is about the behavior and how it's working.

VincentLanglet:
>  My main concern is about the behavior and how it's working.

I will try to use this in the week, but since we've overridden the default search behavior I need some work/time.

src/Filter/Filter.php
```diff
@@ -141,4 +153,57 @@ protected function getNewParameterName(BaseProxyQueryInterface $query)
...
+     * This method uses a marker (":sonata_admin_datagrid_filter_query_marker") in
+     * the resulting DQL in order to identify the corresponding "WHERE (...)" condition
+     * group each time it is required.
+     * It allows to get queries like "WHERE previous_condition = previous_value AND (filter_1 = value OR filter_2 = value OR ...)",
+     * where the logical "OR" operators added by the filters are grouped inside a condition,
+     * instead of having unfolded "WHERE ..." clauses like "WHERE previous_condition = previous_value OR filter_1 = value OR filter_2 = value OR ...",
+     * which will produce undesired results.
+     *
+     * NEXT_MAJOR: Remove this method and let the consumers of this class to set
+     * the `Orx` expression using `setOrExpression()`.
```
VincentLanglet, line 169: How do you want to do this ?

phansys: By allowing cases [like this](https://github.com/sonata-project/SonataAdminBundle/blob/6292c61255e945a0aa1440db260680c40de79a14/src/Search/SearchHandler.php#L93) to define the `Xor` instance that will be used within the filter query, by creating a new one or receiving the instance from the first configured filter and passing it to the next in a chained process.
I guess the second option is more feasible since it doesn't require the consumer (the AdminBundle) to know too much details about the implementation (`Sonata\AdminBundle\Datagrid\ProxyQueryInterface` doesn't provide access to the query builder).
I imagine somethig like this:
```php
$orExpression = null;
foreach ($datagrid->getFilters() as $filter) {
    /** @var FilterInterface $filter */
    $formName = $filter->getFormName();

    if ($filter->getOption('global_search', false)) {
        $filter->setCondition(FilterInterface::CONDITION_OR);
        if (null === $orExpression) {
            // let the first used filter to create the `Xor` expression.
            $orExpression = $filter->getOrExpression();
            // or $orExpression = $filter->getOption('or_expression'); to avoid the API change.
        }
        $filter->setOrExpression($orExpression);
        // or $filter->setOption('or_expression', $orExpression);
        $datagrid->setValue($formName, null, $term);
    } elseif (isset($datagridValues[$formName])) {
        // Remove any previously set filter that is not configured for the global search.
        $datagrid->removeFilter($formName);
    }
}
```

VincentLanglet: It would require to add getOrExpression and setOrExpression to the filter interface.
Isn't this too specific to the persistence bundle ?

phansys: I think no. Please read the comments in the snippet.

VincentLanglet: In your snippet you're still using a `getOrExpression`.
So the `NEXT_MAJOR: Remove this method` comment seems wrong to me ; and the method need to be public

phansys: Sure. What should be removed in the next major is the workaround related to the marker.
Even we could remove that if we can restrict the constraint for the AdminBundle to a version including the mentioned behavior.

phansys: >In your snippet you're still using a getOrExpression.

The commented lines are showing a second option based in options to avoid the API change.
To clarify, I'll represent a clear version of that choice here:
```php
$orExpression = null;
foreach ($datagrid->getFilters() as $filter) {
    /** @var FilterInterface $filter */
    $formName = $filter->getFormName();

    if ($filter->getOption('global_search', false)) {
        $filter->setCondition(FilterInterface::CONDITION_OR);
        if (null === $orExpression) {
            // let the first used filter to create the `Xor` expression.
            $orExpression = $filter->getOption('or_expression');
        } else {
            $filter->setOption('or_expression', $orExpression);
        }
        $datagrid->setValue($formName, null, $term);
    } elseif (isset($datagridValues[$formName])) {
        $datagrid->removeFilter($formName);
    }
}
```
BTW, I'll push a commit with this change soon.

phansys: @VincentLanglet, I updated this method.
In the last commit I'm using a static array to hold the corresponding `Xor` expression used by each admin. I also left the marker logic as fallback until the constraint for the admin bundle can guarantee that the admin code is provided.

phansys:
@sonata-project/contributors, could you please review?
I'd like to make a release including this fix.

franmomu (approved):
Maybe it's obvious, but can we add a simple test that just adds

franmomu (approved):

jordisala1991 (approved):

VincentLanglet (approved):
I still didn't find time to try this.

Does someone did ?

src/Filter/Filter.php
```diff
@@ -141,4 +151,71 @@ protected function getNewParameterName(BaseProxyQueryInterface $query)
...
+     * which will produce undesired results.
+     *
+     * TODO: Remove the logic related to the ":sonata_admin_datagrid_filter_query_marker" marker when
+     * the constraint for "sonata-project/admin-bundle" guarantees that the "admin_code" option is set.
+     *
+     * @param mixed $parameter
+     */
+    private function addOrParameter(BaseProxyQueryInterface $query, $parameter): void
+    {
+        $adminCode = $this->getOption('admin_code');
```
phansys, line 174: Even if we usually will use the admin code which a filter belongs to as value for this option, I think we could use a more agnostic name, since this option has no actual relation with an admin code; it's just a grouping identifier used to know which filters must share the same condition.
WDYT about "group" or something similar?

PR #1363, opened 2021-03-21
Fix phpstan issue with tests

VincentLanglet:
This provide a way to fix phpstan error with filters in test directory, because `query` and `queryParameters` doesn't exist on both `ProxyQuery` and `QueryBuilder`.

### DIFF
--- a/tests/Filter/CallbackFilterTest.php
+++ b/tests/Filter/CallbackFilterTest.php
@@ -34,80 +34,79 @@ class CallbackFilterTest extends FilterTestCase
 
     public function testFilterClosure(): void
     {
-        $builder = new ProxyQuery($this->createQueryBuilderStub());
+        $proxyQuery = new ProxyQuery($this->createQueryBuilderStub());
 
         $filter = new CallbackFilter();
         $filter->initialize('field_name', [
-            'callback' => static function ($builder, $alias, $field, $value) {
-                $builder->andWhere(sprintf('CUSTOM QUERY %s.%s', $alias, $field));
-                $builder->setParameter('value', $value);
+            'callback' => static function (ProxyQuery $query, string $alias, string $field, array $data): bool {
+                $query->getQueryBuilder()->andWhere(sprintf('CUSTOM QUERY %s.%s', $alias, $field));
+                $query->getQueryBuilder()->setParameter('value', $data['value']);
 
                 return true;
             },
         ]);
 
-        $filter->filter($builder, 'alias', 'field', 'myValue');
+        $filter->filter($proxyQuery, 'alias', 'field', ['value' => 'myValue']);
 
-        $this->assertSame(['WHERE CUSTOM QUERY alias.field'], $builder->query);
-        $this->assertSame(['value' => 'myValue'], $builder->queryParameters);
+        $this->assertSameQuery(['WHERE CUSTOM QUERY alias.field'], $proxyQuery);
+        $this->assertSameQueryParameters(['value' => 'myValue'], $proxyQuery);
         $this->assertTrue($filter->isActive());
     }
 
     public function testFilterMethod(): void
     {
-        $builder = new ProxyQuery($this->createQueryBuilderStub());
+        $proxyQuery = new ProxyQuery($this->createQueryBuilderStub());
 
         $filter = new CallbackFilter();
         $filter->initialize('field_name', [
             'callback' => [$this, 'customCallback'],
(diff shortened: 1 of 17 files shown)


tests/Filter/CallbackFilterTest.php
```diff
@@ -81,33 +81,33 @@ public function testFilterException(): void
     {
         $this->expectException(\RuntimeException::class);
```
phansys, line 82: I know this is not part of the issue fixed in this PR, but could you please move this assertion just before the line intended to throw the exception?

tests/Filter/FilterTestCase.php
```diff
@@ -18,17 +18,44 @@
...
+    {
+        $queryBuilder = $proxyQuery->getQueryBuilder();
+        if (!$queryBuilder instanceof TestQueryBuilder) {
+            throw new \InvalidArgumentException('The query builder should be build with "createQueryBuilderStub()".');
+        }
+
+        $this->assertSame($expected, $queryBuilder->queryParameters);
+    }
+
+    public function createQueryBuilderStub(): TestQueryBuilder
```
phansys, line 54:
```suggestion
    final protected function createQueryBuilderStub(): TestQueryBuilder
```

tests/Filter/FilterTestCase.php
```diff
@@ -18,17 +18,44 @@
 use Doctrine\ORM\Query\Expr\Orx;
 use Doctrine\ORM\QueryBuilder;
 use PHPUnit\Framework\TestCase;
+use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery;
+
+class TestQueryBuilder extends QueryBuilder
```
phansys, line 23:
```suggestion
abstract class TestQueryBuilder extends QueryBuilder
```

tests/Filter/FilterTestCase.php
```diff
@@ -18,17 +18,44 @@
...
+{
+    /** @var string[] */
+    public $query = [];
+
+    /** @var string[] */
+    public $queryParameters = [];
+}
 
 class FilterTestCase extends TestCase
```
phansys, line 32:
```suggestion
abstract class FilterTestCase extends TestCase
```

tests/Filter/FilterTestCase.php
```diff
@@ -18,17 +18,44 @@
...
+        if (!$queryBuilder instanceof TestQueryBuilder) {
+            throw new \InvalidArgumentException('The query builder should be build with "createQueryBuilderStub()".');
+        }
+
+        $this->assertSame($expected, $queryBuilder->queryParameters);
+    }
+
+    public function createQueryBuilderStub(): TestQueryBuilder
+    {
+        $testCase = $this;
```
phansys, line 57: We can use `$this`.
```suggestion
```

VincentLanglet: 1) Sonata\DoctrineORMAdminBundle\Tests\Filter\BooleanFilterTest::testFilterArray
Error: Using $this when not in object context

tests/Filter/FilterTestCase.php
```diff
@@ -18,17 +18,43 @@
 use Doctrine\ORM\Query\Expr\Orx;
 use Doctrine\ORM\QueryBuilder;
 use PHPUnit\Framework\TestCase;
+use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery;
 
-class FilterTestCase extends TestCase
+class TestQueryBuilder extends QueryBuilder
```
phansys, line 23: IMO, we should declare this fixture after the PSR-4 class.

VincentLanglet: Done

tests/Filter/FilterTestCase.php
phansys: Did you remove the `static` keyword in the lambda where this variable is used?

phansys (dismissed):

tests/Filter/FilterTestCase.php
phansys: I'd like to use the changes in the tests as base for the checks I need to add at #1358.
@VincentLanglet, could you please confirm if this variable can be removed?

phansys (approved):

jordisala1991 (approved):

jordisala1991:
Thank you @VincentLanglet

Issue #1370, opened 2021-03-25
Create a test for entity inheritance


franmomu:
## Feature Request

We could use the code from: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1361

Issue #1361, opened 2021-03-21
Paginator broken when using entity inheritance


Devristo:
#### Sonata packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'sonata-project/*'
sonata-project/admin-bundle              3.93.0 3.93.0 The missing Symfony Admin Generator
sonata-project/block-bundle              4.5.2  4.5.2  Symfony SonataBlockBundle
sonata-project/cache                     2.1.1  2.1.1  Cache library
sonata-project/doctrine-extensions       1.12.0 1.12.0 Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.31.0 3.31.0 Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/exporter                  2.5.1  2.5.1  Lightweight Exporter library
sonata-project/form-extensions           1.9.0  1.9.0  Symfony form extensions
sonata-project/twig-extensions           1.5.1  1.5.1  Sonata twig extensions
```

</p>
</details>

#### Symfony packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'symfony/*'
symfony/asset                      v4.4.20 v5.2.4  Manages URL generation and versioning of web assets such as CSS stylesheets, JavaScript files and image files
symfony/cache                      v4.4.20 v5.2.4  Provides an extended PSR-6, PSR-16 (and tags) implementation
symfony/cache-contracts            v2.2.0  v2.2.0  Generic abstractions related to caching
symfony/config                     v4.4.20 v5.2.4  Helps you find, load, combine, autofill and validate configuration values of any kind
symfony/console                    v4.4.20 v5.2.5  Eases the creation of beautiful and testable command line interfaces
symfony/debug                      v4.4.20 v4.4.20 Provides tools to ease debugging PHP code
symfony/dependency-injection       v4.4.20 v5.2.5  Allows you to standardize and centralize the way objects are constructed in your application
symfony/doctrine-bridge            v4.4.20 v5.2.5  Provides integration for Doctrine with various Symfony components
symfony/dotenv                     v4.4.20 v5.2.4  Registers environment variables from a .env file
symfony/error-handler              v4.4.20 v5.2.4  Provides tools to manage errors and ease debugging PHP code
symfony/event-dispatcher           v4.4.20 v5.2.4  Provides tools that allow your application components to communicate with each other by dispatching events and listening to them
[...]

PR #1371, opened 2021-03-26
Add inheritance test

franmomu:
Closes https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1370

It is based on https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1361

### DIFF
--- a/tests/App/Admin/SubAdmin.php
+++ b/tests/App/Admin/SubAdmin.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DoctrineORMAdminBundle\Tests\App\Admin;
+
+use Sonata\AdminBundle\Admin\AbstractAdmin;
+use Sonata\AdminBundle\Datagrid\ListMapper;
+
+final class SubAdmin extends AbstractAdmin
+{
+    protected function configureListFields(ListMapper $list)
+    {
+        $list->addIdentifier('id');
+        $list->add('otherField');
+    }
+}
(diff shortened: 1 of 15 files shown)


franmomu:
🤔 @Devristo I tried to use your example from https://github.com/Devristo/sonata-admin-entity-inheritance, can you please look to see if there is something different? Maybe it's because posgresql.

jordisala1991:
Why dont keep using sqlite?

franmomu:
> Why dont keep using sqlite?

I just wanted to try if the fail was using posgresql (not as a permanent change), it would be interesting to add tests with different platforms to catch these things.

Devristo:
> thinking @Devristo I tried to use your example from https://github.com/Devristo/sonata-admin-entity-inheritance, can you please look to see if there is something different? Maybe it's because posgresql.
> 
> **Update**: Looks like it fails only using posgresql.

Postgresql is indeed more strict about non-aggregated fields in group by than some others. I think more recent MySQL versions are as well: https://dev.mysql.com/doc/refman/5.7/en/group-by-handling.html .

I have no clue about sqlite though.

franmomu:
Needs https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1399

tests/App/Admin/SubAdmin.php
```diff
@@ -0,0 +1,26 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DoctrineORMAdminBundle\Tests\App\Admin;
+
+use Sonata\AdminBundle\Admin\AbstractAdmin;
+use Sonata\AdminBundle\Datagrid\ListMapper;
+
+class SubAdmin extends AbstractAdmin
```
jordisala1991, line 19: missing extends generic type

VincentLanglet (approved):
Does this fail without the fix for #1361 ?

phansys (approved):

franmomu:
> Does this fail without the fix for #1361 ?

It used to fail with the previous configuration I had for `PostgreSQL`, I guess it should also fail now.

jordisala1991 (approved):

jordisala1991:
Thank you @franmomu

PR #1386, opened 2021-04-01
Raise phpstan to max level

VincentLanglet:
Need
- https://github.com/doctrine/persistence/pull/163
- https://github.com/sonata-project/EntityAuditBundle/pull/382
- https://github.com/doctrine/orm/pull/8590

### DIFF
--- a/src/Block/AuditBlockService.php
+++ b/src/Block/AuditBlockService.php
@@ -39,16 +39,21 @@ final class AuditBlockService extends AbstractBlockService
 
     public function execute(BlockContextInterface $blockContext, ?Response $response = null): Response
     {
+        $template = $blockContext->getTemplate();
+        \assert(null !== $template);
+        $limit = $blockContext->getSetting('limit');
+        \assert(\is_int($limit));
+
         $revisions = [];
 
-        foreach ($this->auditReader->findRevisionHistory($blockContext->getSetting('limit'), 0) as $revision) {
+        foreach ($this->auditReader->findRevisionHistory($limit, 0) as $revision) {
             $revisions[] = [
                 'revision' => $revision,
                 'entities' => $this->auditReader->findEntitiesChangedAtRevision($revision->getRev()),
             ];
         }
 
-        return $this->renderResponse($blockContext->getTemplate(), [
+        return $this->renderResponse($template, [
             'block' => $blockContext->getBlock(),
             'settings' => $blockContext->getSettings(),
             'revisions' => $revisions,
--- a/src/Builder/ListBuilder.php
+++ b/src/Builder/ListBuilder.php
@@ -50,6 +50,12 @@ final class ListBuilder implements ListBuilderInterface
     {
         if (null === $type) {
             $guessType = $this->guesser->guess($fieldDescription);
+            if (null === $guessType) {
+                throw new \InvalidArgumentException(sprintf(
+                    'Cannot guess a type for the field description "%s", You MUST provide a type.',
+                    $fieldDescription->getName()
+                ));
+            }
 
             $fieldDescription->setType($guessType->getType());
         } else {
@@ -69,7 +75,16 @@ final class ListBuilder implements ListBuilderInterface
 
     public function fixFieldDescription(FieldDescriptionInterface $fieldDescription): void
     {
-        if (ListMapper::TYPE_ACTIONS === $fieldDescription->getType()) {
+        $type = $fieldDescription->getType();
+        if (!$type) {
+            throw new \RuntimeException(sprintf(
+                'Please define a type for field `%s` in `%s`',
+                $fieldDescription->getName(),
+                \get_class($fieldDescription->getAdmin())
+            ));
+        }
+
+        if (ListMapper::TYPE_ACTIONS === $type) {
             $this->buildActionFieldDescription($fieldDescription);
         }
 
@@ -83,16 +98,8 @@ final class ListBuilder implements ListBuilderInterface
             $fieldDescription->setOption('_sort_order', $fieldDescription->getOption('_sort_order', 'ASC'));
         }
 
-        if (!$fieldDescription->getType()) {
-            throw new \RuntimeException(sprintf(
-                'Please define a type for field `%s` in `%s`',
-                $fieldDescription->getName(),
-                \get_class($fieldDescription->getAdmin())
-            ));
-        }
-
         if (!$fieldDescription->getTemplate()) {
-            $fieldDescription->setTemplate($this->getTemplate($fieldDescription->getType()));
+            $fieldDescription->setTemplate($this->getTemplate($type));
 
             if (!$fieldDescription->getTemplate()) {
                 switch ($fieldDescription->getMappingType()) {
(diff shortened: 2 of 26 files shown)


src/Builder/ListBuilder.php
```diff
@@ -50,6 +50,12 @@ public function buildField(?string $type, FieldDescriptionInterface $fieldDescri
     {
         if (null === $type) {
             $guessType = $this->guesser->guess($fieldDescription);
+            if (null === $guessType) {
```
jordisala1991, line 53: If every time we are returning null we throw an exception, wouldnt it be better to only return non null values or throw exception inside the guesser?

src/Block/AuditBlockService.php
```diff
@@ -39,16 +39,21 @@ public function __construct(Environment $twig, AuditReader $auditReader)
 
     public function execute(BlockContextInterface $blockContext, ?Response $response = null): Response
     {
+        $template = $blockContext->getTemplate();
+        \assert(null !== $template);
+        $limit = $blockContext->getSetting('limit');
```
jordisala1991, line 44: maybe cast is better here?

src/Block/AuditBlockService.php
```diff
@@ -39,16 +39,21 @@ public function __construct(Environment $twig, AuditReader $auditReader)
 
     public function execute(BlockContextInterface $blockContext, ?Response $response = null): Response
     {
+        $template = $blockContext->getTemplate();
```
jordisala1991, line 42: can a block return null here?

src/Builder/ListBuilder.php
```diff
@@ -83,7 +90,8 @@ public function fixFieldDescription(FieldDescriptionInterface $fieldDescription)
             $fieldDescription->setOption('_sort_order', $fieldDescription->getOption('_sort_order', 'ASC'));
         }
 
-        if (!$fieldDescription->getType()) {
+        $type = $fieldDescription->getType();
```
jordisala1991, line 93: the variable was initialized previously in this method

src/Builder/ListBuilder.php
```diff
@@ -83,7 +90,8 @@ public function fixFieldDescription(FieldDescriptionInterface $fieldDescription)
...
+        if (!$type) {
```
jordisala1991, line 94: maybe move this part at the top of the method?

src/Block/AuditBlockService.php
VincentLanglet: I assume that no because of 
```
public function configureSettings(OptionsResolver $resolver): void
    {
        $resolver->setDefaults([
            'limit' => 10,
            'template' => '@SonataDoctrineORMAdmin/Block/block_audit.html.twig',
        ]);
    }
```

src/Block/AuditBlockService.php
VincentLanglet: Since there is
```
public function configureSettings(OptionsResolver $resolver): void
    {
        $resolver->setDefaults([
            'limit' => 10,
            'template' => '@SonataDoctrineORMAdmin/Block/block_audit.html.twig',
        ]);
    }
``` 

What do you think about using `allowedTypes` ?

src/Builder/ListBuilder.php
VincentLanglet: I made https://github.com/sonata-project/SonataAdminBundle/pull/6987 because the TypeGuesserChain could return null...

If you have a TypeGuesserChain with no guesser how can you return a type ?
And the Guesser interface from Symfony was allowing `null` as return type...

The Guesser from this bundle never return null, but just in case, I added the exception...

src/Block/AuditBlockService.php
jordisala1991: Is that a feature from the options resolver? I think it would be great but not sure if phpstan will understand it tho.

VincentLanglet: Phpstan won't understand, so I'll keep the `assert`, but it still a code improvement

src/Datagrid/Pager.php
```diff
@@ -68,8 +63,13 @@ public function init(): void
     {
         $this->resultsCount = $this->computeResultsCount();
 
-        $this->getQuery()->setFirstResult(null);
-        $this->getQuery()->setMaxResults(null);
+        $query = $this->getQuery();
+        if (null === $query) {
```
jordisala1991, line 67: Probably out of scope of this Pr, but wouldnt be nice if we ensure somehow that the query is never null, maybe as a parameter of this method or as a parameter of construct?

src/DependencyInjection/SonataDoctrineORMAdminExtension.php
```diff
@@ -38,6 +38,7 @@ public function load(array $configs, ContainerBuilder $container): void
         $loader->load('doctrine_orm_filter_types.xml');
 
         $bundles = $container->getParameter('kernel.bundles');
+        \assert(\is_array($bundles));
```
jordisala1991, line 41: seems strange having to add this around each getParameter to ensure type

tests/Functional/CRUDTest.php
```diff
@@ -79,7 +80,9 @@ public function testEdit(): void
 
     public function testDelete(): void
     {
-        $entityManager = static::bootKernel()->getContainer()->get('doctrine')->getManager();
+        $doctrine = static::bootKernel()->getContainer()->get('doctrine');
```
jordisala1991, line 83: I think the container resolution can be solved warming up cache before running phpstan. (can be done executing test previously). Then you add the xml to the phpstan config and it can solve those issues

src/Datagrid/Pager.php
VincentLanglet: Currently the pager and the query are constructor param of the Datagrid.
In this one we're running
```
$this->pager->setQuery($this->query);
$this->pager->init();
```

Maybe we could remove setQuery and use the query as a param of init indeed.

src/DependencyInjection/SonataDoctrineORMAdminExtension.php
VincentLanglet: getParameter is considered as returning `mixed` for phpstan because of the symfony phpdoc

VincentLanglet:
The lowest build is throwing 
```
Class 'Doctrine\Common\Persistence\Mapping\Driver\MappingDriverChain' not found
```

I think it's related to the bump you did on SonataAdmin @jordisala1991

VincentLanglet:
This is ready @sonata-project/contributors :)

jordisala1991 (approved):

src/Filter/ChoiceFilter.php
```diff
@@ -54,9 +54,9 @@ public function getRenderSettings(): array
     /**
      * @param mixed[] $data
      *
-     * @phpstan-param array{type?: string|int|null, value?: mixed} $data
+     * @phpstan-param array{type: string|int|null, value: mixed} $data
```
franmomu, line 57: Can `type` be a string? I'm assuming only `int` in https://github.com/sonata-project/SonataAdminBundle/pull/7022

src/Filter/ChoiceFilter.php
```diff
@@ -88,9 +88,9 @@ private function filterWithMultipleValues(ProxyQueryInterface $query, string $al
...
```
franmomu, line 91: Same thing

src/Filter/ChoiceFilter.php
VincentLanglet: Yes I need to update it

jordisala1991 (approved):

PR #1391, opened 2021-04-04
Test on different platforms

franmomu:
This is just copy-pasting the job we have and execute it for `MySQL` and `Postgres`. I created another file so I guess there is no need to modify `dev-kit`. 

We could maybe test just with one variant and do not test lower versions.

Ref: https://github.com/sonata-project/dev-kit/issues/1477

### DIFF
--- a/.github/workflows/test-platforms.yaml
+++ b/.github/workflows/test-platforms.yaml
@@ -54,7 +54,7 @@ jobs:
               run: make test
 
     test-postgres:
-        name: PHP 8.0 + highest + PostgreSQL
+        name: PHP 8.0 + PostgreSQL + highest
 
         runs-on: ubuntu-latest
 
--- a/.github/workflows/test-platforms.yaml
+++ b/.github/workflows/test-platforms.yaml
@@ -1,0 +1,123 @@
+name: Test on platforms
+
+on:
+    push:
+        branches:
+            - 3.x
+            - master
+    pull_request:
+
+jobs:
+    test-mysql:
+        name: PHP ${{ matrix.php-version }} + MySQL + ${{ matrix.dependencies }} + ${{ matrix.variant }}
+
+        runs-on: ubuntu-latest
+
+        continue-on-error: ${{ matrix.allowed-to-fail }}
+
+        env:
+            DATABASE_URL: "mysql://root@127.0.0.1:3306/sonata_tests"
+
+        strategy:
+            matrix:
+                php-version:
+                    - '8.0'
+                dependencies: [highest]
+                allowed-to-fail: [false]
+                variant: [normal]
+
+        services:
+            mysql:
+                image: "mysql:8.0"
+
+                options: >-
+                    --health-cmd "mysqladmin ping --silent"
+                    -e MYSQL_ALLOW_EMPTY_PASSWORD=yes
+                    -e MYSQL_DATABASE=sonata_tests
+                ports:
+                    - "3306:3306"
+
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v2
+
+            - name: Install PHP with extensions
+              uses: shivammathur/setup-php@v2
+              with:
+                  php-version: ${{ matrix.php-version }}
+                  coverage: pcov
+                  tools: composer:v2
+                  extensions: zip, mysqli, pdo_mysql
+
+            - name: Add PHPUnit matcher
+              run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
+
+            - name: Install variant
+              if: matrix.variant != 'normal'
+              run: composer require ${{ matrix.variant }} --no-update
+
+            - name: "Install Composer dependencies (${{ matrix.dependencies }})"
+              uses: "ramsey/composer-install@v1"
+              with:
+                  dependency-versions: "${{ matrix.dependencies }}"
+                  composer-options: "--prefer-dist --prefer-stable"
+
+            - name: Run Tests
+              run: make test
+
+    test-postgres:
+        name: PHP ${{ matrix.php-version }} + ${{ matrix.dependencies }} + Postgres + ${{ matrix.variant }}
+
+        runs-on: ubuntu-latest
+
+        continue-on-error: ${{ matrix.allowed-to-fail }}
+
+        env:
+            DATABASE_URL: "postgresql://postgres:postgres@127.0.0.1:5432/postgres?serverVersion=13&charset=utf8"
+
+        strategy:
+            matrix:
+                php-version:
+                    - '8.0'
+                dependencies: [highest]
+                allowed-to-fail: [false]
+                variant: [normal]
+
+        services:
+            postgres:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-platforms.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+on:
+    push:
+        branches:
+            - 3.x
+            - master
+    pull_request:
+
+env:
+    ACTIONS_ALLOW_UNSECURE_COMMANDS: true
```
OskarStark, line 11: Still needed?

franmomu: 👍 I guess not (I don't know why was needed though).

VincentLanglet: Can it be removed from dev kit then ? https://github.com/sonata-project/dev-kit/blob/master/templates/project/.github/workflows/test.yaml.twig#L15

VincentLanglet:
> I created another file so I guess there is no need to modify `dev-kit`.

If we update the php version on dev-kit or change the variant (if you use them), we'll need to change this file too.

Does testing on different platform can be used on some others project ?
For instance, since `sonata-media-bundle` use this bundle https://github.com/sonata-project/SonataMediaBundle/blob/3.x/composer.json#L97 ; I assume it could be tested on different platform too ?
If yes, a config `platform: ['mysql', 'postgres']` could be used on dev-kit.

If no, we can consider this is enough like this.

.github/workflows/test-platforms.yaml
```diff
@@ -0,0 +1,123 @@
...
+        env:
+            DATABASE_URL: "mysql://root@127.0.0.1:3306/sonata_tests"
+
+        strategy:
+            matrix:
+                php-version:
+                    - '8.0'
+                dependencies: [highest]
+                allowed-to-fail: [false]
+                variant: [normal]
```
VincentLanglet, line 27: Since you're not using variant, you could remove the key from the matrix

.github/workflows/test-platforms.yaml
```diff
@@ -0,0 +1,123 @@
...
+                  coverage: pcov
+                  tools: composer:v2
+                  extensions: zip, mysqli, pdo_mysql
+
+            - name: Add PHPUnit matcher
+              run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
+
+            - name: Install variant
+              if: matrix.variant != 'normal'
+              run: composer require ${{ matrix.variant }} --no-update
```
VincentLanglet, line 57:
```suggestion
```

.github/workflows/test-platforms.yaml
```diff
@@ -0,0 +1,123 @@
...
+on:
+    push:
+        branches:
+            - 3.x
+            - master
+    pull_request:
+
+jobs:
+    test-mysql:
+        name: PHP ${{ matrix.php-version }} + MySQL + ${{ matrix.dependencies }} + ${{ matrix.variant }}
```
VincentLanglet, line 12:
```suggestion
        name: PHP ${{ matrix.php-version }} + MySQL + ${{ matrix.dependencies }}
```

.github/workflows/test-platforms.yaml
```diff
@@ -0,0 +1,123 @@
...
+        continue-on-error: ${{ matrix.allowed-to-fail }}
+
+        env:
+            DATABASE_URL: "mysql://root@127.0.0.1:3306/sonata_tests"
+
+        strategy:
+            matrix:
+                php-version:
+                    - '8.0'
+                dependencies: [highest]
```
VincentLanglet, line 25: Since you're only testing highest dependencies we could remove the key

.github/workflows/test-platforms.yaml
```diff
@@ -0,0 +1,123 @@
...
+                allowed-to-fail: [false]
```
VincentLanglet, line 26: Same

.github/workflows/test-platforms.yaml
```diff
@@ -0,0 +1,123 @@
...
+            - master
+    pull_request:
+
+jobs:
+    test-mysql:
+        name: PHP ${{ matrix.php-version }} + MySQL + ${{ matrix.dependencies }} + ${{ matrix.variant }}
+
+        runs-on: ubuntu-latest
+
+        continue-on-error: ${{ matrix.allowed-to-fail }}
```
VincentLanglet, line 16:
```suggestion
        continue-on-error: false
```

.github/workflows/test-platforms.yaml
OskarStark: We need it if we set output

You can create a proof PR on admin bundle to check

franmomu:
> > I created another file so I guess there is no need to modify `dev-kit`.
> 
> If we update the php version on dev-kit or change the variant (if you use them), we'll need to change this file too.

Yep, what I meant was that we don't need to modify dev-kit right now and by creating this new file there is no need to ignore `test.yaml`.

> Does testing on different platform can be used on some others project ?
> For instance, since `sonata-media-bundle` use this bundle https://github.com/sonata-project/SonataMediaBundle/blob/3.x/composer.json#L97 ; I assume it could be tested on different platform too ?
> If yes, a config `platform: ['mysql', 'postgres']` could be used on dev-kit.

We could also add `mongodb` and have configured the `PhpExtension` needed on each platform, the problem I see is how to configure the `env` vars, like `DATABASE_URL`. 

Anyway that would require some work and I think this is easier right now.

.github/workflows/test-platforms.yaml
franmomu: I left them on purpose so if we change something on `test.yaml`, it would be easier to adapt, I can inline them.

VincentLanglet (dismissed):

.github/workflows/test-platforms.yaml
```diff
@@ -0,0 +1,99 @@
...
+              uses: "ramsey/composer-install@v1"
+              with:
+                  dependency-versions: "highest"
+                  composer-options: "--prefer-dist --prefer-stable"
+
+            - name: Run Tests
+              run: make test
+
+    test-postgres:
+        name: PHP 8.0 + highest + Postgres
```
phansys, line 57:
```suggestion
        name: PHP 8.0 + highest + PostgreSQL
```

phansys (approved):

.github/workflows/test-platforms.yaml
```diff
@@ -0,0 +1,99 @@
...
+              uses: "ramsey/composer-install@v1"
+              with:
+                  dependency-versions: "highest"
+                  composer-options: "--prefer-dist --prefer-stable"
+
+            - name: Run Tests
+              run: make test
+
+    test-postgres:
+        name: PHP 8.0 + highest + PostgreSQL
```
phansys, line 57: Just in order to be consistent with the MySQL variant.
```suggestion
        name: PHP 8.0 + PostgreSQL + highest
```

PR #1395, opened 2021-04-06
Use "BINARY()" DQL function for case sensitive MySQL pattern matching in `StringFilter::filter()`

phansys:
## Subject
Use "BINARY()" DQL function for case sensitive MySQL pattern matching in `StringFilter::filter()`.

I am targeting this branch, because these changes respect BC.

Closes #1387.

## Changelog

```markdown
### Added
- Call to `BINARY()` (`DoctrineExtensions\Query\Mysql\Binary`) DQL function for MySQL pattern matching when it is available and "case_sensitive" option is set to `true` in `StringFilter::filter()`.

### Changed
- Default value for the "case_sensitive" option from `true` to `null` in `StringFilter`.

### Fixed
- Respecting "case_sensitive" when it is set to `true` along MySQL columns using other collations than "utf8mb4_0900_as_cs" or "utf8mb4_bin".

### Deprecated
- Omitting or not passing an instance of `Doctrine\ORM\EntityManagerInterface` as argument 1 to `StringFilter::__construct()`.
```

## To do

- [x] Update the tests;
- [x] Update the documentation;
- [x] Add an upgrade note;
- [x] Determine if we must deprecate something;
- [x] Check if we can / should configure the DQL function automatically.

Issue #1387, opened 2021-04-01
Filter option "case_sensitive" is not being respected with MySQL


phansys:
#### Sonata packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'sonata-project/*'
sonata-project/admin-bundle              3.95.0 3.95.0 The missing Symfony Admin Generator
sonata-project/block-bundle              3.22.0 4.5.3  Symfony SonataBlockBundle
sonata-project/cache                     2.0.1  2.1.1  Cache library
sonata-project/datagrid-bundle           3.3.0  3.3.0  Symfony SonataDatagridBundle
sonata-project/doctrine-extensions       1.11.0 1.12.0 Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.32.0 3.32.0 Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/entity-audit-bundle       1.1.0  1.2.0  Audit for Doctrine Entities
sonata-project/exporter                  2.5.0  2.6.1  Lightweight Exporter library
sonata-project/form-extensions           1.8.1  1.9.0  Symfony form extensions
sonata-project/intl-bundle               2.10.0 2.10.1 Symfony SonataIntlBundle
sonata-project/media-bundle              3.29.0 3.31.0 Symfony SonataMediaBundle
sonata-project/twig-extensions           1.5.0  1.5.1  Sonata twig extensions
sonata-project/user-bundle               4.10.1 4.11.1 Symfony SonataUserBundle
```

</p>
</details>

#### Symfony packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'symfony/*'
symfony/amazon-sqs-messenger       v5.2.2  v5.2.4  Symfony Amazon SQS extension Messenger Bridge
symfony/amqp-messenger             v5.2.2  v5.2.4  Symfony AMQP extension Messenger Bridge
symfony/asset                      v4.4.19 v5.2.4  Manages URL generation and versioning of web assets such as CSS stylesheets, JavaScript files and image files
symfony/browser-kit                v5.2.2  v5.2.4  Simulates the behavior of a web browser, allowing you to make requests, click on links and submit forms programmatically
symfony/cache                      v5.2.2  v5.2.6  Provides an extended PSR-6, PSR-16 (and tags) implementation
symfony/cache-contracts            v2.2.0  v2.2.0  Generic abstractions related to caching
[...]

PR #1402, opened 2021-04-08
[PoC] Case sensitive tests

phansys:
See https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1395#issuecomment-815827140.

### DIFF
--- a/tests/Functional/DatagridTest.php
+++ b/tests/Functional/DatagridTest.php
@@ -30,4 +30,32 @@ final class DatagridTest extends BasePantherTestCase
 
         self::assertSelectorTextContains('.sonata-link-identifier', 'Dystopian');
     }
+
+    public function testFilterRespectingCaseSensitivity(): void
+    {
+        $this->client->request(Request::METHOD_GET, '/admin/tests/app/author/list');
+
+        $this->client->clickLink('Filters');
+        $this->client->clickLink('Address Street');
+
+        $this->client->submitForm('Filter', [
+            'filter[address__street][value]' => 'Mancha',
+        ]);
+
+        self::assertSelectorTextContains('.sonata-link-identifier', 'Miguel de Cervantes');
+    }
+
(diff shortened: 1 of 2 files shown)


tests/Functional/DatagridTest.php
```diff
@@ -30,4 +30,32 @@ public function testFilter(): void
...
+        $this->client->request(Request::METHOD_GET, '/admin/tests/app/author/list');
+
+        $this->client->clickLink('Filters');
+        $this->client->clickLink('Address Street');
+
+        $this->client->submitForm('Filter', [
+            'filter[address__street][value]' => 'mancha',
+        ]);
+
+        self::assertSelectorTextNotContains('.sonata-link-identifier', 'Miguel de Cervantes');
```
phansys, line 59: Seems that we aren't getting case sensitive results with SQLite: https://www.sqlite.org/pragma.html#pragma_case_sensitive_like.

phansys:
After some tests, I'm thinking if the required effort worths about the case sensitivity.
By instance, in order to have a concrete idea if the comparisons are case sensitive or not in MySQL, we should detect if the compared columns are using the collation `utf8mb4_0900_as_cs` or `utf8mb4_bin`, otherwise (with the current implementation) we have to detect if the `DoctrineExtensions\Query\Mysql\Binary` (`BINARY()`) DQL function is registered under the "binary" name, to allow the comparison to use that function.
A similar case is present here in SQLite, where the `PRAGMA case_sensitive_like` directive controls the behavior.
I haven't more evidence about other RDMSs, but I guess those examples are enough to bring a clear idea about the problem.
Could we re-evaluate this suggestion? https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1387#issuecomment-812917465

VincentLanglet:
Should we close this @phansys ?

phansys:
Sure, let's close.

PR #1408, opened 2021-04-11
#1405 Fixed FieldDescriptionFactory check for instance of EntityManag…

### LINKED ISSUE #1405: No entity manager defined for class with Doctrine Decorator : Issue and Fix proposal ;-)
In our project, we use a class to decorate the default doctrine.orm.default_entity_manager.

```
services:
  doctrine.orm.entity_manager:
    class: App\Core\Doctrine\ORM\EntityManager
    autowire: true
    decorates: 'doctrine.orm.default_entity_manager'
```

in this case the method getEntityManager in FieldDescriptionFactory throw an exception because $this->registry->getManagerForClass($class); returns an instance of my class (App\Core\Doctrine\ORM\EntityManager).

it is easily fixable by replacing the "getEntityManager" function in FieldDescriptionFactory by the following

```
private function getEntityManager(string $class): EntityManagerInterface
    {
        $em = $this->registry->getManagerForClass($class);

        if (!$em instanceof EntityManagerInterface) {
            throw new \UnexpectedValueException(sprintf('No entity manager defined for class "%s".', $class));
        }

        return $em;
    }
```

michkinn:
…erInterface to allow use of entityManagerDecorator

## Subject

I am targeting this branch, because Sonata generate an Exception when entityManager use a doctrine decorator.

Closes #1405 

## Changelog

```markdown
### Fixed
- Allow to decorate EntityManager
```

### DIFF
--- a/src/FieldDescription/FieldDescriptionFactory.php
+++ b/src/FieldDescription/FieldDescriptionFactory.php
@@ -13,7 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\DoctrineORMAdminBundle\FieldDescription;
 
-use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Sonata\AdminBundle\FieldDescription\FieldDescriptionFactoryInterface;
 use Sonata\AdminBundle\FieldDescription\FieldDescriptionInterface;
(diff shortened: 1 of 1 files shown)


franmomu:
Closing to restart tests.

franmomu (approved):
I see this change fine.

But I don't really know how this would fix #1405, `$em = $this->registry->getManagerForClass($class);` would return the same instance and if you decorate the service I guess you are also implementing `EntityManagerInterface`.

VincentLanglet (approved):

VincentLanglet:
Thanks @michkinn

michkinn:
@franmomu, You' re right. my decorator had to implements EntityManagerInterface. But that's the only way to make it work.

PR #1409, opened 2021-04-11
Fix Revision instantiation

franmomu:
It fails because of https://github.com/sonata-project/EntityAuditBundle/pull/382, this PR fixes the fail, but I guess we introduced a BC break adding those type declarations.


VincentLanglet (approved):

VincentLanglet:
Since the code of entityAuditBundle was using the property as a datetime, i would say that a real usage of the revision entity would already lead to an error.

I would see this like a faster failure rather than a real bc break...

PR #1411, opened 2021-04-11
Allow sonata 4

VincentLanglet:


composer.json
```diff
@@ -29,7 +29,7 @@
         "doctrine/doctrine-bundle": "^2.3",
         "doctrine/orm": "^2.5",
         "doctrine/persistence": "^2.0",
-        "sonata-project/admin-bundle": "4.x-dev",
+        "sonata-project/admin-bundle": "4.x-dev || ^4.0",
```
jordisala1991, line 32: I think we can leave just ^4.0

VincentLanglet: It during the beta, it will allow us to fix the BC-break on this bundle even if it's not release as a new alpha.

dmaicher: yeah I would also say so

dmaicher (approved):

PR #1412, opened 2021-04-13
Do not restrict to alpha-1

jordisala1991:
## Subject

I am targeting this branch, because this is the one.

## Changelog

```markdown
### Added
- Add compatiblity with all SonataAdmin 4.0 versions
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -91,6 +91,7 @@
             "Sonata\\DoctrineORMAdminBundle\\Tests\\": "tests/"
         }
     },
+    "minimum-stability": "alpha",
     "scripts": {
         "post-install-cmd": [
             "[ $COMPOSER_DEV_MODE -eq 0 ] || vendor/bin/simple-phpunit install"
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "doctrine/doctrine-bundle": "^2.3",
         "doctrine/orm": "^2.5",
         "doctrine/persistence": "^2.0",
-        "sonata-project/admin-bundle": "4.0.0-alpha-1",
+        "sonata-project/admin-bundle": "^4.0",
         "sonata-project/exporter": "^2.0",
         "sonata-project/form-extensions": "^1.4",
         "symfony/config": "^4.4 || ^5.2",


composer.json
```diff
@@ -91,6 +91,7 @@
             "Sonata\\DoctrineORMAdminBundle\\Tests\\": "tests/"
         }
     },
+    "minimum-stability": "alpha",
```
jordisala1991, line 94: We can modify or remove this when we release betas, rcs, and finally the sable version.

jordisala1991:
The only thing to adjust when we break bc with alpha 1 is the conflict section, to avoid installing older alpha version.

composer.json
```diff
@@ -47,6 +47,7 @@
     },
     "conflict": {
         "simplethings/entity-audit-bundle": ">=2.0",
+        "sonata-project/admin-bundle": "4.0.0-alpha",
```
jordisala1991, line 50: this packagist version needs to be removed @rande

core23: Not sure if we need this. An alpha version is always risky

jordisala1991: Yes, but that was tagged accidentaly, the correct one is 4.0.0-alpha-1

core23: I know, but we should remove the conflict once the tagged release was removed from packagist.

jordisala1991: Agree, once the tagged release is removed, I will change this file again.

franmomu (approved):
After this, I'll do the same in https://github.com/sonata-project/SonataTranslationBundle/pull/477

composer.json
VincentLanglet: The conflict is not needed.
The release `4.0.0-alpha` has the same code than the `4.0.0-alpha-1` one.

VincentLanglet (changes requested):

composer.json
jordisala1991: Removed

VincentLanglet (approved):

composer.json
```diff
@@ -29,7 +29,7 @@
         "doctrine/doctrine-bundle": "^2.3",
         "doctrine/orm": "^2.5",
         "doctrine/persistence": "^2.0",
-        "sonata-project/admin-bundle": "4.0.0-alpha-1",
+        "sonata-project/admin-bundle": "^4.0",
```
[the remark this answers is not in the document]
VincentLanglet: Also we should keep in mind that the 4.0-alpha-2 won't work with every futur alpha.
So maybe we should add a conflict for the futur alpha ?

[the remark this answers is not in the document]
jordisala1991: I dont understand, please take a look at the build, there is already one checking against dev master and it is working just fine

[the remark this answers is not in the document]
VincentLanglet: This is my issue:
Let's say we're merging https://github.com/sonata-project/SonataAdminBundle/pull/7022, it introduce a BC-break and a new feature. How can we add compatibility in the persistence bundle ?
- If we only use `^4.0`, we cannot use the last commit of SonataAdmin.
- If we release an alpha-2 from SonataAdmin, during the time we add compatibility, people could load the alpha2 from SonataAdmin with the alpha1 from this bundle.

So how do we do ?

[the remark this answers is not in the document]
VincentLanglet: I'm not sure this PR is fixing the right thing.
We shouldn't allow more alpha than the alpha 1 since we cannot be sure it will work for others alpha.

If I understand correctly we want to fix the build 
```
Test / PHP 8.0 + highest + sonata-project/admin-bundle:"dev-master" (pull_request)
```
from another bundle.

Maybe the only change to do is something like `4.0.0-alpha-1 || dev-master` ?
Or maybe it's just an issue from the other bundle, which should run 
```
Test / PHP 8.0 + highest + sonata-project/admin-bundle:"4.0.0-alpha-1" (pull_request)
```
instead of dev-master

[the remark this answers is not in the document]
jordisala1991: > This is my issue:
> 
> Let's say we're merging https://github.com/sonata-project/SonataAdminBundle/pull/7022, it introduce a BC-break and a new feature. How can we add compatibility in the persistence bundle ?
> 
> - If we only use `^4.0`, we cannot use the last commit of SonataAdmin.
> 
> - If we release an alpha-2 from SonataAdmin, during the time we add compatibility, people could load the alpha2 from SonataAdmin with the alpha1 from this bundle.
> 
> 
> 
> So how do we do ?

If you merge a pr that breaks bc, this master branch will start to fail on the dev-master build. Then when you want to fix it, you should add a conflict with alpha1 and only let install dev-master (through minimum stability dev)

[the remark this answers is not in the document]
VincentLanglet: Ok, so we're gonna update often this composer json.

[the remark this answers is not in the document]
jordisala1991: I think so

VincentLanglet (approved):

core23 (approved):

PR #1416, opened 2021-04-14
Introduce "force_case_insensitivity" option and deprecate "case_sensitive" in `StringFilter`

### LINKED ISSUE #1415: Introduce "force_case_insensitivity" option and deprecate "case_sensitive"
## RFC

Recap from https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1402#issuecomment-817048349 and https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1387#issuecomment-812917465.

I think changing the name of this option to `force_case_insensitivity` (deprecating `case_sensitive`) to make clearer its real behavior will be simpler than adding more specific checks for each RDBMS and its possible scenarios (column collations, extensions availability, etc).
I think it reflects its purpose in a better way.
With this approach I think we could keep the current logic for all the RDBMSs with no extra checks (just inversing the check against this new option), using `true` when we want to force the matching to ignore the case sensitivity, and leaving it unset or using `false` when we want the RDBMS to use its default behavior.

By instance, in order to have a concrete idea if the comparisons are case sensitive or not in MySQL, currently we should detect if the compared columns are using the collation utf8mb4_0900_as_cs or utf8mb4_bin, otherwise we have to detect if the `DoctrineExtensions\Query\Mysql\Binary` (`BINARY()`) DQL function is registered under the "binary" name, to allow the comparison to use that function.
A similar case is present here in SQLite, where the `PRAGMA case_sensitive_like` directive controls the behavior.
I haven't more evidence about other RDMSs, but I guess those examples are enough to bring a clear idea about the problem.

phansys:
## Subject
Introduce "force_case_insensitivity" option and deprecate "case_sensitive" in `StringFilter`.

I am targeting this branch, because these changes respect BC.

Closes #1415.
Reverts #1395.

## Changelog

```markdown
### Added
- "force_case_insensitivity" option to `StringFilter` in order to force the database to ignore the case sensitivity when matching filters.

### Deprecated
- "case_sensitive" option in `StringFilter`.
```

### DIFF
--- a/src/Filter/StringFilter.php
+++ b/src/Filter/StringFilter.php
@@ -87,13 +87,30 @@ class StringFilter extends Filter
         // c.name > '1' => c.name OPERATOR :FIELDNAME
         $parameterName = $this->getNewParameterName($query);
 
-        $or = $query->getQueryBuilder()->expr()->orX();
+        $forceCaseInsensitivity = true === $this->getOption('force_case_insensitivity', false);
 
-        if ($this->getOption('case_sensitive')) {
-            $or->add(sprintf('%s.%s %s :%s', $alias, $field, $operator, $parameterName));
-        } else {
-            $or->add(sprintf('LOWER(%s.%s) %s :%s', $alias, $field, $operator, $parameterName));
+        // NEXT_MAJOR: Remove the following condition and its body.
+        if (null !== $this->getOption('case_sensitive')) {
+            @trigger_error(
+                'Option "case_sensitive" is deprecated since sonata-project/doctrine-orm-admin-bundle 3.x'
+                .' and will be removed in version 4.x. Use the "force_case_insensitivity" option instead.',
+                \E_USER_DEPRECATED
+            );
+
+            if (null === $this->getOption('force_case_insensitivity')) {
+                $forceCaseInsensitivity = false === $this->getOption('case_sensitive', true);
+            }
         }
+
+        if ($forceCaseInsensitivity && '' !== $data['value']) {
+            $clause = 'LOWER(%s.%s) %s :%s';
+        } else {
+            $clause = '%s.%s %s :%s';
+        }
+
+        $or = $query->getQueryBuilder()->expr()->orX(
+            sprintf($clause, $alias, $field, $operator, $parameterName)
+        );
 
         if (StringOperatorType::TYPE_NOT_CONTAINS === $type || StringOperatorType::TYPE_NOT_EQUAL === $type) {
             $or->add($query->getQueryBuilder()->expr()->isNull(sprintf('%s.%s', $alias, $field)));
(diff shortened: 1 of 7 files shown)


src/Filter/StringFilter.php
```diff
@@ -165,17 +146,19 @@ public function filter(BaseProxyQueryInterface $query, $alias, $field, $data)
...
 
     public function getDefaultOptions()
     {
         return [
-            // NEXT_MAJOR: Remove the format option.
+            // NEXT_MAJOR: Remove the "format" option.
             'format' => '%%%s%%',
+            // NEXT_MAJOR: Remove the "case_sensitive" option.
             'case_sensitive' => null,
+            'force_case_insensitivity' => null,
```
VincentLanglet, line 161: Why null ?
What's the difference between null and false ?

If I understand the option, the only allowed value would be `false` and `true`

phansys: We haven't other way to detect if this option is set or not. Currently, I'm using the value from "case_sensitive" if this option is not set.
I think we have 2 choices: add a `hasOption()` method, or leave this value adding a `NEXT_MAJOR` comment.
WDYT?

VincentLanglet: Why do you need to know if the option is set ?

If `case_sensitive` is not null, use this (deprecated) value
If `case_sensitive` is null, use force_case_insensitivity, which is by default to false.

phansys: Because I'd like the new option to be respected if it is set, regardless the value of the deprecated option.

VincentLanglet: Then we could add a NEXT_MAJOR comment to change the default value to `false` and remove the `true ===` or `true !==` checks. But It feel weird to provide a `null` value and a `false` value which has the same behavior. And introducing a `null` default value which is already deprecated and will be changed to false in the next major...

But when I read `force_case_insensitivity => false, case_sensitive => false`:
I would expect
- `force_case_insensitivity` won't force the case insensitivity so it will do nothing. (It never was said that `false` will force the case sensitivity)
- `case_sensitive` will force the case insensitivity because it's already the behavior of this option

phansys: >But when I read force_case_insensitivity => false, case_sensitive => false:
I would expect
>
>force_case_insensitivity won't force the case insensitivity so it will do nothing. (It never was said that false will force the case sensitivity)
case_sensitive will force the case insensitivity because it's already the behavior of this option

I think the idea transmitted by the name `case_sensitive` leads to wrong assumptions. In fact, that's the reason for this change.
That's why I feel like a more natural behavior to completely ignore the option if its counterpart (`force_case_insensitivity`) is set.

phansys: The `NEXT_MAJOR` comment about the default value was added.

VincentLanglet (approved):

src/Filter/StringFilter.php
phansys: @VincentLanglet, I've added a comment in the upgrade document to make this behavior explicit.

VincentLanglet (approved):

jordisala1991 (approved):

PR #1427, opened 2021-04-27
Stop using SonataAdmin config in doctrineORMAdmin

VincentLanglet:
## Subject

## Changelog

```markdown
### Deprecated
- `templates.form` and `templates.filter` config

### Fixed
- Always merge `SonataDoctrineORMAdmin` form and filter templates.
- Using `sonata_admin` configuration.
```

### DIFF
--- a/src/DependencyInjection/Compiler/AddTemplatesCompilerPass.php
+++ b/src/DependencyInjection/Compiler/AddTemplatesCompilerPass.php
@@ -20,7 +20,7 @@ use Symfony\Component\DependencyInjection\Definition;
 /**
  * NEXT_MAJOR: Remove the "since" part of the internal annotation.
  *
- * @internal since sonata-project/admin-bundle version 4.0
+ * @internal since sonata-project/doctrine-orm-admin-bundle version 4.0
  *
  * @final since sonata-project/doctrine-orm-admin-bundle 3.24
  *
--- a/src/DependencyInjection/Compiler/AddTemplatesCompilerPass.php
+++ b/src/DependencyInjection/Compiler/AddTemplatesCompilerPass.php
@@ -18,6 +18,10 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
 
 /**
+ * NEXT_MAJOR: Remove the "since" part of the internal annotation.
+ *
+ * @internal since sonata-project/admin-bundle version 4.0
+ *
  * @final since sonata-project/doctrine-orm-admin-bundle 3.24
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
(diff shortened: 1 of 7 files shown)


VincentLanglet:
I need to wait for https://github.com/sonata-project/SonataAdminBundle/pull/7127

And, @dmaicher, I need to this AFTER the `AddDependencyCallsCompilerPass.php` but before the `AdminAddInitializeCallCompilerPass` ; How do I do this ?

dmaicher:
> And, @dmaicher, I need to this AFTER the `AddDependencyCallsCompilerPass.php` but before the `AdminAddInitializeCallCompilerPass` ; How do I do this ?

Then we should give https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/SonataDoctrineORMAdminBundle.php#L30 a lower priority than 0 so it comes after https://github.com/sonata-project/SonataAdminBundle/blob/3.x/src/SonataAdminBundle.php#L55.

Like 

```php
$container->addCompilerPass(new AddTemplatesCompilerPass(), PassConfig::TYPE_BEFORE_OPTIMIZATION, -1);

**Comment by VincentLanglet** [comment]
Need SonataAdmin 3.98.2

File: src/DependencyInjection/Configuration.php [hunk 1]
```diff
@@ -56,11 +56,15 @@ public function getConfigTreeBuilder()
...
                     ->children()
+                        // NEXT_MAJOR: Remove this option.
                         ->arrayNode('form')
+                            ->setDeprecated('The "%node%" option is deprecated since sonata-project/admin-bundle 3.x.')
                             ->prototype('scalar')->end()
                             ->defaultValue(['@SonataDoctrineORMAdmin/Form/form_admin_fields.html.twig'])
                         ->end()
+                        // NEXT_MAJOR: Remove this option.
                         ->arrayNode('filter')
+                            ->setDeprecated('The "%node%" option is deprecated since sonata-project/admin-bundle 3.x.')
```
jordisala1991, line 67: package name should be changed to the orm bundle, same above

src/DependencyInjection/Compiler/AddTemplatesCompilerPass.php
```diff
@@ -26,7 +30,7 @@ class AddTemplatesCompilerPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container)
     {
-        $overwrite = $container->getParameter('sonata.admin.configuration.admin_services');
+        // NEXT_MAJOR: Remove this line.
```
jordisala1991, line 33: if we remove this line there is no template vars, all the method is to be removed?

dmaicher (dismissed):
LGTM except for the deprecation message namings mentioned by @jordisala1991 

And we need to release this soon as without it admin release 3.98.2 is a bit broken for ORM users

dmaicher (approved):

PR #1438, opened 2021-05-03
Fix compatibility with SonataAdmin and add final

VincentLanglet:


src/Datagrid/Pager.php
```diff
@@ -47,8 +47,8 @@ public function countResults(): int
     public function init(): void
     {
         $query = $this->getQuery();
-        if (null === $query) {
-            throw new \LogicException('The pager need a query to be initialised');
+        if (!$query instanceof ProxyQueryInterface) {
```
VincentLanglet, line 50: This was changed recently.
But we still need the execute method to return a countable value.

src/Filter/AbstractDateFilter.php
```diff
@@ -98,14 +98,7 @@ public function filter(ProxyQueryInterface $query, string $alias, string $field,
         $query->getQueryBuilder()->setParameter($parameterName, $value, $this->getParameterType($value));
     }
 
-    public function getDefaultOptions(): array
```
VincentLanglet, line 101: This method is always overriden.

src/Filter/ClassFilter.php
```diff
@@ -45,33 +45,22 @@ public function filter(ProxyQueryInterface $query, string $alias, string $field,
...
-    }
-
     public function getRenderSettings(): array
     {
         return [DefaultType::class, [
             'operator_type' => $this->getOption('operator_type'),
             'operator_options' => $this->getOption('operator_options'),
             'field_type' => $this->getFieldType(),
-            'field_options' => $this->getFieldOptions(),
+            'field_options' => [
```
VincentLanglet, line 60: I didn't kept the choices option.

I think it's not used. It's supposed to be used this way
```
'choices' => [
    'choices' => [...],
],
```
The syntax is so weird I assume it wasn't used this way.
But I wonder if I shouldn't add `choice_translation_domain => false` in the default values.

jordisala1991 (approved):

dmaicher (approved):

PR #1444, opened 2021-05-13
Implement sensible defaults for Paginator

### LINKED ISSUE #1443: Simple pager crashes on large datasets
#### Sonata packages

<p>

```

sonata-project/admin-bundle              3.99.0 3.99.0 The missing Symfony Admin Generator
sonata-project/block-bundle              4.5.3  4.5.3  Symfony SonataBlockBundle
sonata-project/cache                     2.1.1  2.1.1  Cache library
sonata-project/doctrine-extensions       1.12.0 1.12.0 Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.31.0 3.34.0 Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/exporter                  2.6.2  2.6.2  Lightweight Exporter library
sonata-project/form-extensions           1.9.0  1.9.0  Symfony form extensions
sonata-project/twig-extensions           1.5.1  1.5.1  Sonata twig extensions
vagrant@ubuntu2004:~/apps/api-paztir$
sonata-project/block-bundle              4.5.3  4.5.3  Symfony SonataBlockBundle
sonata-project/cache                     2.1.1  2.1.1  Cache library
sonata-project/doctrine-extensions       1.12.0 1.12.0 Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.31.0 3.34.0 Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/exporter                  2.6.2  2.6.2  Lightweight Exporter library
sonata-project/form-extensions           1.9.0  1.9.0  Symfony form extensions
sonata-project/twig-extensions           1.5.1  1.5.1  Sonata twig extensions

```

</p>
</details>

#### Symfony packages

<p>

```
symfony/contracts                v1.1.10 v2.4.0  A set of abstractions extracted out of the Symfony components
symfony/deprecation-contracts    v2.4.0  v2.4.0  A generic function and convention to trigger deprecation notices
symfony/monolog-bundle           v3.7.0  v3.7.0  Symfony MonologBundle
symfony/phpunit-bridge           v5.2.7  v5.2.7  Provides utilities for PHPUnit, especially user deprecation notices management
symfony/polyfill-ctype           v1.22.1 v1.22.1 Symfony polyfill for ctype functions
symfony/polyfill-intl-grapheme   v1.22.1 v1.22.1 Symfony polyfill for intl's grapheme_* functions
symfony/polyfill-intl-icu        v1.22.1 v1.22.1 Symfony polyfill for intl's ICU-related data and classes
symfony/polyfill-intl-idn        v1.22.1 v1.22.1 Symfony polyfill for intl's idn_to_ascii and idn_to_utf8 functions
symfony/polyfill-intl-normalizer v1.22.1 v1.22.1 Symfony polyfill for intl's Normalizer class and related functions
symfony/polyfill-mbstring        v1.22.1 v1.22.1 Symfony polyfill for the Mbstring extension
symfony/polyfill-php72           v1.22.1 v1.22.1 Symfony polyfill backporting some PHP 7.2+ features to lower PHP versions
symfony/polyfill-php73           v1.22.1 v1.22.1 Symfony polyfill backporting some PHP 7.3+ features to lower PHP versions
symfony/polyfill-php80           v1.22.1 v1.22.1 Symfony polyfill backporting some PHP 8.0+ features to lower PHP versions
symfony/psr-http-message-bridge  v2.1.0  v2.1.0  PSR HTTP message bridge
symfony/security-acl             v3.1.2  v3.1.2  Symfony Security Component - ACL (Access Control List)
symfony/string                   v5.2.6  v5.2.6  Provides an object-oriented API to strings and deals with bytes, UTF-8 code points a...
symfony/symfony                  v4.4.22 v5.2.7  The Symfony PHP framework

```

</p>
</details>

#### PHP version

```
$ php -v
PHP 7.4.18 (cli) (built: May  3 2021 11:27:06) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.18, Copyright (c), by Zend Technologies

```

## Minimal repository with the bug
3.34

## Expected results
should take less than 3 ms

                                                         QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.44..198.90 rows=251 width=1602)
   ->  Group  (cost=0.44..13921690.63 rows=17606664 width=1602)
         Group Key: id
         ->  Index Scan Backward using lock_message_pkey on log c0_  (cost=0.44..13877673.97 rows=17606664 width=1602)

## Actual results

takes 100seconds
                                                       QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=9115117.69..9115119.57 rows=251 width=16)
   ->  Unique  (cost=9115117.69..9247167.67 rows=17606664 width=16)
         ->  Sort  (cost=9115117.69..9159134.35 rows=17606664 width=16)
               Sort Key: (min(dctrn_result.sclr_21)), dctrn_result.id_0
               ->  GroupAggregate  (cost=6511832.11..6819948.73 rows=17606664 width=16)
                     Group Key: dctrn_result.id_0
                     ->  Sort  (cost=6511832.11..6555848.77 rows=17606664 width=16)
                           Sort Key: dctrn_result.id_0
                           ->  Subquery Scan on dctrn_result  (cost=3732479.90..4216663.16 rows=17606664 width=16)
                                 ->  WindowAgg  (cost=3732479.90..4040596.52 rows=17606664 width=2248)
                                       ->  Sort  (cost=3732479.90..3776496.56 rows=17606664 width=8)
                                             Sort Key: c0_.id DESC
                                             ->  Seq Scan on log c0_  (cost=0.00..1472558.64 rows=17606664 width=8)

alfabetagama:
## Subject

Set Paginator to not use fetchJoin if query has single primary key and no joins
Set CountWalker::HINT_DISTINCT to false if there are no joins in query
Set useOutputWalkers to false for simple queries without joins, having clause and single primary key

I am targeting this branch, because pagination is slow for large datasets.

Closes #1443.

## Changelog

```markdown

### Fixed
Fixed issue where pagination of large dataset would take very long time or cause database engine  to swap even for simplest queries without joins.

```

### DIFF
--- a/src/Datagrid/ProxyQuery.php
+++ b/src/Datagrid/ProxyQuery.php
@@ -19,6 +19,7 @@ use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
 use Doctrine\ORM\Tools\Pagination\Paginator;
+use Doctrine\ORM\Tools\Pagination\CountWalker;
 
 /**
  * This class try to unify the query usage with Doctrine.
@@ -201,6 +202,27 @@ class ProxyQuery implements ProxyQueryInterface
         }
 
         $query = $this->getDoctrineQuery();
+        $queryBuilder = $this->getQueryBuilder();
+
+        $identifierFieldNames = $this
+            ->getQueryBuilder()
+            ->getEntityManager()
+            ->getMetadataFactory()
+            ->getMetadataFor(current($this->getQueryBuilder()->getRootEntities()))
+            ->getIdentifierFieldNames();
+        $hasSingleIdentifierName = 1 === \count($identifierFieldNames);
+        $hasNoJoins = 0 ===  \count($queryBuilder->getDQLPart('join'));
+        $havingPart = $queryBuilder->getDQLPart('having');
+
+        if (\is_array($havingPart)  && (\count($havingPart) > 0)) {
+            $hasHavingPart = true;
+        } else {
+            $hasHavingPart = false;
+        }
+
+        if ($hasNoJoins) {
+            $query->setHint(CountWalker::HINT_DISTINCT, false);
+        }
 
         foreach ($this->hints as $name => $value) {
             $query->setHint($name, $value);
@@ -211,16 +233,17 @@ class ProxyQuery implements ProxyQueryInterface
             return $query->execute($params, $hydrationMode);
         }
 
-        $identifierFieldNames = $this
-            ->getQueryBuilder()
-            ->getEntityManager()
-            ->getMetadataFactory()
-            ->getMetadataFor(current($this->getQueryBuilder()->getRootEntities()))
-            ->getIdentifierFieldNames();
-
         // Paginator with fetchJoinCollection doesn't work with composite primary keys
         // https://github.com/doctrine/orm/issues/2910
-        return new Paginator($query, 1 === \count($identifierFieldNames));
+        // To stay safe fetch join always except for queries with single primary key and no joins
+        $paginator = new Paginator($query, !($hasSingleIdentifierName && $hasNoJoins));
+
+        // it is only safe to disable output walkers for really simple queries
+        if (!$hasHavingPart && $hasNoJoins && $hasSingleIdentifierName) {
+            $paginator->setUseOutputWalkers(false);
+        }
+
+        return $paginator;
     }
 
     /**


src/Datagrid/ProxyQuery.php
```diff
@@ -201,6 +202,27 @@ public function execute(array $params = [], $hydrationMode = null)
...
+            ->getEntityManager()
+            ->getMetadataFactory()
+            ->getMetadataFor(current($this->getQueryBuilder()->getRootEntities()))
+            ->getIdentifierFieldNames();
+        $hasSingleIdentifierName = 1 === \count($identifierFieldNames);
+        $hasNoJoins = 0 ===  \count($queryBuilder->getDQLPart('join'));
+        $havingPart = $queryBuilder->getDQLPart('having');
+
+        if (\is_array($havingPart)  && (\count($havingPart) > 0)) {
+            $hasHavingPart = true;
```
jordisala1991, line 218: `$hasHavingPart` = \is_array($havingPart) && (\count($havingPart) > 0)

franmomu:
Maybe you should bump `sonata-project/admin-bundle` to `3.99.1` in `composer.json` to fix the build

src/Datagrid/ProxyQuery.php
```diff
@@ -211,16 +228,17 @@ public function execute(array $params = [], $hydrationMode = null)
...
-            ->getEntityManager()
-            ->getMetadataFactory()
-            ->getMetadataFor(current($this->getQueryBuilder()->getRootEntities()))
-            ->getIdentifierFieldNames();
-
         // Paginator with fetchJoinCollection doesn't work with composite primary keys
         // https://github.com/doctrine/orm/issues/2910
-        return new Paginator($query, 1 === \count($identifierFieldNames));
+        // To stay safe fetch join always except for queries with single primary key and no joins
+        $paginator = new Paginator($query, !($hasSingleIdentifierName && $hasNoJoins));
```
VincentLanglet, line 234:
```
$paginator = new Paginator($query, $hasSingleIdentifierName && !$hasNoJoins);
```

VincentLanglet:
```
// Paginator with fetchJoinCollection doesn't work with composite primary keys
// https://github.com/doctrine/orm/issues/2910
// And fetchJoinCollection is not needed if the query has no join
```

src/Datagrid/ProxyQuery.php
```diff
@@ -201,6 +202,22 @@ public function execute(array $params = [], $hydrationMode = null)
         }
 
         $query = $this->getDoctrineQuery();
+        $queryBuilder = $this->getQueryBuilder();
+
+        $identifierFieldNames = $this
+            ->getQueryBuilder()
```
VincentLanglet, line 208:
```suggestion
        $identifierFieldNames = $queryBuilder
```

src/Datagrid/ProxyQuery.php
```diff
@@ -201,6 +202,22 @@ public function execute(array $params = [], $hydrationMode = null)
...
+            ->getEntityManager()
+            ->getMetadataFactory()
+            ->getMetadataFor(current($this->getQueryBuilder()->getRootEntities()))
+            ->getIdentifierFieldNames();
+        $hasSingleIdentifierName = 1 === \count($identifierFieldNames);
+        $hasNoJoins = 0 === \count($queryBuilder->getDQLPart('join'));
```
VincentLanglet, line 214: Might be better to create a variable `$hasJoin` instead.
This will be similar to `hasHavingPart`. and this avoid `!$hasNoJoin` which is a double negation.

src/Datagrid/ProxyQuery.php
```diff
@@ -201,6 +202,21 @@ public function execute(array $params = [], $hydrationMode = null)
         }
 
         $query = $this->getDoctrineQuery();
+        $queryBuilder = $this->getQueryBuilder();
+
+        $identifierFieldNames = $queryBuilder
+            ->getEntityManager()
+            ->getMetadataFactory()
+            ->getMetadataFor(current($this->getQueryBuilder()->getRootEntities()))
```
VincentLanglet, line 210:
```suggestion
            ->getMetadataFor(current($queryBuilder->getRootEntities()))
```

src/Datagrid/ProxyQuery.php
```diff
@@ -201,6 +202,21 @@ public function execute(array $params = [], $hydrationMode = null)
...
+            ->getIdentifierFieldNames();
+        $hasSingleIdentifierName = 1 === \count($identifierFieldNames);
+        $hasJoins = 0 !== \count($queryBuilder->getDQLPart('join'));
```
VincentLanglet, line 213: Maybe it's better to write
```suggestion
        $hasJoins = \count($queryBuilder->getDQLPart('join')) > 0;
```

VincentLanglet (dismissed):

dmaicher (commented):
nice :+1: thanks for looking into this. This brings performance back to normal levels for my large table admins using the simple pager :rocket: 

Shall we maybe cover this with some tests?

VincentLanglet:
> Shall we maybe cover this with some tests?

I think we can start by merging/releasing this if this PR is ok for you.
Then we could try to add test, but I'm not sure it will be easy to introduce performance tests here.

VincentLanglet:
@franmomu Does your fetch-join queries still work with this ?

dmaicher:
>  Then we could try to add test, but I'm not sure it will be easy to introduce performance tests here.

@VincentLanglet I don't think we need performance tests. But the logic for detecting join/having should have some tests imho

VincentLanglet:
> > Then we could try to add test, but I'm not sure it will be easy to introduce performance tests here.
> 
> @VincentLanglet I don't think we need performance tests. But the logic for detecting join/having should have some tests imho

Oh, I thought you were talking bout functional tests but you meant unit tests.
I agree, it would be nice to add them.

franmomu:
great @alfabetagama! looks like it works fine.

Maybe we could extract this logic to a `SmartPaginatorFactory` with a `createPaginator(QueryBuilder $queryBuilder)` method or something like that.

VincentLanglet:
@alfabetagama Do you have time to create a `SmartPaginatorFactory:: createPaginator` method and add unit test or should someone overtake the PR ? :)

franmomu:
I'll do it and push it here

src/Datagrid/ProxyQuery.php
```diff
@@ -202,6 +202,7 @@ public function execute(array $params = [], $hydrationMode = null)
 
         $query = $this->getDoctrineQuery();
 
+        // NEXT_MAJOR: Remove this block.
         foreach ($this->hints as $name => $value) {
             $query->setHint($name, $value);
         }
```
franmomu, line 208: We could move maybe this`foreach` at the end of `getDoctrineQuery` in `4.x` so we can remove it from `SmartPaginatorFactory` (not sure if it is used though).

src/Datagrid/ProxyQuery.php
```diff
@@ -211,16 +212,7 @@ public function execute(array $params = [], $hydrationMode = null)
...
-            ->getQueryBuilder()
-            ->getEntityManager()
-            ->getMetadataFactory()
-            ->getMetadataFor(current($this->getQueryBuilder()->getRootEntities()))
-            ->getIdentifierFieldNames();
-
-        // Paginator with fetchJoinCollection doesn't work with composite primary keys
-        // https://github.com/doctrine/orm/issues/2910
-        return new Paginator($query, 1 === \count($identifierFieldNames));
+        return SmartPaginatorFactory::create($this, $this->hints);
```
franmomu, line 215: At first I thought the `QueryBuilder` was enough, but we need the query from `getDoctrineQuery` and also `QueryBuilder`.

src/Util/SmartPaginatorFactory.php
```diff
@@ -0,0 +1,67 @@
...
+        foreach ($hints as $name => $value) {
+            $query->setHint($name, $value);
+        }
+
+        // Paginator with fetchJoinCollection doesn't work with composite primary keys
+        // https://github.com/doctrine/orm/issues/2910
+        // To stay safe fetch join only when we have single primary key and joins
+        $paginator = new Paginator($query, $hasSingleIdentifierName && $hasJoins);
+
+        $hasHavingPart = null !== $queryBuilder->getDQLPart('having');
```
franmomu, line 58: I've change this part since [`having` part is `null` by default](https://github.com/doctrine/orm/blob/535bc92dc8280732cdbe1d6ed0a4034a6103996c/lib/Doctrine/ORM/QueryBuilder.php#L84) and [when it's filled it's a `Expr\Andx`](https://github.com/doctrine/orm/blob/535bc92dc8280732cdbe1d6ed0a4034a6103996c/lib/Doctrine/ORM/QueryBuilder.php#L1236-L1249).

VincentLanglet: When calling `resetDQLPart` the having part is reset to `[]`.

src/Util/SmartPaginatorFactory.php
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        foreach ($hints as $name => $value) {
+            $query->setHint($name, $value);
+        }
+
+        // Paginator with fetchJoinCollection doesn't work with composite primary keys
+        // https://github.com/doctrine/orm/issues/2910
+        // To stay safe fetch join only when we have single primary key and joins
+        $paginator = new Paginator($query, $hasSingleIdentifierName && $hasJoins);
```
VincentLanglet, line 56: Do we plan to add an option in order to let the user decide if `fetchJoin` should be true ?
You might want to set `fetchJoin = false` even if there are join, because you're not using fetch join ; no ?

If so, the config will be pass in the constructor of the Factory and this method won't be able to be static. Should we anticipate this now ?

src/Datagrid/ProxyQuery.php
```diff
@@ -202,22 +201,8 @@ public function execute(array $params = [], $hydrationMode = null)
...
-        $hasSingleIdentifierName = 1 === \count($identifierFieldNames);
-        $hasJoins = 0 !== \count($queryBuilder->getDQLPart('join'));
-        $havingPart = $queryBuilder->getDQLPart('having');
-        $hasHavingPart = \is_array($havingPart) && (\count($havingPart) > 0);
-
-        if (!$hasJoins) {
-            $query->setHint(CountWalker::HINT_DISTINCT, false);
-        }
 
+        // NEXT_MAJOR: Remove this block.
```
VincentLanglet, line 205: Why this block can't be remove now since it's in the Paginator ?

src/Datagrid/ProxyQuery.php
VincentLanglet: I'm wonder if it's not better to deprecate it since it's not part of the API...

VincentLanglet (changes requested):

src/Datagrid/ProxyQuery.php
franmomu: It was added in https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/730 which seems legit.

src/Util/SmartPaginatorFactory.php
franmomu: not really, see https://github.com/doctrine/orm/blob/535bc92dc8280732cdbe1d6ed0a4034a6103996c/lib/Doctrine/ORM/QueryBuilder.php#L1492-L1498

it depends on the current value

src/Util/SmartPaginatorFactory.php
franmomu: TBH I don't know, I created `SmartPaginatorFactory` as internal to just extract that code, so it could be changed the way we want. I would wait to do that when these options are created.

src/Datagrid/ProxyQuery.php
franmomu: I left it because there is this code after:

```php
// NEXT_MAJOR: Remove this.
if (\func_num_args() > 0) {
    return $query->execute($params, $hydrationMode);
}
```
So it could break BC.

src/Datagrid/ProxyQuery.php
VincentLanglet: Ok

src/Util/SmartPaginatorFactory.php
VincentLanglet: Yes, so if you add an having part, it will become an array. And then when you reset it, since it's an array it will be reset to an empty array.

dmaicher: I don't think so. The `having` part is never an array internally, or?

VincentLanglet: I thought it was, but might be ok.

VincentLanglet (dismissed):

src/Datagrid/ProxyQuery.php
```diff
@@ -202,6 +202,7 @@ public function execute(array $params = [], $hydrationMode = null)
 
         $query = $this->getDoctrineQuery();
```
dmaicher, line 203: hm now we execute `->getDoctrineQuery()` twice? I think we can avoid that, or?

VincentLanglet: All the logic could be in `if (\func_num_args() > 0) {`

VincentLanglet (approved):

dmaicher (approved):
:+1:

VincentLanglet:
Thanks

PR #1446, opened 2021-05-17
Improve autocomplete test with more assertions

franmomu:
Looking at the test I added, there should be assertions on the result.


franmomu:
@phansys looks like now it the autocomplete filter is showing all the items

PR #1457, opened 2021-06-06
Add `treat_null_as` option to BooleanFilter

### LINKED ISSUE #1436: Add an option to BooleanFilter in order to handle `null` values.
## Feature Request

I often use nullable boolean field, and I'd like to use filter on these fields.

I wrote a callbackFilter to consider `null` values as false.
```
        if (!$data->hasValue()) {
            return false;
        }

        if (BooleanType::TYPE_YES === $data->getValue()) {
            $query->andWhere("$alias.$field = 1");
        } elseif (BooleanType::TYPE_NO === $data->getValue()) {
            $query->andWhere("$alias.$field = 0 OR $alias.$field IS NULL");
        } else {
            throw new \InvalidArgumentException(sprintf('Unsupported value %s', $data->getValue()));
        }

        return true;
```
I wonder if an option could be added to the BooleanFilter.

I see three use case
- You might want to treat `null` value as false.
- You might want to treat `null` value as true.
- You might want to treat `null` as a different value.

The first two case could be handle by an option `treat_null_as => true/false (null by default)`. 
The last one cannot by handle currently since the BooleanType has only TYPE_YES/TYPE_NO values. (But maybe a `nullable` option could be added).

What do you think @sonata-project/contributors ?

VincentLanglet:
## Subject

Closes #1436

## Changelog

```markdown
### Added
- `treat_null_as` option to BooleanFilter
```

### DIFF
--- a/src/Filter/BooleanFilter.php
+++ b/src/Filter/BooleanFilter.php
@@ -42,28 +42,9 @@ class BooleanFilter extends Filter
         }
 
         if (\is_array($data['value'])) {
-            $values = [];
-            foreach ($data['value'] as $v) {
-                if (!\in_array($v, [BooleanType::TYPE_NO, BooleanType::TYPE_YES], true)) {
-                    continue;
-                }
-
-                $values[] = (BooleanType::TYPE_YES === $v) ? 1 : 0;
-            }
-
-            if (0 === \count($values)) {
-                return;
-            }
-
-            $this->applyWhere($query, $query->getQueryBuilder()->expr()->in(sprintf('%s.%s', $alias, $field), $values));
+            $this->filterWithMultipleValues($query, $alias, $field, $data);
         } else {
-            if (!\in_array($data['value'], [BooleanType::TYPE_NO, BooleanType::TYPE_YES], true)) {
-                return;
-            }
-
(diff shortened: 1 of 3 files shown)


VincentLanglet:
Please review @sonata-project/contributors

VincentLanglet:
Please review @jordisala1991 @franmomu @phansys

src/Filter/BooleanFilter.php
```diff
@@ -86,4 +68,57 @@ public function getRenderSettings()
...
+            $or->add($query->getQueryBuilder()->expr()->isNull(sprintf('%s.%s', $alias, $field)));
+        }
+
+        $or->add($query->getQueryBuilder()->expr()->in(sprintf('%s.%s', $alias, $field), $values));
+
+        $this->applyWhere($query, $or);
+    }
+
+    /**
+     * NEXT_MAJOR: Change the typehint to ProxyQueryInterface.
```
phansys, line 105:
```suggestion
     * NEXT_MAJOR: Change the typehint of argument 1 to ProxyQueryInterface.
```

phansys (approved):

jordisala1991 (approved):

PR #1467, opened 2021-06-24
Enable global search on string filters only

EmmanuelVella:
## Enable global search on string filters only

Adding global search on field of type `string` is not correct. In case of `string` field using a filter with a `ChoiceFilter`, the filter will be ignored (if search term is not in the choices list), which can lead to unfiltered query (with no `WHERE` clause).

I am targeting this branch, because it's a bug fix.

## Changelog

```markdown
### Fixed

- Global search returning wrong results.
```

### DIFF
--- a/src/Builder/DatagridBuilder.php
+++ b/src/Builder/DatagridBuilder.php
@@ -26,6 +26,7 @@ use Sonata\AdminBundle\Filter\FilterFactoryInterface;
 use Sonata\AdminBundle\Guesser\TypeGuesserInterface as DeprecatedTypeGuesserInterface;
 use Sonata\DoctrineORMAdminBundle\Datagrid\Pager;
 use Sonata\DoctrineORMAdminBundle\Filter\ModelAutocompleteFilter;
+use Sonata\DoctrineORMAdminBundle\Filter\StringFilter;
 use Symfony\Component\Form\Extension\Core\Type\FormType;
 use Symfony\Component\Form\FormFactoryInterface;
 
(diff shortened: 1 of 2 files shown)


VincentLanglet (commented):
Only `StringFilter::class` should be kept instead of `doctrine_orm_string`.
Dunno if it works in 3.x but it should be at least deprecated and in 4.0 we should only support `StringFilter::class`.

Also, the issue with this implementation is that if someone is doing a custom StringFilter, it wont be used anymore in the global search. Not sure it's intended.

What is the issue you have with the choiceFilter and the global search ? I'm not sure to understand the unfiltered query behavior

VincentLanglet:
What if we introduce a new interface `SearchableFilter` with a method `isSearchable`.
We use it here: https://github.com/sonata-project/SonataAdminBundle/blob/3.x/src/Search/SearchHandler.php#L91, changing the code to `if ($filter instanceof SearchableFilter and $filter->isSearchable)`.

Then in the StringFilter, we implement the interface and add the method
```
public function isSearchable(): bool
{
    return $this->getOption('global_search', true);
}
```

This way:
- Only the StringFilter is used for the globalSearch
- If someone implement a custom filter, he can be searchable
- passing `global_search` => false, still disable the StringFilter.

VincentLanglet:
I started https://github.com/sonata-project/SonataAdminBundle/pull/7275 in order to discuss about the idea.

EmmanuelVella:
Nice ! Thank you. Let me close my PR then !

VincentLanglet:
Yes, the PR https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1470 will solve your issue.

PR #1484, opened 2021-07-21
Add phpstan strict rules

VincentLanglet:
I have two issue to fix
```
 ------ --------------------------------------------------------------- 
  Line   src/Filter/ModelAutocompleteFilter.php                         
 ------ --------------------------------------------------------------- 
  89     Construct empty() is not allowed. Use more strict comparison.  
 ------ --------------------------------------------------------------- 

 ------ --------------------------------------------------------------- 
  Line   src/Filter/ModelFilter.php                                     
 ------ --------------------------------------------------------------- 
  28     Construct empty() is not allowed. Use more strict comparison.  
 ------ --------------------------------------------------------------- 
```

But I'm not sure how to replace the empty checks of `$data->getValue()`.

I never used those filter, what can be the values @sonata-project/contributors ?

### DIFF
--- a/src/Filter/ModelAutocompleteFilter.php
+++ b/src/Filter/ModelAutocompleteFilter.php
@@ -86,10 +86,6 @@ final class ModelAutocompleteFilter extends Filter
 
     protected function handleModel(ProxyQueryInterface $query, string $alias, FilterData $data): void
     {
-        if (empty($data->getValue())) {
-            return;
-        }
-
         $parameterName = $this->getNewParameterName($query);
 
         if ($data->isType(EqualOperatorType::TYPE_NOT_EQUAL)) {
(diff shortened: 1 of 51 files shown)


VincentLanglet:
Test are passing, if I remove the empty check, but I'm not sure about this... if someone use those filter, it would be interesting to know if possible empty values are passed.

src/Filter/ModelAutocompleteFilter.php
```diff
@@ -86,10 +86,6 @@ protected function handleMultiple(ProxyQueryInterface $query, string $alias, Fil
 
     protected function handleModel(ProxyQueryInterface $query, string $alias, FilterData $data): void
     {
-        if (empty($data->getValue())) {
```
jordisala1991, line 89: maybe this should check for hasValue?

VincentLanglet: Already done in the filter method, 

If we want to add check if would be aroung falsy value like `''`, `null` or `[]` but I don't know which one are really possible to land here.

dmaicher (approved):

PR #1511, opened 2021-08-24
In the example form docs it does not work with the serialize function

vindert:
## Subject

When using the example from the docs on docs/reference/filter_field_definition.rst (section about the StringListFilter) it does not work with the serialize function.
When i remove it, it does work. But maybe it does need to be here for other uses of this filter?

I am targeting this branch, because not sure.

## Changelog

```markdown
### Fixed
Remove serialize function form StringListFilter

```

### DIFF
--- a/src/Filter/StringListFilter.php
+++ b/src/Filter/StringListFilter.php
@@ -39,7 +39,7 @@ final class StringListFilter extends Filter
             $parameterName = $this->getNewParameterName($query);
             $andConditions->add(sprintf('%s.%s %s :%s', $alias, $field, $operator, $parameterName));
 
-            $query->getQueryBuilder()->setParameter($parameterName, '%'.serialize($item).'%');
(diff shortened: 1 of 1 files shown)


VincentLanglet (changes requested):
It works perfectly fine in my project with the following code
```
->add('roles', StringListFilter::class, [
                'field_type'    => ChoiceType::class,
                'field_options' => [
                    'choices'                   => AdminUser::ROLES,
                    'choice_translation_domain' => false,
                    'multiple'                  => true,
                ],
            ])
```
and the field
```
    /**
     * @var string[]
     *
     * @ORM\Column(type="array", options={"default":"a:0:{}"})
     */
    private $roles = [];
```

The array has the following syntax in the database
```
a:2:{i:1;s:3:"foo";i:2;s:3:"bar";}
```
so looking for `s:3:"foo";` is the best solution to avoid false positive.

PR #1545, opened 2021-10-01
Deprecate ModelAutocompleteFilter

### LINKED ISSUE #1513: The ModelAutocompleteType is a duplicated class of ModelFilter with a different field_type.
I took a look at https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1040, I wanted to see what was missing to fix the issue ; so I looked at the difference between ModelFilter and ModelAutocompleteType.
1) The filter method is the same
2) The handle multiple method is the same, except that ModelFilter add the null check for NotEqual
3) The handle single method is the same, except that ModelFilter add the null check for NotEqual
4) The association method is the same, except that ModelFilter add a check about allowed values for mapping type
5) ModelFilter has a private method getParentAlias needed for the null check 
6) ModelFilter has a `'mapping_type'` option.
7) ModelFilter has a `'field_type' => EntityType::class,` default option when ModelAutocompleteFilter use `ModelAutocompleteType`.

If we change NotEqual behavior to return null value too in ModelAutocompleteFilter, the differences 2, 3, 4, 5, 6 won't be.

So ModelAutocompleteFilter will be a duplicate of ModelFilter with just a config diff.

Should we then deprecate the ModelAutocompleteFilter, in favor of
```
->add('foo', ModelType::class, [
     'field_type' => ModelAutocompleteType::class, 
])
```
?

### LINKED ISSUE #1526: CallbackFilter and ModelAutocompleteType combination is far too complicated
I currently have a `CallbackFilter` on an admin service to set a custom database query:

```php
class ServerAdmin extends AbstractAdmin
{
    public function callbackFilterOwners(ProxyQuery $queryBuilder, string $alias, string $field, array $value): bool
    {
        $queryBuilder
            ->leftJoin('o.organizations', 'org')
            ->andWhere(
                $queryBuilder->expr()->isMemberOf('org', 'o.organizations'),
                $queryBuilder->expr()->in('org', ':userOrganization')
            )
            ->orWhere(':user MEMBER OF o.owners')
            ->setParameter('user', $value['value'])
            ->setParameter('userOrganization', $value['value']->getOrganizations())
        ;

        return true;
    }

    protected function configureDatagridFilters(DatagridMapper $datagridMapper): void
    {
        $datagridMapper
            ->add('owners', CallbackFilter::class, [
                'callback' => [$this, 'callbackFilterOwners'],
                'field_type' => \App\Form\Type\EntityUserType::class,
            ])
        ;
    }
```

This works as expected, however it renders only local html select, which cause performance issue on large databases.

This is why I need the autocomplete. However, I can not use `CallbackFilter` and `ModelAutocompleteFilter` together.

The solution is to keep the `CallbackFilter` and specify the `ModelAutocompleteType` to it's `field_type` option.

This is possible, but then we have to pass 4 new options to this type. Here is the compete configuration code:

```php
$datagridMapper
    ->add('owners', CallbackFilter::class, [
        'callback' => [$this, 'callbackFilterOwners'],
        'field_name' => 'owners',
    ], [
        'field_type' => ModelAutocompleteType::class,
        'field_options' => [
            'property' => [
                'username',
                'firstname',
                'lastname',
            ],
            'to_string_callback' => static fn (User $entity) => $entity->getFullNameWithUsername(),
            'model_manager' => $this->getModelManager(),
            'class' => User::class,
            'admin_code' => $this->getCode(),
            'context' => 'filter',
        ],
    ])
;
```

I assume this is because of this part: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/9e65730b021ad5af251811f8b0d929b7821ed8c3/src/Builder/DatagridBuilder.php#L84-L91

This produce a lot of configuration noise. By the way, the `context` option to give was hard to find because of lack of documentation (or my blindness).

VincentLanglet:
## Subject

Closes #1513.
Closes #1526.

## Changelog

```markdown
### Changed
- `field_options` are automatically passed to the Filter when an `EntityType` or a `ModelAutocompleteType` is used.

### Deprecated
- `ModelAutocompleteFilter` in favor of `ModelFilter` with a `field_type` `ModelAutocompleteType`
```

### DIFF
--- a/src/Resources/config/doctrine_orm_filter_types.xml
+++ b/src/Resources/config/doctrine_orm_filter_types.xml
@@ -28,6 +28,7 @@
         <service id="sonata.admin.orm.filter.type.datetime_range" class="Sonata\DoctrineORMAdminBundle\Filter\DateTimeRangeFilter">
             <tag name="sonata.admin.filter.type" alias="doctrine_orm_datetime_range"/>
         </service>
+        <!-- NEXT_MAJOR: Remove the sonata.admin.orm.filter.type.model_autocomplete service -->
         <service id="sonata.admin.orm.filter.type.model_autocomplete" class="Sonata\DoctrineORMAdminBundle\Filter\ModelAutocompleteFilter">
             <tag name="sonata.admin.filter.type" alias="doctrine_orm_model_autocomplete"/>
         </service>
--- a/UPGRADE-4.x.md
+++ b/UPGRADE-4.x.md
@@ -1,0 +1,20 @@
+UPGRADE 4.x
+===========
+
+UPGRADE FROM 4.x to 4.x
+=======================
+
+### Sonata\DoctrineORMAdminBundle\Filter\ModelAutocompleteFilter
+
+The `Sonata\DoctrineORMAdminBundle\Filter\ModelAutocompleteFilter` filter is deprecated.
+
+Instead of
+```php
+->add('foo', ModelAutocompleteFilter::class)
+```
+use
+```php
+->add('foo', ModelFilter::class, [
+     'field_type' => ModelAutocompleteType::class,
+])
+```
--- a/src/Builder/DatagridBuilder.php
+++ b/src/Builder/DatagridBuilder.php
@@ -22,9 +22,12 @@ use Sonata\AdminBundle\Datagrid\SimplePager;
 use Sonata\AdminBundle\FieldDescription\FieldDescriptionInterface;
 use Sonata\AdminBundle\FieldDescription\TypeGuesserInterface;
 use Sonata\AdminBundle\Filter\FilterFactoryInterface;
+use Sonata\AdminBundle\Form\Type\ModelAutocompleteType;
 use Sonata\DoctrineORMAdminBundle\Datagrid\Pager;
 use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface;
 use Sonata\DoctrineORMAdminBundle\Filter\ModelAutocompleteFilter;
+use Sonata\DoctrineORMAdminBundle\Filter\ModelFilter;
+use Symfony\Bridge\Doctrine\Form\Type\EntityType;
 use Symfony\Component\Form\Extension\Core\Type\FormType;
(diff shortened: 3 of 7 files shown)


VincentLanglet:
There is still the code 
```
if (ModelAutocompleteFilter::class === $fieldDescription->getType()) {
            $fieldDescription->mergeOption('field_options', [
                'class' => $fieldDescription->getTargetModel(),
                'model_manager' => $fieldDescription->getAdmin()->getModelManager(),
                'admin_code' => $fieldDescription->getAdmin()->getCode(),
                'context' => 'filter',
            ]);
        }
```
to resolve.

And I think if it's done another way it could sove https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1526 too

VincentLanglet:
PLease review @sonata-project/contributors

core23 (approved):
Looks okay so far.

The downside is that you need to write more code to archive the same behaviour.

src/Resources/config/doctrine_orm_filter_types.xml
```diff
@@ -28,6 +28,7 @@
         <service id="sonata.admin.orm.filter.type.datetime_range" class="Sonata\DoctrineORMAdminBundle\Filter\DateTimeRangeFilter">
             <tag name="sonata.admin.filter.type" alias="doctrine_orm_datetime_range"/>
         </service>
+        <!-- NEXT_MAJOR: Remove the sonata.admin.orm.filter.type.model_autocomplete service -->
```
jordisala1991, line 31: xml, we didnt move this to php?

jordisala1991 (approved):

src/Resources/config/doctrine_orm_filter_types.xml
VincentLanglet: Seems like we didnt https://github.com/sonata-project/SonataDoctrineORMAdminBundle/tree/4.x/src/Resources/config

VincentLanglet: Created https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1546

PR #1547, opened 2021-10-14
Reduce visibility from getEntityManager

VincentLanglet:
## Changelog

```markdown
### Deprecated
- Calling `ModelManager::getEntityManager()`
```

### DIFF
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -167,6 +167,8 @@ final class ModelManager implements ModelManagerInterface, LockInterface
     }
 
     /**
+     * NEXT_MAJOR: Change the visibility to private.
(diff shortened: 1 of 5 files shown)


src/Model/ModelManager.php
```diff
@@ -167,6 +167,8 @@ public function findOneBy(string $class, array $criteria = []): ?object
     }
 
     /**
+     * NEXT_MAJOR: Change the visibility to private.
```
core23, line 170: Maybe we could use `@internal` to mark this as private?

jordisala1991:
This needs to be merged first to fix the build right?

jordisala1991 (approved):

Issue #1548, opened 2021-10-15
Support doctrine/dbal 3.1


simonberger:
All packages we use support `doctrine/dbal` `3.1` already or will do with the next release.
I looked through all dbal uses and there is nothing affected by the 3.0 changes or by 3.1 deprecations.

Requires: https://github.com/sonata-project/EntityAuditBundle/pull/444

PR #1549, opened 2021-10-15
Allow doctrine/dbal 3.1

simonberger:
## Subject
This allows doctrine/dbal 3.1. breaking changes do not affect us.

## Todos

- [ ]  A Release with https://github.com/sonata-project/EntityAuditBundle/pull/444

## Changelog
```markdown
### Added
- Added support for Doctrine DBAL 3.1.
```

## Closes

https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1548

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,9 +26,9 @@
     ],
     "require": {
         "php": "^7.3 || ^8.0",
-        "doctrine/dbal": "^2.13",
+        "doctrine/dbal": "^2.13 || ^3.1",
         "doctrine/doctrine-bundle": "^2.3",
-        "doctrine/orm": "^2.8",
+        "doctrine/orm": "^2.10",
         "doctrine/persistence": "^2.1",
         "sonata-project/admin-bundle": "^4.0",
         "sonata-project/exporter": "^2.0",


core23 (dismissed):

VincentLanglet (changes requested):
Test are running on dbal 2

simonberger:
Yes, we need to re-run once the new EntitiyAuditBundle release is there and should also be merge just then.

VincentLanglet:
> Yes, we need to re-run once the new EntitiyAuditBundle release is there and should also be merge just then.

Not sure about this, since unit tests and phpstan is already failing.
Another PR to just fix tests could be helpful

simonberger:
Ah ok didn't look at the check output at all. Most is related to new ORM versions but also sonata-project/twig-extensions.

I could have a look at that next week. More than welcome if someone else finds the time.

simonberger:
Reverted increasing the minimum ORM version (for now) to see if tests run through then. The missing table errors are looking strange.
It seems I need to be allowed to run the pipelines for every commit tho.

simonberger:
I fixed the PHPStan problems from a previous run. In my setup most likely the relevant tests causing the issue when increasing the ORM version are skipped which means it is impossible for me to reproduce and fix the possible problem with the schema preparation in the database (or it was just temporary).
Having to wait to see the pipelines in the PR does not make it easier. 😉

PR #1564, opened 2021-10-28
Fix test platforms workflow

jordisala1991:
Since the workflow is not handled by dev-kit, we forgot to modify it when changing from master to 4.x, and when creating 5.x branch and when we did some modification to the quotes of all workflows. This PR fixed all of it. It is pedantic since it doesn't affect our users.


jordisala1991:
We have some problems with the sessions on the tests, not sure why or how can it be fixed, but its happening

VincentLanglet (approved):

PR #1566, opened 2021-10-31
DevKit updates for 3.x branch

SonataCI:


jordisala1991:
We probably need to avoid dbal 3 here.

VincentLanglet:
> We probably need to avoid dbal 3 here.

You mean adding a conflict ? Indeed it seems like we dont support it.

jordisala1991:
> > We probably need to avoid dbal 3 here.
> 
> 
> 
> You mean adding a conflict ? Indeed it seems like we dont support it.

We dont support it on 3.x , we only do on 4.x

PR #1615, opened 2022-01-11
Fix phpstan issues

VincentLanglet:
Pedantic

### DIFF
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -69,11 +69,11 @@ final class ModelManager implements ModelManagerInterface, LockInterface
             $entityManager = $this->getEntityManager($object);
             $entityManager->persist($object);
             $entityManager->flush();
-        } catch (\PDOException | Exception $e) {
+        } catch (\PDOException | Exception $exception) {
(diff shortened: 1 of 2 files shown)


src/Model/ModelManager.php
```diff
@@ -130,8 +130,12 @@ public function lock(object $object, ?int $expectedVersion): void
         try {
             $entityManager = $this->getEntityManager($object);
             $entityManager->lock($object, LockMode::OPTIMISTIC, $expectedVersion);
-        } catch (OptimisticLockException $e) {
-            throw new LockException($e->getMessage(), (int) $e->getCode(), $e);
+        } catch (OptimisticLockException $exception) {
+            throw new LockException(
+                $exception->getMessage(),
+                $exception->getCode(),
```
jordisala1991, line 136: why in this case you don't cast to int?

VincentLanglet: Only PDOException return a string code.
Phpstan now understand this. Psalm will soon.

jordisala1991 (approved):

PR #1583, opened 2021-11-19
Paginator with useOutputWalkers=false allows query with joins

ossinkine:
## Subject

I am targeting this branch, because it fixing the issue.

Closes #1582

## Changelog

```markdown
### Fixed
Paginator set useOutputWalkers to false for query with joins
```

PR #1585, opened 2021-11-24
Paginator with useOutputWalkers=false allows query with joins (#1583) (Backport commit from 4.x to 3.x) 

JustDylan23:
(cherry picked from commit 77cd79f925d069435593ca72599a2e4cf1b3b8df)

PR description: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1583

Since the new paginator was introduced in 3.x I have backported this fix via git cherry-pick as suggested by https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1583#issuecomment-977863640

I hope this is what you meant when you told me to backport it, if not please correct me in the comments

## Changelog

```markdown
### Fixed
Paginator set useOutputWalkers to false for query with joins
```

PR #1616, opened 2022-01-13
Fix detection of output walkers in pagination

franmomu:
## Subject

After https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1583, queries with an order by using a field from a join association failed throwing an `RuntimeException`. So something like:

```php
->createQueryBuilder()
->from(Author::class, 'author')
->leftJoin('author.books', 'book')
->orderBy('book.title')
```

results in:

```
"Cannot select distinct identifiers from query with LIMIT and ORDER BY on a column from a fetch joined to-many association. Use output walkers."
```

This PR tries to disable output walkers when possible according to https://github.com/doctrine/orm/issues/8278#issue-705517756.

I am targeting this branch, because this is a bugfix.

## Changelog

```markdown
### Fixed
- Disabling output walkers when paginating with order by from an association
```

### DIFF
--- a/src/Util/SmartPaginatorFactory.php
+++ b/src/Util/SmartPaginatorFactory.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\DoctrineORMAdminBundle\Util;
 
+use Doctrine\ORM\QueryBuilder;
 use Doctrine\ORM\Tools\Pagination\CountWalker;
 use Doctrine\ORM\Tools\Pagination\Paginator;
 use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface;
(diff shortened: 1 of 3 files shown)


src/Util/SmartPaginatorFactory.php
```diff
@@ -60,13 +61,91 @@ public static function create(ProxyQueryInterface $proxyQuery, array $hints = []
...
+        foreach ($joinParts as $joinPart) {
+            foreach ($joinPart as $join) {
+                $joinAliases[] = $join->getAlias();
+            }
+        }
+
+        foreach ($orderByParts as $orderByPart) {
+            foreach ($orderByPart->getParts() as $part) {
+                foreach ($joinAliases as $joinAlias) {
+                    if (0 === strpos($part, $joinAlias)) {
```
phansys: Which values can be held in `$part` and `$joinAlias`? My concern is about avoiding `strpos()` to return `0` in cases like `strpos('alias', 'a')`, as I think it isn't the expected behavior here.

franmomu: Interesting, thanks for pointing it out, I think it could be fixed appending a dot to the `$joinAlias` part. 

`$joinAliases` could contain any join alias (could be `a`) and `$part` contains the ordering field that could be `alias.name`.

I'll try and add a test for that.

phansys (approved):

VincentLanglet (approved):

krewetka:
Is there an option to apply/port it also to version 3.35?

I experience this problem in this version as well. 

In version 3.35 there is also only condition `  if (!$hasHavingPart && $hasSingleIdentifierName) {`

From what I see it was introduced to 3.x here https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1585/files

VincentLanglet:
> Is there an option to apply/port it also to version 3.35?

Hi ! Feel free to backport the commit to the PR 3.x, I'll do the release.

PR #1622, opened 2022-01-17
Remove list field definition

eerison:
As it was moved to Sonata admin bundle, IMO it can be removed!


VincentLanglet (approved):
I assume we can do the same for SonataDoctrineMongoDBAdminBundle @franmomu ?

eerison:
> I assume we can do the same for SonataDoctrineMongoDBAdminBundle @franmomu ?

https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/pull/724

PR #1631, opened 2022-02-14
Add support for the `model_class` attribute & fix compiler pass argument error

nocive:
## Subject

I am targeting this branch, because it's a backwards compatible change.

Fixes

```
In Definition.php line 321:
                                                                                               
  [Symfony\Component\DependencyInjection\Exception\OutOfBoundsException]                       
  The argument "1" doesn't exist in class "My\Sonata\AdminClass".  
                                                                                               

Exception trace:
  at /src/application/core/vendor/symfony/dependency-injection/Definition.php:321
 Symfony\Component\DependencyInjection\Definition->getArgument() at /src/application/core/vendor/sonata-project/doctrine-orm-admin-bundle/src/DependencyInjection/Compiler/AddAuditEntityCompilerPass.php:52
 Sonata\DoctrineORMAdminBundle\DependencyInjection\Compiler\AddAuditEntityCompilerPass->process() at /src/application/core/vendor/symfony/dependency-injection/Compiler/Compiler.php:82
```

When using the `model_class` tag attribute, rather than using explicit indexed arguments.

## Changelog

```markdown
### Fixed
- Support for the `model_class` attribute for sonata admin classes in `AddAuditEntityCompilerPass`.
```

### DIFF
--- a/src/DependencyInjection/Compiler/AddAuditEntityCompilerPass.php
+++ b/src/DependencyInjection/Compiler/AddAuditEntityCompilerPass.php
@@ -48,6 +48,7 @@ final class AddAuditEntityCompilerPass implements CompilerPassInterface
             }
 
             $definition = $container->getDefinition($id);
+            // NEXT_MAJOR: Support only model_class and remove indexed argument support
             $modelClass = $attributes[0]['model_class'] ?? $definition->getArgument(1);
             $auditedEntities[] = $this->getModelName($container, $modelClass);
         }
--- a/src/DependencyInjection/Compiler/AddAuditEntityCompilerPass.php
+++ b/src/DependencyInjection/Compiler/AddAuditEntityCompilerPass.php
@@ -48,7 +48,8 @@ final class AddAuditEntityCompilerPass implements CompilerPassInterface
             }
 
             $definition = $container->getDefinition($id);
-            $auditedEntities[] = $this->getModelName($container, $definition->getArgument(1));
+            $modelClass = $attributes[0]['model_class'] ?? $definition->getArgument(1);
+            $auditedEntities[] = $this->getModelName($container, $modelClass);
         }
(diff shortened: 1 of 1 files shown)


nocive:
cc @VincentLanglet

VincentLanglet (dismissed):

jordisala1991 (dismissed):

src/DependencyInjection/Compiler/AddAuditEntityCompilerPass.php
```diff
@@ -48,7 +48,8 @@ public function process(ContainerBuilder $container): void
             }
 
             $definition = $container->getDefinition($id);
-            $auditedEntities[] = $this->getModelName($container, $definition->getArgument(1));
+            $modelClass = $attributes[0]['model_class'] ?? $definition->getArgument(1);
```
jordisala1991, line 52: I think we could add a next major comment

nocive: Happy to, care to suggest the format for one? :)
Something like?

```suggestion
            // NEXT_MAJOR: Support only model_class and remove indexed argument support
            $modelClass = $attributes[0]['model_class'] ?? $definition->getArgument(1);
```

jordisala1991: Exactly, that would be nice!

nocive: Done :+1: 

Please note that SonataAdmin also supports using `setModelClass` via `calls` in the service definition, which this code won't be compatible with, but I suppose that's part of a bigger and wider problem that this PR doesn't address.

VincentLanglet (approved):

PR #1635, opened 2022-02-23
Fix pagination with ToMany sort set at ProxyQuery

jordisala1991:
## Subject

## Changelog

```markdown
### Fixed
- Fixed smart pagination disabling output walkers when the orderBy is set to a ToMany at proxy level.
```

### DIFF
--- a/src/Util/SmartPaginatorFactory.php
+++ b/src/Util/SmartPaginatorFactory.php
@@ -13,7 +13,6 @@ declare(strict_types=1);
 
 namespace Sonata\DoctrineORMAdminBundle\Util;
 
-use Doctrine\ORM\QueryBuilder;
 use Doctrine\ORM\Tools\Pagination\CountWalker;
 use Doctrine\ORM\Tools\Pagination\Paginator;
 use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery;
--- a/src/Util/SmartPaginatorFactory.php
+++ b/src/Util/SmartPaginatorFactory.php
@@ -57,7 +57,7 @@ final class SmartPaginatorFactory
         $paginator = new Paginator($query, $hasSingleIdentifierName && $hasJoins);
 
         // it is only safe to disable output walkers for really simple queries
-        if (self::canDisableOutPutWalkers($queryBuilder)) {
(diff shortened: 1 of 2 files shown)


src/Util/SmartPaginatorFactory.php
```diff
@@ -102,24 +104,26 @@ private static function canDisableOutPutWalkers(QueryBuilder $queryBuilder): boo
...
+        $queryBuilder = $proxyQuery->getQueryBuilder();
+
         $joinParts = $queryBuilder->getDQLPart('join');
 
         if (0 === \count($joinParts)) {
             return false;
         }
 
-        $orderByParts = $queryBuilder->getDQLPart('orderBy');
+        $sortBy = $proxyQuery->getSortBy();
```
jordisala1991, line 123: the sortBy for the paginator is only by one field right?

VincentLanglet: I think you need to check both 
```
$queryBuilder->getDQLPart('orderBy');
```
and `getSortBy`.

Because people can call directly `addOrderBy` in the configureQuery.

VincentLanglet (changes requested):

src/Util/SmartPaginatorFactory.php
jordisala1991: You're right, both things needs to be check:

```
        if ($this->getSortBy()) {
            $orderByDQLPart = $queryBuilder->getDQLPart('orderBy');
            $queryBuilder->resetDQLPart('orderBy');

            $sortBy = $this->getSortBy();
            if (false === strpos($sortBy, '.')) {
                $sortBy = $rootAlias.'.'.$sortBy;
            }

            $queryBuilder->addOrderBy($sortBy, $this->getSortOrder());
            foreach ($orderByDQLPart as $orderBy) {
                $queryBuilder->addOrderBy($orderBy);
            }
        }
```

Here it will reset the orderBy from the queryBuilder, and add the order by from the sortOrder, but later will do a foreach to add the rest of orders.

jordisala1991:
This will need to be done on 4.x too

VincentLanglet (approved):

VincentLanglet:
> This will need to be done on 4.x too

The merge 3.x into 4.x should do the job

PR #1656, opened 2022-05-20
allow doctrine/persistence ^3.0

dmaicher:
## Subject

Allow to install `doctrine/persistence` 3

## Changelog

```markdown
### Added
- Added support for `doctrine/persistence` 3
```

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -15,3 +15,6 @@ parameters:
         - # https://github.com/doctrine/orm/pull/8767
             message: '#Method Sonata\\DoctrineORMAdminBundle\\Model\\ModelManager::getMetadata\(\) should return Doctrine\\ORM\\Mapping\\ClassMetadata\<TObject of object\> but returns Doctrine\\ORM\\Mapping\\ClassMetadata\<object\>\.$#'
             path: src/Model/ModelManager.php
+        - # https://github.com/doctrine/orm/pull/9778
+            message: '#Parameter \#1 \$className of method Doctrine\\Persistence\\Mapping\\AbstractClassMetadataFactory<Doctrine\\ORM\\Mapping\\ClassMetadata>::getMetadataFor\(\) expects class-string, string given.#'
+            path: src/Datagrid/ProxyQuery.php
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "doctrine/dbal": "^2.13 || ^3.0",
         "doctrine/doctrine-bundle": "^2.3",
         "doctrine/orm": "^2.8",
-        "doctrine/persistence": "^2.1",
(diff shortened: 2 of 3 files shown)


dmaicher:
Needs new releases for admin and entity-audit bundle first to be installed on CI:

```
$ composer why-not doctrine/persistence ^3.0
sonata-project/admin-bundle         4.11.1  requires  doctrine/persistence (^2.0)  
sonata-project/entity-audit-bundle  1.7.0   requires  doctrine/persistence (^2.2)

**Comment by VincentLanglet** [comment]
@dmaicher Release are done. But now the build need to be fixed

File: src/Datagrid/ProxyQuery.php [hunk 1]
```diff
@@ -177,6 +177,7 @@ public function getDoctrineQuery(): Query
             }
         }
 
+        /** @var class-string|false $rootEntity */
         $rootEntity = current($queryBuilder->getRootEntities());
```
VincentLanglet: Should we provide a PR to doctrine/orm to update the phpdoc of https://github.com/doctrine/orm/blob/2.12.x/lib/Doctrine/ORM/QueryBuilder.php#L507 then ?

jordisala1991: Probably yes but it needs some changes related to that method.

Should we go with this solution or with a supress?

VincentLanglet: A suppress would trigger a failure as soon as the pr is merged/released on orm.
So it seems better because with the phpdoc solution we will forget to remove the phpdoc.

Do you agree ?

jordisala1991: I agree, better to suppress. Can you do it @dmaicher ?

VincentLanglet:
I'm on the doctrine/orm pr btw https://github.com/doctrine/orm/pull/9778

phpstan-baseline.neon
```diff
@@ -15,3 +15,6 @@ parameters:
         - # https://github.com/doctrine/orm/pull/8767
             message: '#Method Sonata\\DoctrineORMAdminBundle\\Model\\ModelManager::getMetadata\(\) should return Doctrine\\ORM\\Mapping\\ClassMetadata\<TObject of object\> but returns Doctrine\\ORM\\Mapping\\ClassMetadata\<object\>\.$#'
             path: src/Model/ModelManager.php
+        -
```
VincentLanglet:
```suggestion
        - # https://github.com/doctrine/orm/pull/9778
```

src/Datagrid/ProxyQuery.php
jordisala1991: You will also need to supress psalm report

jordisala1991 (approved):

franmomu (approved):

PR #1683, opened 2022-08-10
Drop support for PHP 7.3

VincentLanglet:
Pedantic since it require sonataAdmin which doesnt support 7.3


jordisala1991:
Wow, we missed that one.

jordisala1991 (approved):

PR #1691, opened 2022-08-27
Update ModelManager to support Doctrine\ORM\Query directly

djpretzel:
Makes life easier when you want to set sorting, caching, etc. directly from an admin, e.g. in configureFormFields

## Subject

I am hitting use cases in admins where I want to change sort order, add caching, etc. which requires modifying the actual Doctrine query object, not the QueryBuilder, and I need to then be able to pass it in. This worked previously in 3.X because there was a default call to execute() which just happened to work on Doctrine's Query class; I think it should be explicitly supported in ModelManager, to make things easier.

I am targeting this branch, because it does not break BC.

## Changelog

```markdown
### Added
 - ModelManager::supportsQuery now supports AbstractQuery
 - ModelManager::execute now supports AbstractQuery
```

### DIFF
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -218,7 +218,11 @@ final class ModelManager implements ModelManagerInterface, LockInterface, ProxyR
             return $query->getQuery()->execute();
         }
 
-        if ($query instanceof Query || $query instanceof ProxyQuery) {
+        if ($query instanceof Query) {
+            return $query->execute();
+        }
+
+        if ($query instanceof ProxyQuery) {
             /** @phpstan-var Paginator<T> $results */
             $results = $query->execute();
 
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -18,10 +18,10 @@ use Doctrine\DBAL\Exception;
 use Doctrine\DBAL\LockMode;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
+use Doctrine\ORM\AbstractQuery;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\OptimisticLockException;
-use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
 use Doctrine\ORM\Tools\Pagination\Paginator;
 use Doctrine\ORM\UnitOfWork;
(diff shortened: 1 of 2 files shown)


src/Model/ModelManager.php
```diff
@@ -217,18 +218,19 @@ public function executeQuery(object $query)
             return $query->getQuery()->execute();
         }
 
-        if ($query instanceof ProxyQuery) {
+        if ($query instanceof Query || $query instanceof ProxyQuery) {
```
VincentLanglet: Please use a different if
```suggestion
        if ($query instanceof Query) {
            return $query->execute();
        }

        if ($query instanceof ProxyQuery) {
```

Because the Query is not returning a Paginator.

VincentLanglet (changes requested):

VincentLanglet (commented):
A case could be added for supportsQueryDataProvider
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/4.x/tests/Model/ModelManagerTest.php#L141-L154

djpretzel:
Committed change & added test - I think it should work?

VincentLanglet (approved):

tests/Model/ModelManagerTest.php
```diff
@@ -149,6 +149,7 @@ public function testSupportsQuery(bool $expected, object $object): void
     public function supportsQueryDataProvider(): iterable
     {
         yield [true, new ProxyQuery($this->createMock(QueryBuilder::class))];
+        yield [true, $this->createMock(Query::class)];
```
VincentLanglet: You forgot `use Doctrine\ORM\Query;`

VincentLanglet (changes requested):

tests/Model/ModelManagerTest.php
djpretzel: So I did! I need to get comfortable doing this via IDE instead of the github website, so I get warnings - added!

VincentLanglet (approved):

tests/Model/ModelManagerTest.php
VincentLanglet: Also with an IDE you can install the vendor and run the tests, seems like it's failing

tests/Model/ModelManagerTest.php
```diff
@@ -149,6 +150,7 @@ public function testSupportsQuery(bool $expected, object $object): void
...
```
VincentLanglet: You have to use
```
        yield [true, new Query()];
```
to avoid the error
```
1) Error
The data provider specified for Sonata\DoctrineORMAdminBundle\Tests\Model\ModelManagerTest::testSupportsQuery is invalid.
PHPUnit\Framework\MockObject\ClassIsFinalException: Class "Doctrine\ORM\Query" is declared "final" and cannot be doubled
/home/runner/work/SonataDoctrineORMAdminBundle/SonataDoctrineORMAdminBundle/tests/Model/ModelManagerTest.php:153
```

VincentLanglet (changes requested):

VincentLanglet (approved):

djpretzel:
Switched to AbstractQuery - supports NativeQuery + easier to mock

jordisala1991 (approved):

PR #1695, opened 2022-09-27
Fix batch delete on entities with json columns (PostgreSQL)

jordisala1991:
## Subject

Closes issues related with batch delete on entities with json columns. Will still need to remove the jsonb from media-bundle to avoid problem for non postgres users.

## Changelog

```markdown
### Fixed
- Fix batch delete on entities with json columns on PostgreSQL.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,7 @@
         "phpunit/phpunit": "^9.5",
         "psalm/plugin-phpunit": "^0.17",
         "psalm/plugin-symfony": "^3.0",
-        "rector/rector": "^0.13",
+        "rector/rector": "^0.14",
         "sonata-project/block-bundle": "^4.2",
         "sonata-project/entity-audit-bundle": "^1.1",
         "symfony/browser-kit": "^4.4 || ^5.4 || ^6.0",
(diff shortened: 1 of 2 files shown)


jordisala1991:
Does this fixes your issue @eerison ?

What do you think about it @VincentLanglet ? The code has the distinct since a commit 8 years ago. Can this cause trouble in some edge cases? I do not know in which cases I would want to do a distinct here.

eerison:
> Does this fixes your issue @eerison ?
> 
> What do you think about it @VincentLanglet ? The code has the distinct since a commit 8 years ago. Can this cause trouble in some edge cases? I do not know in which cases I would want to do a distinct here.

Hey @jordisala1991 I tested your Branch and it works ❤️

eerison (dismissed):

eerison (approved):

VincentLanglet (approved):

jordisala1991:
Should we release this as patch or minor? @VincentLanglet

VincentLanglet:
Maybe safer as a minor ?

PR #1700, opened 2022-10-21
Fix batch delete

### LINKED ISSUE #1699: Batch delete of over 20 entities fail
#### Sonata packages

sonata-project/doctrine-orm-admin-bundle: 4.8.0

## Subject

When using the delete batch feature of the admin with more than 20 entities it cause a "“Detached entity xxx cannot be removed".

This is cause by the clear of the entity manager after 20 iteration.

The code in 4.8.0 as change to rely on "execute" instead of "iterator". The iterator is loading one entity at the time which is make safe to clear the entity manager. 

Issue was introduce by this commit:
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/commit/35930ce8d47b73ca285f1a404ff2ca0d97d28dc8

## Minimal repository with the bug

sonata-project/doctrine-orm-admin-bundle: 4.8.0

## Steps to reproduce

Have 21 entities of one type.
Go on the admin.
Select all of them to delete by batch.
Execute.

## Expected results

Everything get deleted

## Actual results

The 20 first entity get deleted and then it fail.

jordisala1991:
## Subject

I am targeting this branch, because this is a bugfix.

Closes #1699 

## Changelog

```markdown
### Fixed
- Fix batch delete for more than 20 entities.
```


src/Model/ModelManager.php
```diff
@@ -349,7 +349,7 @@ public function batchDelete(string $class, BaseProxyQueryInterface $query): void
             $entityManager = $this->getEntityManager($class);
 
             $i = 0;
-            foreach ($query->execute() as $object) {
+            foreach ($query->toIterable() as $object) {
```
jordisala1991:
```suggestion
            foreach ($query->getQueryBuilder()->getQuery()->toIterable() as $object) {
```

src/Model/AuditReader.php
```diff
@@ -47,7 +47,7 @@ public function find(string $className, $id, $revisionId): ?object
         }
     }
 
-    public function findRevisionHistory(string $className, ?int $limit = 20, ?int $offset = 0): array
+    public function findRevisionHistory(string $className, int $limit = 20, int $offset = 0): array
```
jordisala1991: @VincentLanglet the interface does not have the "?" on both parameters. Is it a BC break to remove them here?

src/Model/ModelManager.php
```diff
@@ -349,7 +349,7 @@ public function batchDelete(string $class, BaseProxyQueryInterface $query): void
             $entityManager = $this->getEntityManager($class);
 
             $i = 0;
-            foreach ($query->execute() as $object) {
+            foreach ($query->getDoctrineQuery()->toIterable() as $object) {
```
jordisala1991, line 361: this previously was `$query->getQueryBuilder()->getQuery()->toIterable()`. The `getDoctrineQuery()` docs says its better to use it in favour of the `getQueryBuilder()->getQuery()`, is it correct @VincentLanglet ?

jordisala1991:
Not sure if this is the best option to fix the problem.

src/Model/AuditReader.php
VincentLanglet: To me it's a BC break indeed.
```
AuditReader::findRevisionHistory(Foo::class, null, null)
```
is working and won't after your PR.

src/Model/ModelManager.php
VincentLanglet: The DoctrineQuery will add the "correct" sortBy and sortOrder.
It's useless here, but won't hurt. So we can be consistent.

VincentLanglet:
image

What about doing
```
$rootAlias = current($query->getQueryBuilder()->getRootAliases());
$query->getQueryBuilder()->select($rootAlias);
```

jordisala1991:
I have added a failing test, so we can see better how to fix it.

jordisala1991:
I am trying this and it also does not work for left joined entities, maybe this has never worked? 😓 

```
            if ($query->getQueryBuilder()->getDQLPart('join')) {
                $query->getQueryBuilder()->distinct();
            }
```

Tonight or tomorrow will try to restore old code with this new test, to see if everything works or not.

jordisala1991:
> image
> 
> What about doing
> 
> ```
> $rootAlias = current($query->getQueryBuilder()->getRootAliases());
> $query->getQueryBuilder()->select($rootAlias);
> ```

Also this solution doesn't work unless you also reset the join part. BUT, that could lead to non correct behavior, since a left join might alter the list of shown items.

Maybe we need to end up doing some logic around this. 

IF query has left join -> do a non efficient iteration and removal
IF query has not left join (normal cases) -> do the efficient solution.

VincentLanglet:
> Also this solution doesn't work unless you also reset the join part. BUT, that could lead to non correct behavior, since a left join might alter the list of shown items.

Why do we also need to reset the join part ?
We can have some join, even if they are not added to the select part.

It's work here: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/4.x/src/Exporter/DataSource.php#L34

jordisala1991:
And the export has the same problem with the distinct thing. 

https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/4.x/src/Exporter/DataSource.php#L33

There are 2 problems here:
1. Distinct makes json fields unable to work on postgres unless you add jsonb => true (this is what I tried to fix at the start)
2. Left joins make impossible to add a toIterable() call

I didn't try to add distinct on your proposal since it would defeat the point 1.

jordisala1991:
We have a little problem with the functional tests on this bundle. They rely on a global database that never gets rolled back, so its shared between all tests. If I do a batch delete on one test, the next one will not have those items, and it might rely on them.

We might need to solve this issue if we want to add removal tests.

src/Model/AuditReader.php
jordisala1991: We will need to find out a way to fix phpstan then, because it is failing because it thinks the method can accept null.

tests/Functional/FetchJoinAdminTest.php
```diff
@@ -27,4 +27,28 @@ public function testCountFetchJoined(): void
...
+    public function testBatchActionFetchJoined(): void
+    {
+        $this->client->request(Request::METHOD_GET, '/admin/tests/app/author/list');
+        $this->client->submitForm('OK', [
+            'all_elements' => true,
+            'action' => 'delete',
+        ]);
+        $this->client->submitForm('Yes, execute');
+
+        self::assertResponseIsSuccessful();
```
jordisala1991: With those 2 batch delete tests we test behavior on both cases. Joined query and normal query. It still missing the json field to make postgres fail.

jordisala1991:
My last commit should fix issues with the batch delete. BUT test won't pass since I am deleting most of the database.

What to do @VincentLanglet ?
1. Try to adapt current database fill to be like on other bundles (fill data and rollback on each test)
2. Create new and independent entities to test this whole removal

I like more the 1., because we can guarantee next test won't break again in a similar way. wdyt?

VincentLanglet:
> My last commit should fix issues with the batch delete. BUT test won't pass since I am deleting most of the database.
> 
> What to do @VincentLanglet ?
> 
> 1. Try to adapt current database fill to be like on other bundles (fill data and rollback on each test)
> 2. Create new and independent entities to test this whole removal
> 
> I like more the 1., because we can guarantee next test won't break again in a similar way. wdyt?

Yes I agree. Not sure if there was a reason for not using https://github.com/dmaicher/doctrine-test-bundle

src/Model/AuditReader.php
VincentLanglet: Take a look at https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1703

tests/Functional/FetchJoinListTest.php
```diff
@@ -15,7 +15,7 @@
 
 use Symfony\Component\HttpFoundation\Request;
 
-final class FetchJoinListTest extends BasePantherTestCase
+final class FetchJoinListTest extends BaseFunctionalTestCase
```
jordisala1991:
```suggestion
final class FetchJoinListTest extends BasePantherTestCase
```

jordisala1991:
@mpoiriert , can you confirm this fixes your issue? Now we have functional test for both batch action with joined entity, and normal query. It should work as expected.

src/Model/ModelManager.php
```diff
@@ -346,10 +346,19 @@ public function batchDelete(string $class, BaseProxyQueryInterface $query): void
...
 
         try {
+            if ([] !== $query->getQueryBuilder()->getDQLPart('join')) {
+                $rootAlias = current($query->getQueryBuilder()->getRootAliases());
+
+                // Distinct is needed to iterate, even if group by is used
+                // @see https://github.com/doctrine/orm/issues/5868
+                $query->getQueryBuilder()->distinct();
+                $query->getQueryBuilder()->select($rootAlias);
+            }
```
VincentLanglet: Does it need to be in the try catch ?

The less code we have inside, the better we understand what we want to catch (?)

jordisala1991: I moved code outside the try catch.

jordisala1991:
The builds for mysql and postgres give some strange message:

```
Error:  Failed to delete object: Sonata\DoctrineORMAdminBundle\Tests\App\Entity\Author
```

Even if they do not fail, it doesn't look nice.

VincentLanglet:
> The builds for mysql and postgres give some strange message:
> 
> ```
> Error:  Failed to delete object: Sonata\DoctrineORMAdminBundle\Tests\App\Entity\Author
> ```
> 
> Even if they do not fail, it doesn't look nice.

Is there a way to improve the test ? Like checking there is a success flash bag and not an error flash bag ?

tests/App/Entity/Book.php
```diff
@@ -34,6 +34,7 @@ class Book
 
     /**
      * @ORM\ManyToOne(targetEntity="Author", inversedBy="books")
+     * @ORM\JoinColumn(name="author_id", referencedColumnName="id", onDelete="SET NULL")
```
jordisala1991, line 37: It was missing this line to avoid problems when deleting authors.

VincentLanglet: Since the test was green without this, there is something to do in order to improve the test no ? Like checking there is no 'error' flash bag.

jordisala1991: Now it should be better.

VincentLanglet (approved):

mpoiriert (approved):
I confirm that my use case is working (deleting by batch more than the flush iteration)

PR #1703, opened 2022-10-30
Fix build

VincentLanglet:
## Changelog

```markdown
### Deprecated
- Passing a null offset to `AuditReader::findRevisionHistory`
```

### DIFF
--- a/src/Model/AuditReader.php
+++ b/src/Model/AuditReader.php
@@ -47,6 +47,9 @@ final class AuditReader implements AuditReaderInterface
         }
     }
 
+    /**
+     * NEXT_MAJOR: Change the default limit value to `null`.
+     */
     public function findRevisionHistory(string $className, ?int $limit = 20, ?int $offset = 0): array
     {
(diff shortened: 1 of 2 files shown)


src/Model/AuditReader.php
```diff
@@ -47,8 +47,21 @@ public function find(string $className, $id, $revisionId): ?object
         }
     }
 
+    /**
+     * NEXT_MAJOR: Change the default limit value to `null`.
```
VincentLanglet, line 51: Consistent with https://github.com/sonata-project/SonataAdminBundle/pull/7959

jordisala1991: Why allow null? should the change be only the deprecation of null?

VincentLanglet: What if you want to get all the revision history ?

The auditReader allow to use null limit, so why not keeping the feature ?

jordisala1991: OH I see, the deprecation is for offset, and the allow null is for limit. 👍

jordisala1991 (approved):

PR #1639, opened 2022-02-28
Export not working when sorting on ManyToOne column

willemverspyck:
## Export not working when sorting on ManyToOne column

I am targeting this branch, because BC.

Closes #1586.

## Changelog

```markdown
### Fixed
Add "sortBy" entity to "select" part of DQL to fix sorting on ManyToOne column

```

Issue #1586, opened 2021-11-25
Export not working when sorting on ManyToOne column


willemverspyck:
#### Sonata packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'sonata-project/*'
sonata-project/admin-bundle              4.3.2  4.3.2  The missing Symfony Admin Generator
sonata-project/block-bundle              4.8.0  4.8.0  Symfony SonataBlockBundle
sonata-project/cache                     2.2.0  2.2.0  Cache library
sonata-project/doctrine-extensions       1.15.0 1.15.0 Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 4.2.0  4.2.0  Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/exporter                  2.9.1  2.9.1  Lightweight Exporter library
sonata-project/form-extensions           1.12.1 1.12.1 Symfony form extensions
sonata-project/twig-extensions           1.9.0  1.9.0  Sonata twig extensions
```

</p>
</details>

#### Symfony packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'symfony/*'
symfony/amqp-messenger             v5.3.7  v5.3.7  Symfony AMQP extension Messenger Bridge
symfony/apache-pack                v1.0.1  v1.0.1  A pack for Apache support in Symfony
symfony/asset                      v5.3.11 v5.3.11 Manages URL generation and versioning of web assets such as CSS stylesheets, Jav...
symfony/cache                      v5.3.12 v5.3.12 Provides an extended PSR-6, PSR-16 (and tags) implementation
symfony/cache-contracts            v2.5.0  v3.0.0  Generic abstractions related to caching
symfony/config                     v5.3.11 v5.3.11 Helps you find, load, combine, autofill and validate configuration values of any...
symfony/console                    v5.3.11 v5.3.11 Eases the creation of beautiful and testable command line interfaces
symfony/css-selector               v5.3.4  v5.3.4  Converts CSS selectors to XPath expressions
symfony/debug-bundle               v5.3.4  v5.3.4  Provides a tight integration of the Symfony Debug component into the Symfony ful...
symfony/dependency-injection       v5.3.11 v5.3.11 Allows you to standardize and centralize the way objects are constructed in your...
symfony/deprecation-contracts      v2.5.0  v3.0.0  A generic function and convention to trigger deprecation notices
[...]

PR #1705, opened 2022-11-09
Add tests for export

### LINKED ISSUE #1701: Unable to export admin list when sorted by a translated property
#### Sonata packages

<p>

```
sonata-project/admin-bundle              4.20.0 4.20.0 The missing Symfony Admin Generator
sonata-project/block-bundle              4.19.0 4.19.0 Symfony SonataBlockBundle
sonata-project/cache                     2.2.0  2.2.0  Cache library
sonata-project/doctrine-extensions       1.18.1 1.18.1 Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 4.8.0  4.8.0  Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/exporter                  2.13.0 2.13.0 Lightweight Exporter library
sonata-project/form-extensions           1.18.0 1.18.0 Symfony form extensions
sonata-project/media-bundle              4.5.1  4.5.1  Symfony SonataMediaBundle
sonata-project/twig-extensions           1.12.0 1.12.0 Sonata twig extensions
sonata-project/user-bundle               5.4.0  5.4.0  Symfony SonataUserBundle
```

</p>
</details>

#### Symfony packages

<p>

```
symfony/asset                      v4.4.46 v5.4.13 Manages URL generation and versioning of web assets such as CSS stylesheets, JavaScript files and image files
symfony/browser-kit                v4.4.44 v5.4.11 Simulates the behavior of a web browser, allowing you to make requests, click on links and submit forms programmatically
symfony/cache                      v4.4.46 v5.4.13 Provides extended PSR-6, PSR-16 (and tags) implementations
symfony/cache-contracts            v2.5.2  v2.5.2  Generic abstractions related to caching
symfony/config                     v4.4.44 v5.4.11 Helps you find, load, combine, autofill and validate configuration values of any kind
symfony/console                    v4.4.47 v5.4.14 Eases the creation of beautiful and testable command line interfaces
symfony/debug                      v4.4.44 v4.4.44 Provides tools to ease debugging PHP code
symfony/debug-bundle               v4.4.37 v5.4.11 Provides a tight integration of the Symfony VarDumper component and the ServerLogCommand from MonologBridge into the Symfony full-stack framework
symfony/dependency-injection       v4.4.44 v5.4.13 Allows you to standardize and centralize the way objects are constructed in your application
symfony/deprecation-contracts      v2.5.2  v2.5.2  A generic function and convention to trigger deprecation notices
symfony/doctrine-bridge            v4.4.46 v5.4.14 Provides integration for Doctrine with various Symfony components
symfony/dom-crawler                v4.4.45 v5.4.12 Eases DOM navigation for HTML and XML documents
symfony/error-handler              v4.4.44 v5.4.14 Provides tools to manage errors and ease debugging PHP code
symfony/event-dispatcher           v4.4.44 v5.4.9  Provides tools that allow your application components to communicate with each other by dispatching events and listening to them
symfony/event-dispatcher-contracts v1.1.13 v2.5.2  Generic abstractions related to dispatching event
symfony/expression-language        v4.4.47 v5.4.14 Provides an engine that can compile and evaluate expressions
symfony/filesystem                 v4.4.42 v5.4.13 Provides basic utilities for the filesystem
symfony/finder                     v4.4.44 v5.4.11 Finds files and directories via an intuitive fluent interface
symfony/flex                       v1.19.3 v1.19.3 Composer plugin for Symfony
symfony/form                       v4.4.46 v5.4.13 Allows to easily create, process and reuse HTML forms
symfony/framework-bundle           v4.4.47 v5.4.14 Provides a tight integration between Symfony components and the Symfony full-stack framework
symfony/http-client                v4.4.47 v5.4.14 Provides powerful methods to fetch HTTP resources synchronously or asynchronously
symfony/http-client-contracts      v2.5.2  v2.5.2  Generic abstractions related to HTTP clients
symfony/http-foundation            v4.4.47 v5.4.14 Defines an object-oriented layer for the HTTP specification
symfony/http-kernel                v4.4.47 v5.4.14 Provides a structured process for converting a Request into a Response
symfony/inflector                  v4.4.44 v5.4.11 Converts words between their singular and plural forms (English only)
symfony/intl                       v4.4.47 v5.4.14 Provides a PHP replacement layer for the C intl extension that includes additional data from the ICU library
symfony/mailer                     v4.4.46 v5.4.13 Helps sending emails
symfony/maker-bundle               v1.38.0 v1.43.0 Symfony Maker helps you create empty commands, controllers, form classes, tests and more so you can forget about writing boilerplate code.
symfony/mime                       v4.4.47 v5.4.14 Allows manipulating MIME messages
symfony/monolog-bridge             v4.4.43 v5.4.10 Provides integration for Monolog with various Symfony components
symfony/monolog-bundle             v3.8.0  v3.8.0  Symfony MonologBundle
symfony/options-resolver           v4.4.44 v5.4.11 Provides an improved replacement for the array_replace PHP function
symfony/polyfill-ctype             v1.26.0 v1.26.0 Symfony polyfill for ctype functions
symfony/polyfill-iconv             v1.26.0 v1.26.0 Symfony polyfill for the Iconv extension
symfony/polyfill-intl-grapheme     v1.26.0 v1.26.0 Symfony polyfill for intl's grapheme_* functions
symfony/polyfill-intl-icu          v1.26.0 v1.26.0 Symfony polyfill for intl's ICU-related data and classes
symfony/polyfill-intl-idn          v1.26.0 v1.26.0 Symfony polyfill for intl's idn_to_ascii and idn_to_utf8 functions
symfony/polyfill-intl-normalizer   v1.26.0 v1.26.0 Symfony polyfill for intl's Normalizer class and related functions
symfony/polyfill-mbstring          v1.26.0 v1.26.0 Symfony polyfill for the Mbstring extension
symfony/polyfill-php72             v1.26.0 v1.26.0 Symfony polyfill backporting some PHP 7.2+ features to lower PHP versions
symfony/polyfill-php73             v1.26.0 v1.26.0 Symfony polyfill backporting some PHP 7.3+ features to lower PHP versions
symfony/polyfill-php80             v1.26.0 v1.26.0 Symfony polyfill backporting some PHP 8.0+ features to lower PHP versions
symfony/polyfill-php81             v1.26.0 v1.26.0 Symfony polyfill backporting some PHP 8.1+ features to lower PHP versions
symfony/process                    v4.4.44 v5.4.11 Executes commands in sub-processes
symfony/property-access            v4.4.44 v5.4.11 Provides functions to read and write from/to an object or array using a simple string notation
symfony/routing                    v4.4.44 v5.4.11 Maps an HTTP request to a set of configuration variables
symfony/security-acl               v3.3.2  v3.3.2  Symfony Security Component - ACL (Access Control List)
symfony/security-bundle            v4.4.44 v5.4.11 Provides a tight integration of the Security component into the Symfony full-stack framework
symfony/security-core              v4.4.47 v5.4.14 Symfony Security Component - Core Library
symfony/security-csrf              v4.4.37 v5.4.11 Symfony Security Component - CSRF Library
symfony/security-guard             v4.4.46 v5.4.13 Symfony Security Component - Guard
symfony/security-http              v4.4.44 v5.4.13 Symfony Security Component - HTTP Integration
symfony/serializer                 v4.4.47 v5.4.14 Handles serializing and deserializing data structures, including object graphs, into array structures or other formats like XML and JSON.
symfony/service-contracts          v2.5.2  v2.5.2  Generic abstractions related to writing services
symfony/stopwatch                  v4.4.46 v5.4.13 Provides a way to profile code
symfony/string                     v5.4.14 v5.4.14 Provides an object-oriented API to strings and deals with bytes, UTF-8 code points and grapheme clusters in a unified way
symfony/swiftmailer-bundle         v3.5.4  v3.5.4  Symfony SwiftmailerBundle
Package symfony/swiftmailer-bundle is abandoned, you should avoid using it. Use symfony/mailer instead.
symfony/translation                v4.4.47 v5.4.14 Provides tools to internationalize your application
symfony/translation-contracts      v2.5.2  v2.5.2  Generic abstractions related to translation
symfony/twig-bridge                v4.4.45 v5.4.14 Provides integration for Twig with various Symfony components
symfony/twig-bundle                v4.4.41 v5.4.8  Provides a tight integration of Twig into the Symfony full-stack framework
symfony/validator                  v4.4.47 v5.4.14 Provides tools to validate values
symfony/var-dumper                 v4.4.47 v5.4.14 Provides mechanisms for walking through any arbitrary PHP variable
symfony/var-exporter               v4.4.43 v5.4.10 Allows exporting any serializable PHP data structure to plain PHP code
symfony/web-profiler-bundle        v4.4.47 v5.4.14 Provides a development tool that gives detailed information about the execution of any request
symfony/yaml                       v4.4.45 v5.4.14 Loads and dumps YAML files
```

</p>
</details>

#### PHP version

```
$ php -v
PHP 8.1.3 (cli) (built: Feb 18 2022 19:37:16) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.3, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.3, Copyright (c), by Zend Technologies
    with Xdebug v3.2.0RC1, Copyright (c) 2002-2022, by Derick Rethans
```

## Subject

In #1639 a fix was included in `Sonata\DoctrineORMAdminBundle\Exporter\DataSource.php` for exporting when sorting on ManyToOne column that breaks exporting when a column is from a translation entity and is sorted by it.

With this change the export fails with the following error when trying to export an entity `Entity` associated with its translation `EntityTranslation`

```
Iterate with fetch join in class AppBundle\Entity\EntityTranslation using association translatable not allowed.
```
Before this change the export would work although it was not exported sorted by the property.

jordisala1991:
## Subject

I am targeting this branch, because this is fixing a bug.

Closes #1701 .

## Changelog

```markdown
### Fixed
- Fix export for admins sorted on a property that belongs to a one to many.
```

### DIFF
--- a/src/Datagrid/ProxyQuery.php
+++ b/src/Datagrid/ProxyQuery.php
@@ -196,11 +196,14 @@ final class ProxyQuery implements ProxyQueryInterface
 
     public function setSortBy(array $parentAssociationMappings, array $fieldMapping): BaseProxyQueryInterface
     {
-        $alias = $this->entityJoin($parentAssociationMappings);
+        if (!isset($fieldMapping['fieldName'])) {
+            $this->sortBy = null;
 
-        $this->sortBy = isset($fieldMapping['fieldName']) ?
-            $alias.'.'.$fieldMapping['fieldName'] :
-            null;
+            return;
+        }
+
+        $alias = $this->entityJoin($parentAssociationMappings);
+        $this->sortBy = $alias.'.'.$fieldMapping['fieldName'];
 
         return $this;
     }
--- a/src/Datagrid/ProxyQuery.php
+++ b/src/Datagrid/ProxyQuery.php
@@ -199,7 +199,7 @@ final class ProxyQuery implements ProxyQueryInterface
         if (!isset($fieldMapping['fieldName'])) {
             $this->sortBy = null;
 
-            return;
+            return $this;
         }
 
         $alias = $this->entityJoin($parentAssociationMappings);
--- a/src/Datagrid/ProxyQuery.php
+++ b/src/Datagrid/ProxyQuery.php
@@ -197,7 +197,10 @@ final class ProxyQuery implements ProxyQueryInterface
     public function setSortBy(array $parentAssociationMappings, array $fieldMapping): BaseProxyQueryInterface
     {
         $alias = $this->entityJoin($parentAssociationMappings);
-        $this->sortBy = $alias.'.'.$fieldMapping['fieldName'];
+
+        $this->sortBy = isset($fieldMapping['fieldName']) ?
+            $alias.'.'.$fieldMapping['fieldName'] :
+            null;
 
         return $this;
     }
--- a/src/Exporter/DataSource.php
+++ b/src/Exporter/DataSource.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\DoctrineORMAdminBundle\Exporter;
 
+use Doctrine\ORM\Mapping\ClassMetadata;
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface as BaseProxyQueryInterface;
 use Sonata\AdminBundle\Exporter\DataSourceInterface;
 use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface;
@@ -36,12 +37,16 @@ final class DataSource implements DataSourceInterface
         $sortBy = $query->getSortBy();
 
         // AddSelect is needed when exporting the results sorted by a column that is part of ManyToOne relation
+        // For OneToMany the toIterable() doctrine method is not supported so the select is not added and the sort is removed.
+        //
         // @see https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1586
         if (null !== $sortBy) {
             $rootAliasSortBy = strstr($sortBy, '.', true);
 
-            if ($rootAliasSortBy !== $rootAlias) {
-                $query->getQueryBuilder()->addSelect($rootAliasSortBy);
+            if (false !== $rootAliasSortBy && $rootAliasSortBy !== $rootAlias) {
+                $this->isOneToMany($query, $rootAliasSortBy) ?
+                    $query->setSortBy([], []) :
+                    $query->getQueryBuilder()->addSelect($rootAliasSortBy);
             }
         }
 
@@ -49,5 +54,54 @@ final class DataSource implements DataSourceInterface
         $query->setMaxResults(null);
 
         return new DoctrineORMQuerySourceIterator($query->getDoctrineQuery(), $fields);
+    }
+
+    /**
+     * @param ProxyQueryInterface<object> $query
+     */
+    private function isOneToMany(ProxyQueryInterface $query, string $alias): bool
+    {
+        $fieldName = $this->findFieldName($query, $alias);
+
+        if (null === $fieldName) {
+            return false;
+        }
+
+        $rootEntity = current($query->getQueryBuilder()->getRootEntities());
+
+        if (false === $rootEntity) {
+            return false;
+        }
+
+        $associationMapping = $query
+            ->getQueryBuilder()
+            ->getEntityManager()
+            ->getClassMetadata($rootEntity)
+            ->getAssociationMapping($fieldName);
+
+        return ClassMetadata::ONE_TO_MANY === $associationMapping['type'];
+    }
+
+    /**
+     * @param ProxyQueryInterface<object> $query
+     */
+    private function findFieldName(ProxyQueryInterface $query, string $alias): ?string
+    {
+        $joins = $query->getQueryBuilder()->getDQLPart('join');
+
+        foreach ($joins as $joined) {
+            foreach ($joined as $joinPart) {
+                $join = $joinPart->getJoin();
+                $joinAlias = $joinPart->getAlias();
+
+                if ($joinAlias === $alias) {
+                    $joinParts = explode('.', $join);
+
+                    return end($joinParts);
+                }
+            }
+        }
+
+        return null;
     }
 }
--- a/tests/App/Admin/AuthorAdmin.php
+++ b/tests/App/Admin/AuthorAdmin.php
@@ -39,6 +39,15 @@ class AuthorAdmin extends AbstractAdmin
             ])
             ->add('numberOfReaders', FieldDescriptionInterface::TYPE_INTEGER, [
                 'template' => 'author/list_number_of_readers_field.html.twig',
+            ])
+            ->addIdentifier('firstBook', null, [
+                'sortable' => true,
+                'sort_field_mapping' => [
+                    'fieldName' => 'name',
+                ],
+                'sort_parent_association_mappings' => [[
+                    'fieldName' => 'books',
+                ]],
             ]);
     }
 
--- a/tests/Functional/ExportActionsTest.php
+++ b/tests/Functional/ExportActionsTest.php
@@ -25,6 +25,7 @@ final class ExportActionsTest extends BaseFunctionalTestCase
      */
     public function testExportAction(string $url, array $parameters, array $expected): void
     {
+        // TODO: Remove the $content variable when drop support for Symfony < 6.2
         ob_start();
         $this->client->request(Request::METHOD_GET, $url, $parameters);
         $content = ob_get_contents();
--- a/tests/Functional/ExportActionsTest.php
+++ b/tests/Functional/ExportActionsTest.php
@@ -1,0 +1,119 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DoctrineORMAdminBundle\Tests\Functional;
+
+use Symfony\Component\HttpFoundation\Request;
+
+final class ExportActionsTest extends BaseFunctionalTestCase
+{
+    /**
+     * @dataProvider provideExportActionCases
+     *
+     * @param array<mixed> $parameters
+     * @param array<mixed> $expected
+     */
+    public function testExportAction(string $url, array $parameters, array $expected): void
+    {
+        ob_start();
+        $this->client->request(Request::METHOD_GET, $url, $parameters);
+        $content = ob_get_contents();
+        ob_end_clean();
+
+        self::assertResponseIsSuccessful();
+        static::assertSame(
+            $expected,
+            json_decode(
+                '' !== $content && false !== $content ? $content : $this->client->getInternalResponse()->getContent(),
+                true,
+                512,
+                \JSON_THROW_ON_ERROR
+            )
+        );
+    }
+
+    /**
+     * @return iterable<array<string>>
+     *
+     * @phpstan-return iterable<array{0: string, 1: array<mixed>, 2: array<mixed>}>
+     */
+    public static function provideExportActionCases(): iterable
+    {
+        yield 'Normal export' => ['/admin/tests/app/book/export', [
+            'format' => 'json',
+        ], [
+            ['Id' => 'book_1', 'Name' => 'Book 1'],
+            ['Id' => 'book_2', 'Name' => 'Book 2'],
+            ['Id' => 'book_id', 'Name' => 'Don Quixote'],
+        ]];
+
+        yield 'Normal export with normal sort' => ['/admin/tests/app/book/export', [
+            'format' => 'json',
+            'filter' => [
+                '_sort_by' => 'name',
+                '_sort_order' => 'DESC',
+            ],
+        ], [
+            ['Id' => 'book_id', 'Name' => 'Don Quixote'],
+            ['Id' => 'book_2', 'Name' => 'Book 2'],
+            ['Id' => 'book_1', 'Name' => 'Book 1'],
+        ]];
+
+        yield 'Joined export' => ['/admin/tests/app/author/export', [
+            'format' => 'json',
+        ], [
+            ['Id' => 'anonymous', 'Name' => 'Anonymous', 'Address Street' => ''],
+            ['Id' => 'author_with_two_books', 'Name' => 'Author with 2 books', 'Address Street' => ''],
+            ['Id' => 'autocompletion_author', 'Name' => 'autocompletion author', 'Address Street' => ''],
+            ['Id' => 'miguel_de_cervantes', 'Name' => 'Miguel de Cervantes', 'Address Street' => 'Somewhere in La Mancha, in a place whose name I do not care to remember'],
+        ]];
+
+        yield 'Joined export with normal sort' => ['/admin/tests/app/author/export', [
+            'format' => 'json',
+            'filter' => [
+                '_sort_by' => 'id',
+                '_sort_order' => 'DESC',
+            ],
+        ], [
+            ['Id' => 'miguel_de_cervantes', 'Name' => 'Miguel de Cervantes', 'Address Street' => 'Somewhere in La Mancha, in a place whose name I do not care to remember'],
+            ['Id' => 'autocompletion_author', 'Name' => 'autocompletion author', 'Address Street' => ''],
+            ['Id' => 'author_with_two_books', 'Name' => 'Author with 2 books', 'Address Street' => ''],
+            ['Id' => 'anonymous', 'Name' => 'Anonymous', 'Address Street' => ''],
+        ]];
+
+        yield 'Joined export with oneToMany sort' => ['/admin/tests/app/author/export', [
+            'format' => 'json',
+            'filter' => [
+                '_sort_by' => 'firstBook',
+                '_sort_order' => 'DESC',
+            ],
+        ], [
+            ['Id' => 'miguel_de_cervantes', 'Name' => 'Miguel de Cervantes', 'Address Street' => 'Somewhere in La Mancha, in a place whose name I do not care to remember'],
+            ['Id' => 'autocompletion_author', 'Name' => 'autocompletion author', 'Address Street' => ''],
+            ['Id' => 'author_with_two_books', 'Name' => 'Author with 2 books', 'Address Street' => ''],
+            ['Id' => 'anonymous', 'Name' => 'Anonymous', 'Address Street' => ''],
+        ]];
+
+        yield 'Joined export with ManyToOne sort' => ['/admin/tests/app/item/export', [
+            'format' => 'json',
+            'filter' => [
+                '_sort_by' => 'product.name',
+                '_sort_order' => 'DESC',
+            ],
+        ], [
+            ['Offered Price' => '3'],
+            ['Offered Price' => '2'],
+            ['Offered Price' => '2'],
+        ]];
+    }
+}
--- a/tests/App/Admin/ItemAdmin.php
+++ b/tests/App/Admin/ItemAdmin.php
@@ -25,8 +25,24 @@ final class ItemAdmin extends AbstractAdmin
     protected function configureListFields(ListMapper $list): void
     {
         $list
-            ->add('command')
-            ->add('product')
+            ->add('command.id', null, [
+                'sortable' => true,
+                'sort_field_mapping' => [
+                    'fieldName' => 'id',
+                ],
+                'sort_parent_association_mappings' => [[
+                    'fieldName' => 'command',
+                ]],
+            ])
+            ->add('product.name', null, [
+                'sortable' => true,
+                'sort_field_mapping' => [
+                    'fieldName' => 'name',
+                ],
+                'sort_parent_association_mappings' => [[
+                    'fieldName' => 'product',
+                ]],
+            ])
             ->add('offeredPrice');
     }
 }
--- a/tests/App/AppKernel.php
+++ b/tests/App/AppKernel.php
@@ -22,6 +22,7 @@ use Sonata\BlockBundle\Cache\HttpCacheHandler;
 use Sonata\BlockBundle\SonataBlockBundle;
 use Sonata\Doctrine\Bridge\Symfony\SonataDoctrineBundle;
 use Sonata\DoctrineORMAdminBundle\SonataDoctrineORMAdminBundle;
+use Sonata\Exporter\Bridge\Symfony\SonataExporterBundle;
 use Sonata\Form\Bridge\Symfony\SonataFormBundle;
 use Sonata\Twig\Bridge\Symfony\SonataTwigBundle;
 use Symfony\Bundle\FrameworkBundle\FrameworkBundle;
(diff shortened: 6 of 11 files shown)


jordisala1991:
For some reason it does not work on Github actions on certain cases. It works on local dev env. Do you have any idea @VincentLanglet ?

VincentLanglet:
Works fine on my local too...

The test output is weird
image

The json data is "dumped" instead of being returned...

jordisala1991:
I don't know how to fix this, I have been investigating and this is the only I found: https://github.com/symfony/symfony/issues/25005

It doesn't work either.

tests/Functional/ExportActionsTest.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        ob_start();
+        $this->client->request(Request::METHOD_GET, $url, $parameters);
+        $content = ob_get_contents();
+        ob_end_clean();
+
+        self::assertResponseIsSuccessful();
+        static::assertSame(
+            json_encode($expected, \JSON_THROW_ON_ERROR),
+            '' !== $content ? $content : $this->client->getInternalResponse()->getContent()
```
jordisala1991: have to do this hack, depends on the php version it works with the ob_get_contents or not.

VincentLanglet: Can you add a comment to explain the php version we need `ob_get_contents` and the one we don't need it ?
This way we'll know when to remove the hack.

tests/App/Admin/AuthorAdmin.php
```diff
@@ -71,6 +80,16 @@ protected function configureFormFields(FormMapper $form): void
             ]);
     }
 
+    protected function configureExportFields(): array
+    {
+        return [
+            'id',
```
core23: Maybe you can add a test for virtual fields like `count(address.id) as addressCount`.

jordisala1991: Im trying to debug: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1701

Hopefully I will find the bug and fix it, and this could be the foundation for future tests that ensure things related with exports works correctly.

jordisala1991:
So I added test for all the working exports, including the joined with many to one sort (which is working).

Next commit will add the one to many that is not working, and then we will try to fix all of them.

src/Exporter/DataSource.php
```diff
@@ -33,18 +33,6 @@ public function createIterator(BaseProxyQueryInterface $query, array $fields): \
...
-
-        // AddSelect is needed when exporting the results sorted by a column that is part of ManyToOne relation
-        // @see https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1586
-        if (null !== $sortBy) {
-            $rootAliasSortBy = strstr($sortBy, '.', true);
-
-            if ($rootAliasSortBy !== $rootAlias) {
-                $query->getQueryBuilder()->addSelect($rootAliasSortBy);
-            }
-        }
```
VincentLanglet: Since we remove the fix for https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1586, 
can you try this branch @willemverspyck ?

jordisala1991: nope, don't try it , test don't even pass. I don't know how to fix both errors. I would need to know on the DataSource if the sort is not a many to one or a one to one so it should be removed.

VincentLanglet: So it works for sort by OneToOne but not OneToMany ?
The question would be why in the issue https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1701 there is a sort on translation (which is OneToMany if I understand).

Sure we could try to remove it dynamically, but shouldn't it be added in the first time then ?
Or is this "OneToMany sort" working in some other situations ?

jordisala1991: It is working on the list, but not on the export, because on the export we do a distinct and we iterate over the results. On the list we execute the query normally

jordisala1991: I think we might have 2 options:
1. Try to know the cardinality of the sort by, so we can avoid adding it if one to many.
2. Try to know if it can be fixed somehow here: https://github.com/sonata-project/exporter/blob/3.x/src/Source/DoctrineORMQuerySourceIterator.php#L56-L60

I don't think 2 is possible, so im going for 1 at the moment, it might take a while. Test are red, so if you want to try to fix it would be helpful, I don't have a lot of ideas on how to fix this.

tests/Functional/ExportActionsTest.php
jordisala1991: Yes, I will add a reminder here, I think it starts changing behavior on php 8.0, but not completely sure, will have to try.

src/Exporter/DataSource.php
VincentLanglet: I would say it will require to access the metadata like we did here https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/4.x/src/FieldDescription/FieldDescriptionFactory.php

In the `getDoctrineQuery` we're doing
```
$rootEntity = current($queryBuilder->getRootEntities());
        if (false === $rootEntity) {
            throw new \RuntimeException('There are not root entities defined in the query.');
        }

        $identifierFields = $queryBuilder
            ->getEntityManager()
            ->getMetadataFactory()
            ->getMetadataFor($rootEntity)
            ->getIdentifierFieldNames();
```

We might need to play again with 
```
$queryBuilder
            ->getEntityManager()
            ->getMetadataFactory()
            ->getMetadataFor()
```

jordisala1991:
I think Im somewhat close to fixing this: Now locally with mysql database I no longer get errors.

src/Exporter/DataSource.php
```diff
@@ -50,4 +55,41 @@ public function createIterator(BaseProxyQueryInterface $query, array $fields): \
...
+            return false;
+        }
+
+        $associationMapping = $query
+            ->getQuerybuilder()
+            ->getEntityManager()
+            ->getClassMetadata(current($query->getQueryBuilder()->getRootEntities()))
+            ->getAssociationMapping($fieldName);
+
+        return $associationMapping['type'] === ClassMetadata::ONE_TO_MANY;
```
jordisala1991:
```suggestion
        return ClassMetadata::ONE_TO_MANY === $associationMapping['type'];
```

jordisala1991:
@Cenadros and @willemverspyck, can you check if this PR solves your export use cases?

src/Datagrid/ProxyQuery.php
```diff
@@ -202,6 +202,13 @@ public function setSortBy(array $parentAssociationMappings, array $fieldMapping)
         return $this;
     }
 
+    public function resetSortBy(): BaseProxyQueryInterface
```
VincentLanglet: Do you plan to add this to the ProxyQueryInterface interface ? @jordisala1991

jordisala1991: I guess we will need to... unless there is another way to accomplish the same effect. getDoctrineQuery uses the sortBy to add more sorts to the final query. Is there another workaround that avoids this new method?

VincentLanglet: I see some issues with this:
- If we're introducing it to SonataAdmin/ProxyQueryInterface, this is a method which is not needed for other ORM
- If we're introducing it to SonataORM/ProxyQueryInterface, it just makes the interface bigger and bigger (See https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/4.x/src/Datagrid/ProxyQueryInterface.php vs https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/blob/4.x/src/Datagrid/ProxyQueryInterface.php)
- If we introduce this, there should be a `resetSortOrder` too to be consistent
- One concern is the fact that `$query->setOrderBy(...)->resetOrderBy()` is doing something under the hood: It add a LEFT JOIN in the query https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/4.x/src/Datagrid/ProxyQuery.php#L199 when you could believe the reset does cancel the full setOrderBy call.

The only other option I see would be to change the `getDoctrineQuery` call depends on the reason we have to remove the sort by:
- If this is a temporary solution until some support is added to Doctrine, we could use `$query->getQueryBuild()->setSortBy(null)->getQuery()` in the case of OneToMany...
- If this is a long-term solution cause no support will be added we could consider changing `getDoctrineQuery` to `getDoctrineQuery(bool $withSort = true)`.

Also, why I don't understand between this PR and the issue https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1701 is the fact it was said that before this fix (https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1639) it was working. But the quoted fix didn't add any sortBy, just a select. So why the `$query->resetSortBy()` is needed (or was it really working before) ?

jordisala1991: About the "how to deal with the sort by" I will think about all the points you mentioned.

And about the why is this needed. I will introduce a commit reverting the fix, so the test fail and we can see the error message (just in case there is a better fix). The issue is present only on one to many. And I think the test cover all possible sort by options (property on the class, relation oto, otm, mtm, mto)

src/Exporter/DataSource.php
```diff
@@ -36,12 +37,16 @@ public function createIterator(BaseProxyQueryInterface $query, array $fields): \
...
+        // For OneToMany the toIterable() doctrine method is not supported so the sorting is removed
+        //
         // @see https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1586
         if (null !== $sortBy) {
             $rootAliasSortBy = strstr($sortBy, '.', true);
 
             if ($rootAliasSortBy !== $rootAlias) {
-                $query->getQueryBuilder()->addSelect($rootAliasSortBy);
+                $this->isOneToMany($query, $rootAliasSortBy) ?
+                    $query->resetSortBy() :
```
jordisala1991:
```suggestion
                    // $query->resetSortBy() :
                    true :
```

src/Datagrid/ProxyQuery.php
VincentLanglet: > And about the why is this needed. I will introduce a commit reverting the fix, so the test fail and we can see the error message (just in case there is a better fix). The issue is present only on one to many. And I think the test cover all possible sort by options (property on the class, relation oto, otm, mtm, mto)

Thanks.

one thing to try would be
```
if ($rootAliasSortBy !== $rootAlias && !$this->isOneToMany($query, $rootAliasSortBy)) {
    $query->getQueryBuilder()->addSelect($rootAliasSortBy);
}
```
According do the message https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1701#issuecomment-1293203154 with the working and not working query.

jordisala1991: Thats what I am trying now, seems that removing the sort is not needed, just not add the select.

src/Exporter/DataSource.php
```diff
@@ -36,11 +37,13 @@ public function createIterator(BaseProxyQueryInterface $query, array $fields): \
...
 
         // AddSelect is needed when exporting the results sorted by a column that is part of ManyToOne relation
+        // For OneToMany the toIterable() doctrine method is not supported so the select is not added.
+        //
         // @see https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1586
         if (null !== $sortBy) {
             $rootAliasSortBy = strstr($sortBy, '.', true);
 
-            if ($rootAliasSortBy !== $rootAlias) {
+            if ($rootAliasSortBy !== $rootAlias && $this->isOneToMany($query, $rootAliasSortBy)) {
```
jordisala1991:
```suggestion
            if ($rootAliasSortBy !== $rootAlias && !$this->isOneToMany($query, $rootAliasSortBy)) {
```

tests/Functional/ExportActionsTest.php
```diff
@@ -0,0 +1,119 @@
...
+        yield 'Joined export with oneToMany sort' => ['/admin/tests/app/author/export', [
+            'format' => 'json',
+            'filter' => [
+                '_sort_by' => 'firstBook',
+                '_sort_order' => 'DESC',
+            ],
+        ], [
+            ['Id' => 'miguel_de_cervantes', 'Name' => 'Miguel de Cervantes', 'Address Street' => 'Somewhere in La Mancha, in a place whose name I do not care to remember'],
+            ['Id' => 'autocompletion_author', 'Name' => 'autocompletion author', 'Address Street' => ''],
+            ['Id' => 'author_with_two_books', 'Name' => 'Author with 2 books', 'Address Street' => ''],
```
jordisala1991, line 104:
```suggestion
            ['Id' => 'author_with_two_books', 'Name' => 'Author with 2 books', 'Address Street' => ''],
            ['Id' => 'autocompletion_author', 'Name' => 'autocompletion author', 'Address Street' => ''],
```

jordisala1991:
See the error now @VincentLanglet . If I dont remove the sort by, the mysql build does not pass. I will have to go back to the initial solution.

VincentLanglet:
> See the error now @VincentLanglet . If I dont remove the sort by, the mysql build does not pass. I will have to go back to the initial solution.

Then I don't understand why
```
SELECT DISTINCT o FROM AppBundle\Entity\Contact o LEFT JOIN o.translations s_translations WHERE o.type = :type ORDER BY s_translations.name ASC, o.id ASC
```
is listed as a working query in https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1701#issuecomment-1293203154

jordisala1991:
> > See the error now @VincentLanglet . If I dont remove the sort by, the mysql build does not pass. I will have to go back to the initial solution.
> 
> 
> 
> Then I don't understand why
> 
> ```
> 
> SELECT DISTINCT o FROM AppBundle\Entity\Contact o LEFT JOIN o.translations s_translations WHERE o.type = :type ORDER BY s_translations.name ASC, o.id ASC
> 
> ```
> 
> is listed as a working query in https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/1701#issuecomment-1293203154

It might depend on the mysql version? He is using mysql 5.6.

src/Exporter/DataSource.php
```diff
@@ -36,12 +37,16 @@ public function createIterator(BaseProxyQueryInterface $query, array $fields): \
         $sortBy = $query->getSortBy();
 
         // AddSelect is needed when exporting the results sorted by a column that is part of ManyToOne relation
+        // For OneToMany the toIterable() doctrine method is not supported so the select is not added.
```
jordisala1991:
```suggestion
        // For OneToMany the toIterable() doctrine method is not supported so the select is not added and the sort is removed.
```

src/Datagrid/ProxyQuery.php
```diff
@@ -202,6 +202,13 @@ public function setSortBy(array $parentAssociationMappings, array $fieldMapping)
         return $this;
     }
 
+    public function resetSortBy(): void
```
jordisala1991:
```suggestion
    public function resetSortBy(): BaseProxyQueryInterface
```

tests/Functional/ExportActionsTest.php
```diff
@@ -0,0 +1,119 @@
...
+        yield 'Joined export with oneToMany sort' => ['/admin/tests/app/author/export', [
+            'format' => 'json',
+            'filter' => [
+                '_sort_by' => 'firstBook',
+                '_sort_order' => 'DESC',
+            ],
+        ], [
+            ['Id' => 'miguel_de_cervantes', 'Name' => 'Miguel de Cervantes', 'Address Street' => 'Somewhere in La Mancha, in a place whose name I do not care to remember'],
+            ['Id' => 'author_with_two_books', 'Name' => 'Author with 2 books', 'Address Street' => ''],
+            ['Id' => 'autocompletion_author', 'Name' => 'autocompletion author', 'Address Street' => ''],
```
jordisala1991:
```suggestion
            ['Id' => 'autocompletion_author', 'Name' => 'autocompletion author', 'Address Street' => ''],
            ['Id' => 'author_with_two_books', 'Name' => 'Author with 2 books', 'Address Street' => ''],
```

willemverspyck:
@jordisala1991: I tried your PR and that still works for me. Thanks!

src/Datagrid/ProxyQuery.php
jordisala1991: I tried another approach, lets see if it is good enough

src/Datagrid/ProxyQuery.php
```diff
@@ -197,7 +197,10 @@ public function getDoctrineQuery(): Query
     public function setSortBy(array $parentAssociationMappings, array $fieldMapping): BaseProxyQueryInterface
     {
         $alias = $this->entityJoin($parentAssociationMappings);
-        $this->sortBy = $alias.'.'.$fieldMapping['fieldName'];
+
+        $this->sortBy = isset($fieldMapping['fieldName']) ?
```
jordisala1991: if we do like this, I can call setSortBy([], []) to reset the sort (It still won't reset the entityJoin, that would require a different approach). TBH I am not sure why a sortBy does and entityJoin that is not easy to undo under the hood.

WDYT about this approach @VincentLanglet ?

jordisala1991:
I have split things in 2 commits, the first one showing the failing tests, and the second one fixing them.

tests/Functional/ExportActionsTest.php
jordisala1991: I have added a note about this, the problem was Symfony < 6.2.

src/Datagrid/ProxyQuery.php
VincentLanglet: This seems a good idea.

I would slighty change the implementation to
```
if (!isset($fieldMapping['fieldName'])) {
    $this->sortBy = null;
    return $this;
}

$alias = $this->entityJoin($parentAssociationMappings);
$this->sortBy = $alias.'.'.$fieldMapping['fieldName'];
return $this;
```
and a test could be nice.

For the entityJoin, we have the call
```
$this->query->setSortBy(
            $this->values[DatagridInterface::SORT_BY]->getSortParentAssociationMapping(),
            $this->values[DatagridInterface::SORT_BY]->getSortFieldMapping()
        );
```
in SonataAdmin, and the entityJoin is not in the interface, so it has to be in the sortby, no ?

jordisala1991: Changed the implementation, will see if I can add a unit test.

About the join, yes it can be applied here, the difference with the sortBy is that the sortBy is only saved to the private property and applied when you getDoctrineQuery()

But I understand why, just a bit clunky, could be improved.

src/Datagrid/ProxyQuery.php
```diff
@@ -196,11 +196,14 @@ public function getDoctrineQuery(): Query
 
     public function setSortBy(array $parentAssociationMappings, array $fieldMapping): BaseProxyQueryInterface
     {
-        $alias = $this->entityJoin($parentAssociationMappings);
+        if (!isset($fieldMapping['fieldName'])) {
+            $this->sortBy = null;
+
+            return;
```
VincentLanglet:
```suggestion
            return $this;
```

src/Datagrid/ProxyQuery.php
jordisala1991: Added test to it.

jordisala1991:
IMO the PR is ready. It adds a lot of tests on the export topic which we had none before, also fixes the oneToMany sort problem.

Besides that, would also be nice @Cenadros input on whether it fixes his issue or not.

VincentLanglet (approved):

PR #1734, opened 2023-04-01
Improve exception message at `ModelManager::batchDelete()`

### LINKED ISSUE #1735: Add tests for `ModelManager::batchDelete()`
This method requires tests. Since [its addition](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/commit/208130ee41900d157a57651585938482ceb491b3#diff-89e73ce5e4dad6ecf6979467d8da4288b7931cfe53ad1885e69d662700fe74d6R300), there were no tests for it.

phansys:
## Subject

Related to https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/pull/811#discussion_r1155126136.

Closes #1735.

## Changelog

```markdown
### Changed
- Exception message at `ModelManager::batchDelete()` in order to provide more details about the failed batch operation
```

### DIFF
--- a/src/Model/ModelManager.php
+++ b/src/Model/ModelManager.php
@@ -45,6 +45,8 @@ final class ModelManager implements ModelManagerInterface, LockInterface, ProxyR
 {
     public const ID_SEPARATOR = '~';
 
+    private const BATCH_SIZE = 20;
+
     /**
      * @var EntityManagerInterface[]
      */
@@ -345,7 +347,7 @@ final class ModelManager implements ModelManagerInterface, LockInterface, ProxyR
         $qb->andWhere(sprintf('( %s )', implode(' OR ', $sqls)));
     }
 
-    public function batchDelete(string $class, BaseProxyQueryInterface $query): void
+    public function batchDelete(string $class, BaseProxyQueryInterface $query, int $batchSize = self::BATCH_SIZE): void
     {
         if (!$query instanceof ProxyQueryInterface) {
             throw new \TypeError(sprintf('The query MUST implement %s.', ProxyQueryInterface::class));
@@ -362,13 +364,15 @@ final class ModelManager implements ModelManagerInterface, LockInterface, ProxyR
 
         $entityManager = $this->getEntityManager($class);
         $i = 0;
+        $confirmedDeletionsCount = 0;
 
         try {
             foreach ($query->getDoctrineQuery()->toIterable() as $object) {
                 $entityManager->remove($object);
 
-                if (0 === (++$i % 20)) {
+                if (0 === (++$i % $batchSize)) {
                     $entityManager->flush();
+                    $confirmedDeletionsCount = $i;
                     $entityManager->clear();
                 }
             }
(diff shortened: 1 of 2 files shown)


src/Model/ModelManager.php
```diff
@@ -45,6 +45,8 @@ final class ModelManager implements ModelManagerInterface, LockInterface, ProxyR
 {
     public const ID_SEPARATOR = '~';
 
+    private const BATCH_SIZE = 20;
```
VincentLanglet, line 48: Could be useful to introduce the size as parameter
```
public function batchDelete(string $class, BaseProxyQueryInterface $query, int $batchSize = self::DEFAULT_BATCH_SIZE): void
```

src/Model/ModelManager.php
```diff
@@ -345,12 +347,16 @@ public function addIdentifiersToQuery(string $class, BaseProxyQueryInterface $qu
...
     }
 
-    public function batchDelete(string $class, BaseProxyQueryInterface $query): void
+    public function batchDelete(string $class, BaseProxyQueryInterface $query/* , int $batchSize = self::BATCH_SIZE */): void
     {
         if (!$query instanceof ProxyQueryInterface) {
             throw new \TypeError(sprintf('The query MUST implement %s.', ProxyQueryInterface::class));
         }
 
+        // NEXT_MAJOR: Remove this assignment.
```
VincentLanglet: You don't need to wait next major.

https://onlinephp.io/c/4b5209

VincentLanglet (changes requested):

src/Model/ModelManager.php
phansys: I forgot our classes are (finally :laughing:) final now :rocket:

VincentLanglet (approved):

jordisala1991 (approved):
Could you apply the same for mongoDB?

src/Model/ModelManager.php
```diff
@@ -345,7 +347,7 @@ public function addIdentifiersToQuery(string $class, BaseProxyQueryInterface $qu
         $qb->andWhere(sprintf('( %s )', implode(' OR ', $sqls)));
     }
 
-    public function batchDelete(string $class, BaseProxyQueryInterface $query): void
+    public function batchDelete(string $class, BaseProxyQueryInterface $query, int $batchSize = self::BATCH_SIZE): void
```
jordisala1991, line 350: shouldn't we add this to the interface? or at least with a next major comment?

VincentLanglet: We cant add this to the interface since it wont be bc but yes we can add the next major constant and introduce the constant in the SonataAdminBundle instead

phansys: I was thinking about the constant in `SonataAdmin`, but I think this must be kept in the scope of each model manager (the behavior regarding the batch size, the way each driver handles the transactions and the cost it demands is specific and not a common concern).

VincentLanglet: If we can't add the constant to the SonataAdmin interface, we can't add the batchSize parameter to the interface too.
I'm fine with both solution.

jordisala1991: Then what is the use case for adding this parameter on the function? How are you suposed to pass it on an admin?

VincentLanglet: It would be easier to add it to the interface, but even if we don't, people can still pass the batch size as third parameter after a instanceof check.

PR #1737, opened 2023-04-10
Add tests for collection removal without triggering validation

jordisala1991:
## Subject

## Changelog

```markdown
### Changed
- Bump `sonata-project/form-extensions` to ^1.19
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "doctrine/persistence": "^3.0.2",
         "sonata-project/admin-bundle": "^4.18",
         "sonata-project/exporter": "^2.0 || ^3.0",
-        "sonata-project/form-extensions": "^1.4",
+        "sonata-project/form-extensions": "1.x-dev",
         "symfony/config": "^4.4 || ^5.4 || ^6.2",
         "symfony/console": "^4.4 || ^5.4 || ^6.2",
         "symfony/dependency-injection": "^4.4 || ^5.4 || ^6.2",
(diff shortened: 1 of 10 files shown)


jordisala1991:
Once the PR is merged and released, this should not fail (It will require bump requirements to not fail on lower deps @VincentLanglet

jordisala1991:
With this we have tests for the CollectionType, case where we fail validation (not removing an item) and case where the validation is not executing because even it the collection item is not valid, it is also removed.

composer.json
```diff
@@ -32,7 +32,7 @@
         "doctrine/persistence": "^3.0.2",
         "sonata-project/admin-bundle": "^4.18",
         "sonata-project/exporter": "^2.0 || ^3.0",
-        "sonata-project/form-extensions": "^1.4",
+        "sonata-project/form-extensions": "1.x-dev",
```
jordisala1991:
```suggestion
        "sonata-project/form-extensions": "^1.19",
```

VincentLanglet (approved):

VincentLanglet:
I would consider bumping a dependency minor as pedantic, no ?

jordisala1991:
Not sure about that TBH , first I tagged as pedantic, but since it can affect users of this package, I changed it to minor.

VincentLanglet:
> Not sure about that TBH , first I tagged as pedantic, but since it can affect users of this package, I changed it to minor.

Developper doesn't need any notification since composer will do the version resolution.
And it would be better to tell dev-kit that no release is needed so far for this library.

PR #1767, opened 2023-10-03
AbstractDateFilter::filterRange() applies a `WHERE` condition separately for the `start` and `end` values in the case of DateRangeOperatorType::TYPE_NOT_BETWEEN

### LINKED ISSUE #1766: DateTimeRangeFilter "Too few parameters" exception occurs when either the 'start' or 'end' field is empty
#### Sonata packages

<p>

```
Direct dependencies required in composer.json:
sonata-project/admin-bundle              4.14.0 4.27.1 The missing Symfony Admin Generator
sonata-project/doctrine-orm-admin-bundle 4.3.3  4.13.0 Integrate Doctrine ORM into the SonataAdminBundle
```

</p>
</details>

#### Symfony packages

<p>

```
symfony/asset                         v6.2.7  v6.3.0  Manages URL generation and versioning of web assets such as CS...
symfony/browser-kit                   v6.2.7  v6.3.2  Simulates the behavior of a web browser, allowing you to make ...
symfony/cache                         v6.2.13 v6.3.5  Provides extended PSR-6, PSR-16 (and tags) implementations
symfony/console                       v6.2.13 v6.3.4  Eases the creation of beautiful and testable command line inte...
symfony/css-selector                  v6.2.13 v6.3.2  Converts CSS selectors to XPath expressions
symfony/debug-bundle                  v6.2.13 v6.3.2  Provides a tight integration of the Symfony VarDumper componen...
symfony/dotenv                        v6.2.8  v6.3.0  Registers environment variables from a .env file
symfony/filesystem                    v6.2.12 v6.3.1  Provides basic utilities for the filesystem
symfony/finder                        v6.2.14 v6.3.5  Finds files and directories via an intuitive fluent interface
symfony/flex                          v2.3.2  v2.3.3  Composer plugin for Symfony
symfony/form                          v6.2.13 v6.3.5  Allows to easily create, process and reuse HTML forms
symfony/framework-bundle              v6.2.13 v6.3.5  Provides a tight integration between Symfony components and th...
symfony/intl                          v6.2.13 v6.3.2  Provides access to the localization data of the ICU library
symfony/lock                          v6.2.12 v6.3.2  Creates and manages locks, a mechanism to provide exclusive ac...
symfony/lokalise-translation-provider v6.2.8  v6.3.0  Symfony Lokalise Translation Provider Bridge
symfony/mailer                        v6.2.12 v6.3.5  Helps sending emails
symfony/maker-bundle                  v1.50.0 v1.51.1 Symfony Maker helps you create empty commands, controllers, fo...
symfony/mime                          v6.2.13 v6.3.5  Allows manipulating MIME messages
symfony/monolog-bundle                v3.8.0  v3.8.0  Symfony MonologBundle
symfony/phpunit-bridge                v6.3.2  v6.3.2  Provides utilities for PHPUnit, especially user deprecation no...
symfony/property-info                 v6.2.11 v6.3.0  Extracts information about PHP class' properties using metadat...
symfony/proxy-manager-bridge          v6.2.7  v6.3.0  Provides integration for ProxyManager with various Symfony com...
symfony/runtime                       v6.2.13 v6.3.2  Enables decoupling PHP applications from global state
symfony/security-bundle               v6.2.13 v6.3.5  Provides a tight integration of the Security component into th...
symfony/serializer                    v6.2.13 v6.3.5  Handles serializing and deserializing data structures, includi...
symfony/translation                   v6.2.11 v6.3.3  Provides tools to internationalize your application
symfony/twig-bundle                   v6.2.7  v6.3.0  Provides a tight integration of Twig into the Symfony full-sta...
symfony/validator                     v6.3.2  v6.3.5  Provides tools to validate values
symfony/web-profiler-bundle           v6.2.13 v6.3.2  Provides a development tool that gives detailed information ab...
symfony/workflow                      v6.2.13 v6.3.4  Provides tools for managing a workflow or finite state machine
symfony/yaml                          v6.2.10 v6.3.3  Loads and dumps YAML files
```

</p>
</details>

#### PHP version

```
$ php -v
PHP 8.2.9 (cli) (built: Aug 17 2023 23:22:23) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.9, Copyright (c) Zend Technologies
    with Zend OPcache v8.2.9, Copyright (c), by Zend Technologies
    with Xdebug v3.2.2, Copyright (c) 2002-2023, by Derick Rethans
```

## Subject

When setting a value for only one of the fields, either `start` or `end`, DateTimeRangeFilter throws an exception:
```
An exception has been thrown during the rendering of a template ("Too few parameters: the query defines 2 parameters but you only bound 1").
```
The problem is that AbstractDateFilter::filterRange doesn't has any additional logic for `DateRangeOperatorType::TYPE_NOT_BETWEEN` operator https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/4.x/src/Filter/AbstractDateFilter.php#L163

Also there is no validation for the form input.

## Steps to reproduce

1. Choose filter by Period
2. Advanced filters
3. Select Period -> `not between`
4. Set value only for field `From`
5. Filter

issue

## Expected results

Option 1: show the message that both fields `start` or `end` are required
Option 2: change behaviour as <>
Option 3: <suggest yours>

## Actual results

Exception: 
```
An exception has been thrown during the rendering of a template ("Too few parameters: the query defines 2 parameters but you only bound 1").
```

tonyaxo:
## Subject

AbstractDateFilter::filterRange() applies a `WHERE` condition separately for the `start` and `end` values in the case of DateRangeOperatorType::TYPE_NOT_BETWEEN.

I am targeting this branch, because it fixes bug in 4.x.

This fixes a bug that arises when one of this range borders is not set.

Closes #1766.

## Changelog

```markdown
### Fixed
- DateTimeRangeFilter exception occurs when either the `start` or `end` field is empty.
```

### DIFF
--- a/src/Filter/AbstractDateFilter.php
+++ b/src/Filter/AbstractDateFilter.php
@@ -160,7 +160,13 @@ abstract class AbstractDateFilter extends Filter
         $endDateParameterName = $this->getNewParameterName($query);
 
         if (DateRangeOperatorType::TYPE_NOT_BETWEEN === $type) {
-            $this->applyWhere($query, sprintf('%s.%s < :%s OR %s.%s > :%s', $alias, $field, $startDateParameterName, $alias, $field, $endDateParameterName));
+            if (null !== $value['start']) {
+                $this->applyWhere($query, sprintf('%s.%s %s :%s', $alias, $field, '<', $startDateParameterName));
+            }
+
+            if (null !== $value['end']) {
+                $this->applyWhere($query, sprintf('%s.%s %s :%s', $alias, $field, '>', $endDateParameterName));
+            }
         } else {
             if (null !== $value['start']) {
                 $this->applyWhere($query, sprintf('%s.%s %s :%s', $alias, $field, '>=', $startDateParameterName));
--- a/tests/Filter/DateTimeRangeFilterTest.php
+++ b/tests/Filter/DateTimeRangeFilterTest.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Sonata\DoctrineORMAdminBundle\Tests\Filter;
 
 use Sonata\AdminBundle\Filter\Model\FilterData;
+use Sonata\AdminBundle\Form\Type\Operator\DateRangeOperatorType;
 use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery;
 use Sonata\DoctrineORMAdminBundle\Filter\DateTimeRangeFilter;
 use Sonata\Form\Type\DateTimeRangeType;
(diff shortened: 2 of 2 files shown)


dmaicher (approved):

VincentLanglet (approved):

VincentLanglet:
Thanks, sorry for the delay

p1nGgeR:
Hi @VincentLanglet @tonyaxo 

It seems to me that this fix has broken the filer

Query before the fix 
`SELECT * FROM table WHERE column < :start OR column > :end;`

Query after the fix
`SELECT * FROM table WHERE column < :start AND column > :end;`

Am I missing something?

VincentLanglet:
Indeed, can you make a fix ?

I assume we have to check

If both start and end are not null, do the previous query with OR elseif start id not null do the start query elseif end is not null do the end query.

p1nGgeR:
@VincentLanglet 
I'm just thinking, maybe it doesn't make sense to use "filterRange" if both "start" and "end" dates are not provided?
If only one (start or end) date is provided, then it is not a correct use case for a date range filter.
I ask because there is already an embedded "if" and adding another "if" will not make this method any easier.

What if this check
```
if (
    !$value['start'] instanceof \DateTimeInterface
    && !$value['end'] instanceof \DateTimeInterface
) {
    return;
}
```
will be changed to 
```
if (
    !$value['start'] instanceof \DateTimeInterface
    || !$value['end'] instanceof \DateTimeInterface
) {
    return;
}
```

This change will make it possible to remove null checks for `$value['start'] / $value['end']` and apply single where for both `between` and `not between` types
```
if (DateRangeOperatorType::TYPE_NOT_BETWEEN === $type) {
    $this->applyWhere($query, sprintf('%s.%s < :%s OR %s.%s > :%s', $alias, $field, $startDateParameterName, $alias, $field, $endDateParameterName));
} else {
    $this->applyWhere($query, sprintf('%s.%s >= :%s AND %s.%s <= :%s', $alias, $field, $startDateParameterName, $alias, $field, $endDateParameterName));
}
```


I'm not sure I'm considering every scenario, though.
What are your thoughts?

VincentLanglet:
I agree that using the filter without both start and end is weird but if it just require an embedded if to support it I would prefer to keep the support rather than ignoring such case.

VincentLanglet:
I'll make a fix https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1807

p1nGgeR:
I was preparing a PR, but okay, thanks

PR #1788, opened 2023-12-19
Update UidFilter.php

josephzhao:
isSearchEnabled function returned null error

## Subject

I am targeting this branch, because {reason}.

Closes #{put_issue_number_here}.

## Changelog

```markdown
### Added
- Some `Class::newMethod()` to do great stuff

### Changed

### Deprecated

### Removed

### Fixed

### Security
```

### DIFF
--- a/src/Filter/UidFilter.php
+++ b/src/Filter/UidFilter.php
@@ -50,7 +50,7 @@ final class UidFilter extends Filter implements SearchableFilterInterface
 
     public function isSearchEnabled(): bool
     {
-        return $this->getOption('global_search');
+        return $this->getOption('global_search')===null?false:$this->getOption('global_search');
     }
 
     public function getDefaultOptions(): array


VincentLanglet:
Hi @josephzhao 

The issue is more with the fact there is no default value for the `global_search` option, like there is in the StringFilter
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/4.x/src/Filter/StringFilter.php#L115

@jordisala1991 When you introduced this filter it implemented the SearchableFilterInterface, is there a reason ?
Which value should be use as default for the global search (true ? false ?)

BillyMorgan:
Hate to be that *bump* guy but... bump? This is being made even worse for me cause I'm hitting it via the SonataUser bundle, which doesn't allow me to disable searching on the UUID fields. I'm gunna log an issue on that repo too and whoever fixes it first will be the coolest. That's right, it's a who's the coolest race.

VincentLanglet:
> Hate to be that _bump_ guy but... bump? This is being made even worse for me cause I'm hitting it via the SonataUser bundle, which doesn't allow me to disable searching on the UUID fields. I'm gunna log an issue on that repo too and whoever fixes it first will be the coolest. That's right, it's a who's the coolest race.

Instead being the one complaining, you can be the one fixing it.

This PR does not pass requirement therefor cannot be merged.
Also, I explained in my post the right way to fix it.

It's just to add `'global_search' => false,` in the getDefaultOptions method.

Feel free to open the PR instead waiting someone to work for you.

BillyMorgan:
Amazing! Thanks heaps for the fix @VincentLanglet :bow:

